PR #75, opened 2016-05-25
Add few more radio documentation pages

gbaman:
Add additional documentation pages for radio section for:
- Receive-string
- Send-string
- Set-transmit-power
- Write-value-to-serial


pelikhan:
Very nice!

PR #136, opened 2016-06-20
Documentation for Bluetooth IO Pin service block++

bluetooth-mdw:
Addition of event handler blocks for Bluetooth connection state changes.


docs/reference/bluetooth/start-io-pin-service.md
```diff
@@ -0,0 +1,32 @@
+# Bluetooth IO Pin Service
```
pelikhan: Please follow the same API doc structure as https://raw.githubusercontent.com/Microsoft/pxt-microbit/master/docs/reference/basic/show-number.md.

In this case, you are simply missing the first part where the signature is defined.

bluetooth-mdw:
Added a sig section. It doesn't get rendered as a block.... do I need to do something else?

Cheers

PR #206, opened 2016-08-18
Fix changeTempo() calculation

sgk:
Since I am very new to this project, I am afraid that I understand your intention wrongly.

`beat(BeatFraction.Whole)` is a duration of a beat in milliseconds. It seems to be wrong calculation to add the duration with bpm and then pass it to `setTempo()` which takes bpm as an argument.


pelikhan:
You are correct - this is a bug. You also need to call "init()" or use ``tempo()`` instead of ``beatsPerMinute`` so that it gets called.

sgk:
Oh, I see. I will add another commit.

PR #207, opened 2016-08-18
Converting the term "Compile" to "Download" in the docs

microsoftsam:
Fixes Microsoft/pxt#236


microsoftsam:
@pelikhan could you please add who ever is relevant that could review the doc changes.

PR #266, opened 2016-09-29
fixes builtin parts list for micro:bit

darzu:


pelikhan:
Integrated in other change.

PR #269, opened 2016-10-05
Deploy no longer throws when no drives found

guillaumejenkins:


cmds/cmds.ts
```diff
@@ -14,8 +14,11 @@ export function deployCoreAsync(res: ts.pxtc.CompileResult) {
         .then(drives => {
             if (drives.length == 0) {
                 let msg = "cannot find any drives to deploy to";
+                let err = new Error(msg);
+
+                (<any>err).code = "NOBOARD";
                 console.log(msg);
-                return Promise.reject(new Error(msg));
+                return Promise.reject(err);
```
pelikhan: we should not reject - instead we should return a result object with more info and let the caller deal with it.

cmds/cmds.ts
```diff
@@ -13,20 +13,19 @@ export function deployCoreAsync(res: ts.pxtc.CompileResult) {
...
 
             return Promise.map(drives, d =>
                 writeFileAsync(d + ts.pxtc.BINARY_HEX, res.outfiles[ts.pxtc.BINARY_HEX])
                     .then(() => {
-                        console.log("wrote hex file to " + d)
+                        console.log("wrote hex file to " + d);
                     }))
-        })
-        .then(() => { })
+                .then(() => drives.length);
```
pelikhan: Can you fix the indentation?

pelikhan:
Looks good

PR #273, opened 2016-10-11
Subcategories

riknoll:
In support of https://github.com/Microsoft/pxt/issues/436

If you have any suggestions for what should or shouldn't be "advanced", let me know.


libs/core/led.cpp
```diff
@@ -85,7 +88,7 @@ namespace led {
      */
     //% weight=1 help=led/set-display-mode
     //% parts="ledmatrix"
-    void setDisplayMode(DisplayMode_ mode) { 
+    void setDisplayMode(DisplayMode_ mode) {
```
pelikhan: Advanced

riknoll:
It's just removing trailing whitespace (VS code does that by default).

If you add `?w=1` to the end of a URL in Github while looking at a diff, whitespace changes will be ignored.

libs/core/led.cpp
riknoll: This one isn't actually a block

PR #280, opened 2016-10-19
New Radio API

riknoll:
Relies on this pxt PR: https://github.com/Microsoft/pxt/pull/522

The names of the packet parts in those screenshots are not right; I changed them but didn't want to redo all the pictures. Something to note is that changing the names of the "parameters" is a breaking change because we're using object destructuring rather than parameter names. So, please let me know if I should change them. The current names are:

```typescript
export class Packet {
    public sentNumber: number;
    public time: number;
    public serialNumber: number;
    public sentName: string;
    public signalStrength: number;
}
```

Receiving strings that were sent using `sendString()` is not yet supported, but I'll work on that soon.


libs/radio/radio.cpp
```diff
@@ -154,11 +154,31 @@ namespace radio {
     }
 
     /**
+     * Reads a string received in a packet received by ``receive number``. Not supported in simulator.
+     */
+    //% help=radio/received-value-name
+    //% weight=45 debug=true
+    StringData* receivedValueName() {
+        if (packet.length() > 12) {
```
abchatra: Don't you require 'if (radioEnable() != MICROBIT_OK) return 0; '?

riknoll: Yep!

riknoll:
I will reopen this in a bit, but I want to do all the breaking radio stuff at once so I'm going to add some more stuff to this PR

PR #284, opened 2016-10-22
Added a new project, the magic button trick

gingemonster:
Added a new project based on the tutorial video at https://www.youtube.com/watch?v=0Icwbcf3nXM


docs/projects.md
```diff
@@ -67,6 +67,10 @@ Fun games to build with your micro:bit.
   "name": "Radio",
   "url":"/projects/radio",
   "imageUrl":"/static/mb/projects/a9-radio.png"
+},{
```
pelikhan: How about starting a new section "Community"?

docs/projects/magic-button-trick.md
```diff
@@ -0,0 +1,97 @@
...
+## step 2: checking if the magnetic is nearby
+
+Now we can measure the magnetic force near the microbit, we can check if the value we measure is so big that it means there must be a strong magnet nearby. 
+
+If you have ever played with magnets you know they have two ends, often called a north and south 'pole'. Depending on which end of the magnet is pointing at the microbit the magnetic force measurement will either be a negative number like -100 or a positive number like 100. We just want to know if the strength is at least 100 we don't care if its negative or positive so we also use the 'absolute value of' block from the maths menu to tell our code to ignore the negative sign and just treat -100 as if its 100.
+
+So in the code below we will check if the absolute value of our magnetic field strength reading is more than 100 and save the result of that check in a new variable called 'isSwitched':
+
+```blocks
+var isSwitched = Math.abs(input.magneticForce(Dimension.Strength)) > 100
```
pelikhan: We try to use "let" instead of "var" in docs. The semantics are less confusing.

docs/projects/magic-button-trick.md
```diff
@@ -0,0 +1,97 @@
...
+
+```blocks
+
+var isSwitched = false;
+basic.forever(() => {
+    isSwitched = Math.abs(input.magneticForce(Dimension.Strength)) > 100
+})
+
+input.onButtonPressed(Button.A, () => {
+    if (isSwitched == true) {
```
pelikhan: Or just 

     if(isSwitched) {

docs/projects/magic-button-trick.md
```diff
@@ -0,0 +1,97 @@
...
+        basic.showString("B")
+    } else {
+        basic.showString("A")
+    }
+})
+input.onButtonPressed(Button.B, () => {
+    if (isSwitched == true) {
```
pelikhan: See comment above

docs/projects/magic-button-trick.md
```diff
@@ -0,0 +1,97 @@
...
+    basic.showString("A")
+})
+input.onButtonPressed(Button.B, () => {
+    basic.showString("B")
+})
+```
+
+## step 1: measuring magnetic force
+
+We will use the micro:bit's compass to detect the magnet. Compass's tell us what direction we are pointing by detecting the Earth's magnetic field but they can also detect any other magnet nearby. We will use that to check if our magnet is next to the micro:bit by using the 'magnetic force' block found in the input menu's 'more' section. As we only want to measure the strength we change the drop down to select 'strength':
```
pelikhan: Cross link to the magnetic force reference?

docs/projects/magic-button-trick.md
```diff
@@ -0,0 +1,97 @@
...
+If you have ever played with magnets you know they have two ends, often called a north and south 'pole'. Depending on which end of the magnet is pointing at the microbit the magnetic force measurement will either be a negative number like -100 or a positive number like 100. We just want to know if the strength is at least 100 we don't care if its negative or positive so we also use the 'absolute value of' block from the maths menu to tell our code to ignore the negative sign and just treat -100 as if its 100.
+
+So in the code below we will check if the absolute value of our magnetic field strength reading is more than 100 and save the result of that check in a new variable called 'isSwitched':
+
+```blocks
+var isSwitched = Math.abs(input.magneticForce(Dimension.Strength)) > 100
+```
+## step 3: running our 'magnet nearby' check all the time
+
+At the moment our code to detect the magnet being nearby will only run once so we need to put it into a 'foreever' block so that it keeps getting run again and again checking for the magnet to come near to the micro:bit. We should also make sure 'isSwitched' is false when our program starts.
```
pelikhan: foreever. Cross link to /reference/basic/forever

docs/projects/magic-button-trick.md
```diff
@@ -0,0 +1,97 @@
...
+```blocks
+var isSwitched = false;
+basic.forever(() => {
+	var isSwitched = Math.abs(input.magneticForce(Dimension.Strength)) > 100
+})
+```
+
+## step 4: swapping the buttons when we know the magnet is nearby
+
+Now we can check the value of our variable 'isChecked' whenever we want and we will know that the magnet is nearby if it's value is 'true'. Let's use that to change how the buttons work and complete the code for our trick. We will add an 'if, else' block to each button's code and check if we should swap over what each button displays because 'isSwitched' is equal to true:
```
pelikhan: isChecked -> isSwitched

PR #287, opened 2016-10-24
New Radio API

riknoll:
Fixes Microsoft/pxt#502, Microsoft/pxt#465, 

This PR creates a new block for receiving the different packet parts in a callback rather than the old API that required you to call `receiveNumber()`, `receiveString()` etc. Instead, we now have one callback that can receive the results from all three packet-sending APIs. The toolbox is pre-populated with three mutations of the callback block:

new_radio_toolbox

You can use the blue gear to add/remove parts of the packet from the block. If any part of the packet is not present in the received packet, the variable that is created will be given a default value. For example, if you receive a packet sent using `sentNumber()`, the `text` field will be set to the empty string. There isn't really a concept of null/undefined in the blocks, so default values is the most elegant way to handle it.

As a side note, this also fixes the bug where packets sent via `sendString()` could not be written to serial via `writeValueToSerial()`.

This change shouldn't really break old programs because all of the old blocks are still present (just hidden under the "More..." subcategory), but I guess there are three potential breaks.:

1. I changed the packet structure (but that isn't something we ever documented)
2. Maximum string length that can be sent via `sendString()` has been decreased to 19 characters to make room for the packet information we weren't sending before (serial number, time, packet type)
3. I removed the `receivedNumberAt()` API, which was marked as "debug" and had no corresponding block but was available in the typescript


riknoll:
The build is failing because this PR relies on a breaking change I made in https://github.com/Microsoft/pxt/pull/564

pelikhan:
- [ ] Can you add the reference doc page as well?
- [ ] Could we mark the "old" apis as obsolete?

riknoll:
Working on that now. I'll mark the APIs with a comment attribute like `//% deprecated=true` and mention it in the docs as well. The comment attribute won't do anything now, but we can do something with it later.

PR #289, opened 2016-10-24
Deprecating old radio APIs and fixing docs

riknoll:
Deprecates the old radio APIs so that they won't show up in the toolbox and documents the new APIs. I also added deprecation notices in the docs and tried to remove all of the references to the old APIs. Sorry about the whitespace changes.


riknoll:
I still need to update the serial API and deprecate the old one, but I'll do that in a separate PR

PR #291, opened 2016-10-25
Move hex cache

guillaumejenkins:
Moved offline HEX cache under /built


.gitignore
```diff
@@ -16,7 +16,6 @@ clients/win10/*.opendb
 clients/**/bin/**
 clients/**/obj/**
 clients/electron/projects
-hexcache
```
pelikhan: let's keep it for legacy

libs/templates/blocksprj/pxt.json
```diff
@@ -1,8 +1,8 @@
 {
     "name": "{0} block",
     "dependencies": {
-        "core": "*",
-        "radio": "*"
+        "core": "file:../../core",
```
pelikhan: Why using the file protocol here?

libs/templates/bluetoothprj/pxt.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "{0} block",
+    "dependencies": {
+        "core": "file:../../core",
+        "radio": "file:../../bluetooth"
```
pelikhan: Wrong package

PR #292, opened 2016-10-25
Renaming the string in onDataPacketReceived and updating docs

riknoll:
As discussed in stand up, this renames the string in `onDataPacketRecevied()`


pelikhan:
Looks. This should be pushed through ASAP as it is a breaking change.

PR #293, opened 2016-10-27
Rename the parameters for the sendValue block

riknoll:
Part of https://github.com/Microsoft/pxt/issues/595

By the way, it's sort of weird how we specify string values. Took me a while to realize I had to change the "e.g." example in the jsdoc for the function.


abchatra:
:shipit:

PR #295, opened 2016-10-28
Project templates

guillaumejenkins:


abchatra:
How does this work? How will the pxt target pickup these default projects?

pelikhan:
It gets injected in target.json/target.js at build time.

abchatra:
Got it thanks!

PR #298, opened 2016-11-01
Adding an empty proj files to tsproj

abchatra:
This is show the option to switch to blocks editor option once hex file is downloaded from the docs.

### DIFF
--- a/libs/templates/tsprj/pxt.json
+++ b/libs/templates/tsprj/pxt.json
@@ -6,6 +6,7 @@
     },
     "description": "",
     "files": [
+        "main.blocks",
         "main.ts",
         "README.md"
     ]


libs/templates/tsprj/pxt.json
```diff
@@ -6,6 +6,7 @@
     },
     "description": "",
     "files": [
+        "main.blocks",
```
pelikhan: tsprj is only used from the command line, e.g. VS code for example. This will create a useless file on disk that might be confusing.

In the web editor, we could automatically add a .blocks file if missing.

pelikhan: Does blockly load main.blocks properly? It should have an empy XML payload I think.

pelikhan:
I think .blocks file has to be a valid xml payload. You can load blocksprj and override the content of main.blocks.

abchatra:
Closing this in preference to change in https://github.com/Microsoft/pxt/pull/642

PR #305, opened 2016-11-10
pxt-microbit can now be required and exports useful information about its bundled pxt-core

guillaumejenkins:


mainCli.js
```diff
@@ -0,0 +1,13 @@
+const pxtCore = require("pxt-core");
```
pelikhan: - all filenames are typically lower cased, we update this?

mainCli.d.ts
```diff
@@ -0,0 +1,3 @@
+export function mainCli(args: string[]): void;
```
pelikhan: Can we namespace these functions?

pelikhan:
Why not simply returning the pxtCore instance?

pelikhan:
I would keep it simply and simply return pxtCore and let the pxt engine deal with it. Otherwise we are going to duplicate this code in all targets. Let's keep this to a 1 line change.

mainCli.d.ts
guillaumejenkins: This is pure javascript, since we didn't want to have to add a tsconfig.json and a compile mechanism to keep it simple. If we want to move this to typescript I can certainly do it, otherwise we can't

Please note that this will be down to a single property, pxtCore, so maybe there's no need to wrap this anyway

pelikhan:
We already have https://github.com/Microsoft/pxt-microbit/tree/master/cmds which we compile and require in the pxt. We could add a function in there.

PR #315, opened 2016-12-05
add false for advertiseEddystoneUrl

OwenBrotherwood:
I used https://bluetooth-mdw.blogspot.dk/2016/11/microbit-proximity-beacons.html for reference, there is cpp code therin

Currently investigating why the advertised url. http://pxt.io is corruptted on the Windows App Bluetooth Beacon Interactor that I use as a baseline.

Note that https://show.io/ micro:bit hex code does not have a problem with the url with the refereance app.

https://www.microsoft.com/da-dk/store/p/bluetooth-beacon-interactor/9nblggh1z24k

### DIFF
--- a/libs/bluetooth/bluetooth.cpp
+++ b/libs/bluetooth/bluetooth.cpp
@@ -132,7 +132,7 @@ namespace bluetooth {
     //% help=bluetooth/advertise-url
     void advertiseUrl(StringData* url, int power) {
         int8_t level = CALIBRATED_POWERS[min(7, max(0, power))];
-        uBit.bleManager.advertiseEddystoneUrl(ManagedString(url), level);
+        uBit.bleManager.advertiseEddystoneUrl(ManagedString(url), level, false);
     }
 
     /**


OwenBrotherwood:
Further clarification of URL advertised
http://pxt.io is beaconed as ttp:\pxt.io

I look a bit more in the code: but I may fail as it is deep magic.

OwenBrotherwood:
Using the blocks shown, sim shows correct text, but micro:bit shows the same text as the beaconed url ... almost:
http:\pxt.io

The initial h is present but the two // are transformed to \

image

OwenBrotherwood:
Altering bluetooth.cpp for predefined URL works
```

    void advertiseUrl(StringData* url, int power) {
        char URL[] = "http://pxt.io";
        int8_t level = CALIBRATED_POWERS[min(7, max(0, power))];
        uBit.bleManager.advertiseEddystoneUrl(URL, level, false);
    }

```

OwenBrotherwood:
http://goo.gl/TlUTF7 
is transformed to 
ttp:\goo.gl/TIUTF
ie. first and last characters gone and // transformed to \
```

    void advertiseUrl(StringData* url, int power) {
        int8_t level = CALIBRATED_POWERS[min(7, max(0, power))];
        uBit.bleManager.advertiseEddystoneUrl(ManagedString(url), level, false);
    }
```

OwenBrotherwood:
https://github.com/google/eddystone/tree/master/eddystone-url#url-scheme-prefix
Contains references to url encoding in connection with http:// and https:// with the use of a byte code to represent the start of an url with `//` in it

May give a work around until a solution to https://github.com/Microsoft/pxt/issues/846 is found which could be a general issue concerning `//`

I see if I can do a quick workaround this evening unless someone gets there before me

pelikhan:
Seems like an issue with microbit-DAL, you should report it there.

OwenBrotherwood:
I finally found out what the boolean false is controlling

> The boolean argument, set to false here, indicates whether or not the micro:bit should accept connections. Sometimes you might like to have beacon behaviour but still want it to be possible to connect to the micro:bit so you can access one or more of the GATT services that the micro:bit has. In this situation, your config.json file may need to be adjusted to allow for the GATT and security capabilities you want and you will need to ensure your hex file contains the Bluetooth services you plan to use as well.


https://lancaster-university.github.io/microbit-docs/ble/eddystone/

cpseager:
Looks like ManagedStrings were removed from DAL as not working....

https://github.com/lancaster-university/microbit-dal/commit/10ce2b9f7a8f6f09101c87e9e56437018abc4f8b

pelikhan:
They were removed for UID not URL.

OwenBrotherwood:
The added false (from cut and paste of the cpp I saw) is actually default true

https://github.com/lancaster-university/microbit-dal/blob/12bf127ae28c93498a5d3cf50894866496d9d335/source/bluetooth/MicroBitBLEManager.cpp#L515

pelikhan:
@OwenBrotherwood please open a bug on the lancaster/microbit-dal repo. The issue seems to reside in the C++ layer.

OwenBrotherwood:
The "default" of true does NOT work.
So far, only false allows some eddystone to come out.

HARD coding of the char URL[] gives an url as seen by reference app.
ManagedString(URL) also works

```
    void advertiseUrl(StringData* url, int power) {
        char URL[] = "http://pxt.io";
        int8_t level = CALIBRATED_POWERS[min(7, max(0, power))];
        uBit.bleManager.advertiseEddystoneUrl(URL, level, false);
    }

```

image

OwenBrotherwood:
There seems to be a few issues here:
- the default of true for the boolean argument does not work: using my pull will get that to work
- the url that is typed into the block is not presented to the pxt-microbit advertiseUrl routine in bluetooth.cpp: this seems to be a pxt-microbit bug (see also https://github.com/Microsoft/pxt/issues/846)

Until the above are fixed, the beacon is bust.

I opened an issue for some info for the default true scenario, as it would be "nice" that the micro:bit could beacon while also accepting bluetooth connections: https://github.com/lancaster-university/microbit-dal/issues/248

OwenBrotherwood:
I tested this moring with pxt 	0.5.91 and pxt-microbit 0.6.30

OwenBrotherwood:
I have to repeat tests: I cannot get it to work this evening (may have taken a wrong git directory when testing)
I will check the emitted json's to check agains the comment from Lancaster.
>I tested this moring with pxt 0.5.91 and pxt-microbit 0.6.30 : beacon comes but with the problem of http://pxt.io showing wrongly

PR #318, opened 2016-12-07
Update bluetooth.cpp

OwenBrotherwood:
uBit.bleManager.advertiseEddystoneUrl(ManagedString(url), level, true); 
Ensure that the default of true is visible in pxt-microbit code such that at a latter date, a user choice for allowing connections to other services is under user, block, control. false: other services connections not allowed


pelikhan:
Superceeded by https://github.com/Microsoft/pxt-microbit/commit/6c11dbcdf4310617d3cab81b16a23b8e1683e2e2

PR #319, opened 2016-12-07
Update pxt.json

OwenBrotherwood:
ensure that conflict between Pairing Modes, where pairing is required does not affect user choice of Eddystone.
Note: user confusion in connection with the choice of pairing and that default JustWorks is overridden.

Lancaster, mdw input
https://github.com/lancaster-university/microbit-dal/issues/251

### DIFF
--- a/libs/bluetooth/pxt.json
+++ b/libs/bluetooth/pxt.json
@@ -24,7 +24,7 @@
             "microbit-dal": {
                 "bluetooth": {
                     "private_addressing": 0,
-                    "whitelist": 1,
+                    "whitelist": 0,
                     "advertising_timeout": 0,
                     "tx_power": 6,
                     "dfu_service": 1,
@@ -32,8 +32,8 @@
                     "device_info_service": 1,
                     "eddystone_url": 1,
                     "eddystone_uid": 0,
-                    "pairing_mode": 1,
-                    "open": 0,
+                    "pairing_mode": 0,
+                    "open": 1,
                     "security_level": "SECURITY_MODE_ENCRYPTION_NO_MITM"
                 }
             },


pelikhan:
This change would make the "No security" default. This is not what we want. The current default setting is "Just Works". The UI should be smarted that it is the default and turn on the knob accordingly.

OwenBrotherwood:
The conflict is evident: the code to switch knob creates another problem that is not addressed in the pull

OwenBrotherwood:
Do you wish another pull with the knob logic?
I feel that there are other pressing issues that mean Eddystone is (almost) an enhancement due to http:// going to ttp:\ which needs to be solved before any real use of the feature is possible,

pelikhan:
Not needed. Encoding issues need to be solved at the DAL level.
> The conflict is evident: the code to switch knob creates another problem that is not addressed in the pul
I do not understand this comment.

OwenBrotherwood:
OK: I try again.

Summary:
https://github.com/Microsoft/pxt/issues/846 bug p3:  show string block corrupts a\\c to a/c
- not in this issue, url is corrupted from http:// to ttp:\ with url parameter to bluetooth.cpp  void advertiseUrl(StringData* url, int power)
 
I have previously posted that if the url that is presented in bluetooth.cpp  void advertiseUrl(StringData* url, int power) to DAL Level  uBit.bleManager.advertiseEddystoneUrl(ManagedString(url), level, true);, the url is shown correctly.
ie there is **no encoding issue at the DAL level**, but there is corruption of the url presented to  bluetooth.cpp  void advertiseUrl(StringData* url, int power)
I have used a local installation of Yotta to ensure this.

The present analysis of if beacon support is:
- text field is corrupting the block url (also javascript, if I remember correctly)
possible solution is enumerate the parameter for eddystone with drop down for http: (etc) such that the user is clear that the url text field may only hold up to 17 characters,  with an enumeration of the .com type domains 
Reference
https://github.com/google/eddystone/tree/master/eddystone-url#url-scheme-prefix
https://github.com/google/eddystone/tree/master/eddystone-url#eddystone-url-http-url-encoding

The removal of the first character in the http sequence (h) is not present in issue 846 and can not be bypassed at present.

Documentation of the conflict for JustWorks and Eddystone is documented here
https://github.com/lancaster-university/microbit-dal/issues/251

OwenBrotherwood:
The original perceived conflict for JustWorks and Eddystone seems to have been withdrawn
However, testing required.
https://github.com/lancaster-university/microbit-dal/issues/251#issuecomment-265995660

PR #320, opened 2016-12-08
Placement new occurances need explicit delete

abchatra:
We allocate RefRecord and RefAction using placement new. Mapping nicely with the equivalent delete as well


riknoll:
My C++ is a bit rusty, but LGTM

PR #326, opened 2016-12-24
Array enhancements

abchatra:
This is a redesign of the array motivated by https://github.com/Microsoft/pxt/issues/953, https://github.com/Microsoft/pxt/issues/919 & https://github.com/Microsoft/pxt/issues/322
Current design lets us to implement variety of array built-ins such as shift\unshift\sort pretty easily. 

- RefCollection now points to a single segment which has length and size instead of std::vector.
- This segment grows based on the heuristics in _growthFactor_ (initially doubling and then increasing by 1.66 etc).
- The length represents the valid elements of the array. The size includes the unused elements as well.(length<=size). This is to avoid repeated reallocation for every element.
- Missing elements are represented by INT_MIN (0x80000000).  Notion of missing elements is required to do random or reverse set on the array. Access of missing element is still an error.
- The length property is automatically incremented (push) or decremented (pop) or on set beyond the length.
- No shrinking of the array segment yet. 

I am planning to do a separate set of pull requests to mimic similar behavior in the simulator and more robust built-in support once this is merged.


abchatra:
Thanks @riknoll . I am going to merge this and address any more review comments later. I have couple of dependent changes coming in.

mmoskal:
It looks good! Moving away from std::vector is surely a good idea!

My main comment regards semantics: I don't think we crash on array out of bounds access; I understand the previous implementation did this, but I guess this is just TD legacy. Now we're using JavaScript semantics and there invalid array accesses return `undefined`. Because we have no `undefined` (and this is well-documented), we should just return `0`, `null` or `false`.

In light of this, we also don't need the `MissingValue`, which itself is problematic if the `Segment` is used to implement `number[]` and someone stores that exact value in there. We then also wouldn't need the `getNextValidIndex`, since all indices would be valid.

As a side note: it could be nice to have the `Segment` be a general purpose template, parametrized with type. Then we could move it into DAL.

abchatra:
@mmoskal  I am going to make a PR to update the default values. Regarding moving to DAL we can take it up later. 

Quick q: Does empty string ("") good default value for string arrays (or null)? 0, null, false makes sense for the rest.

mmoskal:
I think we should stick with JavaScript, so 'null'

PR #328, opened 2017-01-02
insertAt support for array

abchatra:
This adds insertAt support for array. Also updates removeAt to return the removed element and shift the rest of the elements.

### DIFF
--- a/libs/core/pxt.cpp
+++ b/libs/core/pxt.cpp
@@ -288,7 +288,8 @@ namespace pxt {
       return 0;
     }
 
-    //this function removes index i and     
+    //this function removes an element at index i and shifts the rest of the elements to
+    //left to fill the gap   
     uint32_t Segment::remove(uint32_t i) 
     {
 #ifdef DEBUG_BUILD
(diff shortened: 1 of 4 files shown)


libs/core/pxt.cpp
```diff
@@ -269,17 +288,63 @@ namespace pxt {
       return 0;
     }
 
-    void Segment::remove(uint32_t i) 
+    //this function removes index i and
```
riknoll: Comment seems incomplete here

libs/core/pxt.cpp
```diff
@@ -269,17 +288,63 @@ namespace pxt {
...
+        this->print();
+#endif  
+        return ret;
       }
-      return;
+      error(ERR_OUT_OF_BOUNDS);
+      return 0;
+    }
+
+    //this function removes index i and
```
riknoll: I think this is a copy-pasted duplicate of the comment on remove()

riknoll:
@abchatra build is failing right now

abchatra:
Fixed review comments

PR #340, opened 2017-01-11
fix"string compare works differently in simulator/on target"

chmora:
Fixes Microsoft/pxt#901

### DIFF
--- a/libs/core/core.cpp
+++ b/libs/core/core.cpp
@@ -21,11 +21,11 @@ namespace String_ {
 
     //%
     int compare(StringData *s, StringData *that) {
-      int compareResult  = strcmp(s->data, that->data);
+      int compareResult = strcmp(s->data, that->data);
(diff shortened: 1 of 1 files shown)


libs/core/core.cpp
```diff
@@ -21,7 +21,12 @@ namespace String_ {
 
     //%
     int compare(StringData *s, StringData *that) {
-        return strcmp(s->data, that->data);
+      int compareResult  = strcmp(s->data, that->data);
```
microsoftsam: nit: extra space after compareResult

libs/core/core.cpp
```diff
@@ -21,7 +21,12 @@ namespace String_ {
...
+      if(compareResult < 0)
+        return -1;
+      else if(compareResult > 0)
+          return 1;
```
microsoftsam: nit: make sure both return statements align

PR #344, opened 2017-01-17
package approval docs

pelikhan:


docs/packages.md
```diff
@@ -3,27 +3,43 @@
...
 
-Packages can be published from the pxt command line. Check out [the docs](https://www.pxt.io/packages).
+1. setup a public GitHub repository with your package sources
+2. ensure that the license is **Apache2** or **MIT**
+3. fill ``test.ts`` with a test scenario
+4. ensure that the package builds by running ``pxt`` from the package folder. [Read more...](https://www.pxt.io/packages).
+5. fill up the ``README.md`` file with documentation. See [markdown docs](https://pxt.io/writing-docs) for syntax.
+6. add a screenshot image at ``/docs/static/screenshot.png`` for use in the web site
+7. publish a release by running ``pxt bump`` from the package folder.
+8. open a support ticket at https://support.microbit.org and select the **package approval** option.
```
jaustin: We don't have a 'ticket type' options exposed yet, so for now let's say 'put "package approval" into the subject line. OR we could give you an iframe to embed here that would set the ticket type, so it could just say "fill in the form below" and we could then require things like the link to their GitHub repo, etc.

PR #347, opened 2017-01-26
Fixing singal strength docs and adding an example

riknoll:
The correct range (according to the DAL) is -42 to -128. Fixes https://github.com/Microsoft/pxt/issues/1167

Also adds an example using the rssi to show the distance between two micro:bits since that's fun.


pelikhan:
I will talk to Joe about this today to determine the final range.

PR #349, opened 2017-01-30
Expose pin.eventOn() method in blocks and typescript

riknoll:
Fixes https://github.com/Microsoft/pxt/issues/1215

edge_events

No simulator support (because we don't support pulse events either)!


rharbird:
Hello,

Can you tell me what the deployment schedule is for this? I just need to know whether to fix locally and deploy the dev. environment with the fix on the class laptops. Obviously, I'd prefer not to do that if possible.

Thanks very much,



Rae

riknoll:
@rharbird It's live now at [pxt.microbit.org/beta](http://pxt.microbit.org/beta)

rharbird:
Thanks PXT team! We're using it and it works.

riknoll:
Great!

PR #354, opened 2017-02-10
Delay initializing the global icons

abchatra:
Global icons where polluting the global namespace and increasing hex size considerably. Delay initializing them so that only if they are used then the icons are created.


abchatra:
@pelikhan sign off?

PR #355, opened 2017-02-21
Blockly integrate pxt microbit

chmora:


libs/core/music.ts
```diff
@@ -1,112 +1,112 @@
 enum Note {
-    //% blockIdentity=music.noteFrequency
+    //% blockIdentity=music.noteFrequency noteValue=262
```
pelikhan: Why do we have an additional flag noteValue? We also have enumValue.

microsoftsam: Agreed, I believe this was going to be used in an attempt to convert Blockly note fields -> Typescript Enums. Please remove this in the meantime.

chmora: I use that to convert from TypeScript Enums to Blockly note field. In decompiler.ts I check if that flag exists to pass the frequency value to the field instead of the enum text

microsoftsam: Are you able to use enumValue as peli alluded

PR #358, opened 2017-02-25
Adding support for music melody arrays

microsoftsam:
Some melodies to test: 

```typescript
music.playMelody(['c4:1', 'e', 'g', 'c5', 'e', 'g4', 'c5', 'e', 'c4', 'e', 'g', 'c5', 'e', 'g4', 'c5', 'e', 'c4', 'd', 'g', 'd5', 'f', 'g4', 'd5', 'f', 'c4', 'd', 'g', 'd5', 'f', 'g4', 'd5', 'f', 'b3', 'd4', 'g', 'd5', 'f', 'g4', 'd5', 'f', 'b3', 'd4', 'g', 'd5', 'f', 'g4', 'd5', 'f', 'c4', 'e', 'g', 'c5', 'e', 'g4', 'c5', 'e', 'c4', 'e', 'g', 'c5', 'e', 'g4', 'c5', 'e'
        ]);

music.playMelody(['c4:1', 'd', 'e:2', 'g', 'd:1', 'e', 'f:2', 'a', 'e:1', 'f', 'g:2', 'b', 'c5:4']);

music.playMelody(['g4:1', 'c5', 'e', 'g:2', 'e:1', 'g:3']);

music.playMelody(['g5:1', 'd#', 'c', 'g4:2', 'b:1', 'c5:3'])
```

Added micropython samples: 
screen shot 2017-02-25 at 3 44 38 pm


microsoftsam:
Can also add the micropython melodies in blocks (just like show icon)

libs/core/melodies.ts
```diff
@@ -0,0 +1,108 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+// Melodies from file microbitmusictunes.c https://github.com/bbcmicrobit/MicroPython
+
+enum MelodyNames {
+    //% block="dadadum"
+    DADADADUM = 0,
```
pelikhan: Enum values are capitalized I think.

libs/core/music.ts
```diff
@@ -240,4 +241,54 @@ namespace music {
...
         }
     }
+
+    /**
+     * Plays a melody through pin ``P0``. 
+     * Notes are expressed as a string of characters with this format: NOTE[octave][:duration]
+     * @param melody the melody array to play, eg: ['g5:1', 'f', 'e', 'd', 'c']
+     */
+    export function playMelody(melody: string[]) {
+        let beat = 20000 / beatsPerMinute;
```
pelikhan: The beats could be computed in the loop to allow another fiber to modify it.

libs/core/melodies.ts
```diff
@@ -0,0 +1,108 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+// Melodies from file microbitmusictunes.c https://github.com/bbcmicrobit/MicroPython
+
+enum MelodyNames {
```
pelikhan: Melodies ?

libs/core/melodies.ts
```diff
@@ -0,0 +1,108 @@
...
+     */
+    //% help=music/play-melody weight=100
+    //% blockId=device_play_melody block="play|melody %melody"
+    //% parts="headphone"
+    export function playBuiltinMelody(melody: MelodyNames): void {
+        let res = getMelody(melody)
+        music.playMelody(res)
+    }
+
+    function getMelody(i: MelodyNames): string[] {
```
pelikhan: Separate function seems like overkill here.

libs/core/melodies.ts
microsoftsam: Sounds good

microsoftsam:
Do we need help docs for the built in melody?

libs/core/melodies.ts
```diff
@@ -0,0 +1,130 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+*/
+
+// Melodies from file microbitmusictunes.c https://github.com/bbcmicrobit/MicroPython
+
+enum Melodies {
+    //% block="dadadum"
+    DADADADUM = 0,
```
pelikhan: Capitalized instead of all caps.

pelikhan:
Would be nice to have docs along new apis. We should have docs for showIcon and images too. Documentation debt building up...

libs/core/melodies.ts
```diff
@@ -0,0 +1,130 @@
...
+    //% block="power down"
+    POWER_DOWN,
+}
+
+namespace music {
+
+    /**
+     * Plays the selected built-in melody through pin ``P0``. 
+     */
+    //% help=music/play-melody weight=100
```
pelikhan: Should be pointing to music/play-builtin-melody

pelikhan: I don't think it should be on top of the category. Should be after playnote/tone/rest

libs/core/music.ts
```diff
@@ -240,4 +241,53 @@ namespace music {
             beatsPerMinute = Math.max(1, bpm);
         }
     }
+
+    /**
+     * Plays a melody through pin ``P0``. 
+     * Notes are expressed as a string of characters with this format: NOTE[octave][:duration]
+     * @param melody the melody array to play, eg: ['g5:1']
+     */
+    export function playMelody(melody: string[]) {
```
pelikhan: Missing help link

libs/core/melodies.ts
microsoftsam: Okay moving it down the list.

microsoftsam:
Not sure I've quite nailed the cancelling the way I've implemented it, and it doesn't seem to work the same way as animations. 
First of all, the way it is now, it doesn't get cancelled on the physical device (because the event handler doesn't get a chance to execute and cancel it). 

Also, the way it is now: if you have a forever loop playing a melody it will always win, and cancel any other melody playing.

libs/core/music.ts
```diff
@@ -240,4 +241,90 @@ namespace music {
...
+
+    let currentMelody: Melody;
+
+    /**
+     * Plays a melody through pin ``P0``. 
+     * Notes are expressed as a string of characters with this format: NOTE[octave][:duration]
+     * @param melody the melody array to play, eg: ['g5:1']
+     */
+    export function playMelody(melodyArray: string[]) {
+        let melody = new Melody(melodyArray);
```
pelikhan: Move this line after cleanup to reclaim memory.

pelikhan:
Let's discuss those cancellation semantics. How would one implement a background music? (Or is it not a scenario?)

libs/core/melodies.ts
```diff
@@ -0,0 +1,147 @@
...
+
+
+    /**
+     * Plays the selected built-in melody through pin ``P0`` once, in the foreground.
+     * @param melody the melody to play
+     */
+    //% help=music/play-builtin-melody weight=60
+    //% blockId=device_play_builtin_melody block="play|melody %melody"
+    //% parts="headphone"
+    export function playBuiltinMelody(melody: Melodies): void {
```
pelikhan: use function that converts Melody to string[] instead

microsoftsam: Okay

microsoftsam:
@riknoll is taking a look at a bug in the pxt where blockly doesn't recognize string arrays as the same type. (in the converter function)

PR #366, opened 2017-03-03
Creating images directly from createImage API

abchatra:
reduces memory pressure


libs/core/icons.ts
```diff
@@ -194,317 +194,317 @@ namespace basic {
 
 namespace images {
 
-    function getArrow(i : ArrowNames): string {
+    function getArrow(i : ArrowNames): Image {
```
pelikhan: do we even need "getArrow"?

abchatra: image

pelikhan: I mean, what about moving the content of getArrow under showArrow.

mmoskal:
Lgtm

PR #369, opened 2017-03-11
TOC

microsoftsam:
Table of contents


docs/SUMMARY.md
```diff
@@ -0,0 +1,119 @@
+# Summary
+
+### Projects
+
+* [Things to do](/docs)
+    * [Make](/projects.md)
```
pelikhan: .md not needed

microsoftsam:
Yes.

Can probably write a script to create this file, but then titles will always have to have the same name as the file, additionally it wouldn't be possible to leave out part of the docs from the TOC (e.g: the lessons section)

pelikhan:
Makes sense.

PR #376, opened 2017-03-26
add enum baudrates

OwenBrotherwood:
fixes perceived limitation of available baud rates https://github.com/Microsoft/pxt/issues/1752


libs/core/serial.cpp
```diff
@@ -15,8 +15,26 @@ enum SerialPin {
...
   //% block=9600
   BaudRate9600 = 9600
+  //% block=4800
+  BaudRate4800 = 4800,    
+  //% block=2400
+  BaudRate2400 = 2400,    
+  //% block=1200
+  BaudRate1200 = 1200,    
+  //% block=300
+  BaudRate300 = 300,
```
pelikhan: spurious comma I think

pelikhan:
enum.d.ts needs to be regenerated

OwenBrotherwood:
enum.d.ts and core-strings.json coming soon
I double check to see what else I missed

OwenBrotherwood:
Order of baudrates isn't numeric: hope this is ok
image

pelikhan:
That's ok.

OwenBrotherwood:
Side note while I remember, enum.d.ts and core-strings.json after changes of serial.cpp where created with pxt serve on a git clone pxt-microbit with npm install -g pxt.
I investigate the pxt serve -yt / pxt buildtarget after I do some fix up on this disastrous PR

OwenBrotherwood:
:( My Windows Yotta is bust: I cannot check with pxt buildtarget ( pxt serve -yt).
However, it seems that the files where generated without yotta (enum.d.ts and core-strings.json)
And, if the tests are good, all checks have passed.

PR #383, opened 2017-04-17
Adding baud rate 31250

abchatra:
Fixing issue https://github.com/Microsoft/pxt/issues/1902


abchatra:
I am not sure if we need to bump the minor version. API already accepted int and there is no change there. (Also this is such a minor thing).  @jaustin what are your thoughts. 

Happy to do so though.

pelikhan:
Still it is a new field on the enum.

PR #384, opened 2017-04-17
support for serial read/write buffer

pelikhan:


tballmsft:
Looks good to me, @pelikhan

PR #387, opened 2017-04-20
Fix plotImage and showImage behavior in the sim 

riknoll:
Tested on hardware. Both `plotImage()` and `showImage()` respect the animation queue. The difference is that `showImage()` is async and lets you specify a delay.


tballmsft:
Looks good to me, @riknoll.Thanks.

PR #396, opened 2017-05-12
get showString and showNumber in alignment w.r.t asynch semantics

tballmsft:


tballmsft:
Fixes issue https://github.com/Microsoft/pxt/issues/2060

PR #397, opened 2017-05-12
fix problem with showNumber and showString on negative interval

tballmsft:


tballmsft:
Fixes https://github.com/Microsoft/pxt/issues/2060

PR #398, opened 2017-05-16
add groupID to radio packet

tballmsft:
Fixes https://github.com/Microsoft/pxt/issues/2104


tballmsft:
Oops. Need to change something in pxt.

PR #400, opened 2017-05-17
Pins gestures grid

guillaumejenkins:
**We should discuss this before merging**

Changes pins and gestures to use the grid picker:

gridpickermicrobit


pelikhan:
LGTM

PR #402, opened 2017-05-19
moving pairing options to core, do not disable ble in core

pelikhan:
Fix for https://github.com/Microsoft/pxt/issues/2113


pelikhan:
- [ ] radio test
- [ ] radio -> ble just works pairing test
- [ ] flash over ble test

PR #408, opened 2017-05-28
fix showAnimation and scrollImage

tballmsft:
We were calling the DAL with default value (true) for "autoclear", so that "autoclear" took place after the animation/scroll, clearing the display, which is not the desired behavior (or the behavior of the simulator). Fixes:

https://github.com/Microsoft/pxt/issues/2131

Tested on micro:bit - now has same behavior as simulator

### DIFF
--- a/libs/core/basic.cpp
+++ b/libs/core/basic.cpp
@@ -87,7 +87,7 @@ namespace basic {
     //% help=basic/show-animation imageLiteral=1 async
     //% parts="ledmatrix"
     void showAnimation(ImageLiteral leds, int interval = 400) {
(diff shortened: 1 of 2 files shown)


tballmsft:
In the case of showAnimation, we simply needed to override the default autoclear parameter value. In the case of showImage, @whaleygeek's program still did not work correctly. Using
```
MICROBIT_DISPLAY_ANIMATE_DEFAULT_POS
```
instead of
```
MICROBIT_DISPLAY_WIDTH - 1
```
as the starting position fixed the incorrect behavior of showImage.

PR #428, opened 2017-06-27
Add the 'Conditionals' lesson to csintro.

ganicke:


pelikhan:
Don't forget to search and replace PXT.microbit.org links. LGTM

ganicke:
On the activity pages they have links to the projects like:
See the activity online: [**Scorekeeper**](https://makecode.microbit.org/47893-98679-98470-56344)
Should we just remove those since the same thing is in already in the page **blocks** sections?

pelikhan:
Yes it will be easier to maintain without those.

PR #429, opened 2017-06-27
Egg and Spoon race

microsoftsam:
Balance a micro:bit in an egg and spoon race.


pelikhan:
Cool. Looks like we never ported the level app to microbit!

PR #437, opened 2017-07-03
Remove links to online projects from lesson pages.

ganicke:


ganicke:
Code examples already in lesson content. Links to duplicate online examples are removed.

PR #443, opened 2017-07-10
Update blocks pages to match api changes.

ganicke:


libs/core/enums.d.ts
```diff
@@ -365,7 +365,7 @@ declare namespace control {
 
     declare enum DisplayMode {
     //% block="black and white"
-    BackAndWhite = 0,  // DISPLAY_MODE_BLACK_AND_WHITE
+    BlackAndWhite = 0,  // DISPLAY_MODE_BLACK_AND_WHITE
```
pelikhan: This is a breaking change.

libs/core/led.cpp
```diff
@@ -2,7 +2,7 @@
 
 enum class DisplayMode_ {
     //% block="black and white"
-    BackAndWhite = DISPLAY_MODE_BLACK_AND_WHITE,
+    BlackAndWhite = DISPLAY_MODE_BLACK_AND_WHITE,
```
pelikhan: Breaking change

libs/core/led.cpp
```diff
@@ -99,7 +100,7 @@ namespace led {
     /**
     * Turns on or off the display
     */
-    //% help=led/enable blockId=device_led_enable block="led enable %on"
+    //% help=led/enable blockId=device_led_enable block="enable %on"
```
pelikhan: Why drop the led?

libs/core/led.ts
```diff
@@ -92,9 +94,10 @@
 
     /**
      * Fades in the screen display.
-     * @param ms TODO
+     * @param ms fade-in time in milliseconds
      */
     //% help=led/fade-in
+    //% blockId=device_led_fade_in block="fade in (ms) %ms" icon="\uf204"
```
pelikhan: Icon should not be needed

libs/core/led.cpp
ganicke: Don't have to change it. You're right it could break outside code. I'll roll that back.

libs/core/led.ts
ganicke: Ok, not clear on the use of the icon since toggle() has it as well. This is the led matrix icon and not the namespace icon? It's not used for the non-display methods? Sorry, I should've checked the render code to see how icon is used.

libs/core/led.cpp
ganicke: It was one of the review items from the Cambridge folks. Should led.enable() show in blocks as
"led enable" or just "enable". What about "led toggle" instead of just "toggle", and so on?

PR #451, opened 2017-07-12
Batch of fixes from the Lancaster doc review

ganicke:


ganicke:
Re: https://github.com/Microsoft/pxt/issues/2416

libs/core/basic.cpp
```diff
@@ -85,6 +85,8 @@ namespace basic {
      * @param interval time in milliseconds between each redraw
      */
     //% help=basic/show-animation imageLiteral=1 async
+    //% blockId=device_show_animation 
+    //% block="show animation"
```
pelikhan: This API is not well supported in blocks. Needs testing.

ganicke: Ok, I'll take block support off. Leave it or remove from the basic.md card section?

pelikhan:
Leave it, works in TypeScript

PR #455, opened 2017-07-19
Improve Windows Chrome upload help

HelenFoster:
The Windows half of https://github.com/Microsoft/pxt/issues/2465

Had to reorder some of the other instructions to have this make sense.

### DIFF
--- a/docs/device/usb.md
+++ b/docs/device/usb.md
@@ -17,14 +17,6 @@ directly into the MICROBIT drive!
 
 ### ~
 
-### ~ hint
-
-In Chrome, you can change the **Downloads** settings to let you choose where
-to save each file. Then you can save the ``.hex`` file
-directly into the MICROBIT drive.
-
-### ~
-
 ![](/static/mb/device/usb-thin.jpg)
 
 ## Instructions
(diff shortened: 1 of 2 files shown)


docs/device/usb.md
```diff
@@ -17,6 +17,14 @@ directly into the MICROBIT drive!
 
 ### ~
 
+### ~ hint
+
```
pelikhan: It sure we want this here. There are various different hints for each browser and it will overload this page.

pelikhan:
Try to use the macro @boardname@ instead of micro:bit . Thanks for improving our docs!

docs/device/usb.md
HelenFoster: Should we delete the other one as well then?

HelenFoster:
> Try to use the macro @boardname@ instead of micro:bit

Mostly it's talking about the MICROBIT drive on the computer, so is there anything for that?

I didn't change much of the text in this document - it just looks like a lot of changes because it's in a different order.

ganicke:
I suggest removing the hint(s) since the same info is in the browser specific pages. Try to connect the **Instructions** section to step (3) of "basic steps". I have a suggested replacement here: https://github.com/Microsoft/pxt-microbit/commit/20b385dbbe7b87408fa8531c146492e40a2041ab.

HelenFoster:
I agree with @ganicke

PR #459, opened 2017-07-21
Stop deleted sprites touching others

HelenFoster:
Fixes https://github.com/Microsoft/pxt/issues/2466

### DIFF
--- a/libs/core/game.ts
+++ b/libs/core/game.ts
@@ -644,7 +644,7 @@ namespace game {
         }
 
         /**
-         * Deletes the sprite from the game engine. All further operation of the sprite will not have any effect.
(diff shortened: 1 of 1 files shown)


pelikhan:
What other API need to use the enabled bit?

HelenFoster:
> What other API need to use the enabled bit?

I think there isn't any at the moment (as I discussed in the issue).

pelikhan:
Do you have projects using game APIs? We take contributions for those too :)

HelenFoster:
> Do you have projects using game APIs? We take contributions for those too :)

I've been pondering what to do about that. This isn't really the place to discuss it, so what do you suggest?

pelikhan:
Are you the microbit slack channel? We can use the #pxt channel there

PR #464, opened 2017-07-27
Game library: don't plot if there are no sprites

HelenFoster:
The more straightforward half of https://github.com/Microsoft/pxt/issues/2468

### DIFF
--- a/libs/core/game.ts
+++ b/libs/core/game.ts
@@ -651,8 +651,13 @@ namespace game {
         //% blockId="game_delete_sprite" block="delete %this"
         public delete(): void {
             this._enabled = false;
-            if (sprites.removeElement(this))
+            if (sprites.removeElement(this)) {
                 plot();
+                if (sprites.length == 0) {
+                    //Last sprite deleted, so plot is now disabled
+                    basic.clearScreen();
+                }
+            }
         }
 
         /**
@@ -723,7 +728,7 @@ namespace game {
      */
     //% parts="ledmatrix"
     function plot(): void {
-        if (game.isGameOver()) {
+        if (game.isGameOver() || sprites.length == 0) {
             return;
         }
         let now = input.runningTime();


libs/core/game.ts
```diff
@@ -723,7 +728,7 @@ namespace game {
      */
     //% parts="ledmatrix"
     function plot(): void {
-        if (game.isGameOver()) {
+        if (game.isGameOver() || sprites.length == 0) {
```
pelikhan: call basic.clearScreen here

libs/core/game.ts
```diff
@@ -651,8 +651,13 @@ namespace game {
         //% blockId="game_delete_sprite" block="delete %this"
         public delete(): void {
             this._enabled = false;
-            if (sprites.removeElement(this))
+            if (sprites.removeElement(this)) {
                 plot();
+                if (sprites.length == 0) {
```
pelikhan: That is the responsibility of plot

libs/core/game.ts
HelenFoster: The point of this change is to call `clearScreen` only at the time the last sprite is deleted, and not on subsequent calls to `plot`. So to put it in `plot`, I'd have to create an extra flag or give `plot` an optional parameter instructing it to do this.

pelikhan:
plot is responsible for rendering the state of the game engine. So it should clear the screen since the user might have messed around with the screen.

Are you looking for a way to tell the game engine to "pause" while other animations are running?

HelenFoster:
> plot is responsible for rendering the state of the game engine. So it should clear the screen since the user might have messed around with the screen.

The game engine starts automatically when the first sprite is created, so I thought expected behaviour would be for it to go away when the last sprite is deleted. If the user deletes the last sprite and then messes around with the screen, they almost certainly don't intend to be using the game engine at the moment.

> Are you looking for a way to tell the game engine to "pause" while other animations are running?

I guess we could do that instead.

HelenFoster:
I don't think it's a good idea to change the existing behaviour for while the game engine has active sprites, as people might be relying on that.

Maybe add `pause` and `resume` functions to the game engine? They would need to be available in blocks.

pelikhan:
Yes I was thinking about pause, resume being better semantics.

HelenFoster:
Hmm, I'm not sure about the naming. Do you think game.pause would be confusing with basic.pause? Want to express "Make the game engine stop drawing to the screen."

HelenFoster:
It looks like "(game controller thing) pause" so maybe that's OK, as people would recognise that combination.

game_pause

pelikhan:
See my PR https://github.com/Microsoft/pxt-microbit/pull/476

HelenFoster:
Great!

PR #466, opened 2017-07-27
[Accessibility] Color contrast fixed

veler:
Fixed the color contrast of green buttons.

Also added a outline color for the focus on the simulator and changed the color of the green text in the documentation.

capture

capture

Related issues : [https://github.com/Microsoft/pxt/issues/2535](https://github.com/Microsoft/pxt/issues/2535), [https://github.com/Microsoft/pxt/issues/2550](https://github.com/Microsoft/pxt/issues/2550), [https://github.com/Microsoft/pxt/issues/2528](https://github.com/Microsoft/pxt/issues/2528), [https://github.com/Microsoft/pxt/issues/2550](https://github.com/Microsoft/pxt/issues/2550), [https://github.com/Microsoft/pxt/issues/2519](https://github.com/Microsoft/pxt/issues/2519)


microsoftsam:
Which one is before? and which is after? 
Generally before is left and after is right, but I don't think thats the case here?

microsoftsam:
How did you pick this shade of green? We need to run this through UI review with David and @Jaqster as you're changing the entire site's green color.

microsoftsam:
Also for anyone reviewing, the previous green is the micro:bit foundation green. Since this is not a primary colour for us, we could switch to using the Microsoft green.

veler:
I see.
Yes the "before" is on right, sorry about that.

I picked up a color from a palette. I knew it would probably not be satisfying. This one pass the contrast test, but we can change it or cancel of course.

PR #471, opened 2017-07-31
[Accessibility] Fixed an issue where the accessibility of the Pin was broken and was breaking the simulator.

veler:
Fixed an issue where while using only the A, B button and Pin (but no thermometer or light level) in the microbit simulator, a space/padding was moving the simulator.
After fixing this issue, the tag used as an aria-live was visible, so I hidden it too.
Finally, I fixed an issue where the simulator's UI was not correctly displayed at loading on FireFox while using only the A, B button and Pin (but no thermometer or light level). It had an impact on tabbing into the Pin.

This fix is linked to pxt-core [https://github.com/Microsoft/pxt/pull/2591](https://github.com/Microsoft/pxt/pull/2591)

### DIFF
--- a/sim/state/edgeconnector.ts
+++ b/sim/state/edgeconnector.ts
@@ -3,6 +3,7 @@ namespace pxsim.input {
         let pin = getPin(pinId);
         if (!pin) return;
         pin.isTouched();
+        runtime.queueDisplayUpdate();
         pxtcore.registerWithDal(pin.id, DAL.MICROBIT_BUTTON_EVT_CLICK, handler);
     }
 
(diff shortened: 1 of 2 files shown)


sim/state/edgeconnector.ts
```diff
@@ -3,13 +3,15 @@ namespace pxsim.input {
         let pin = getPin(pinId);
         if (!pin) return;
         pin.isTouched();
+        runtime.queueDisplayUpdate();
         pxtcore.registerWithDal(pin.id, DAL.MICROBIT_BUTTON_EVT_CLICK, handler);
```
microsoftsam: Cherry pick these two line changes to pxt-microbit (master)

PR #473, opened 2017-07-31
Variables updates

dkiang:
Miscellaneous corrections.


docs/courses/csintro/variables/project.md
```diff
@@ -108,7 +108,7 @@ Have students write a reflection of about 150–300 words, addressing the follow
 **2 =** Uses a mathematical operation incorrectly or not in a way that is integral to the program.<br/>
 **1 =** No mathematical operations are used.
 
-### micro:bit program
```
pelikhan: micro:bit is actually lower cased.

pelikhan: Use @boardname@ instead

PR #484, opened 2017-08-01
Add eventTimestamp and eventValue to pxsim.control

caitlin-hennessy:
In conjunction with PR https://github.com/Microsoft/pxt/pull/2615, resolves issue https://github.com/Microsoft/pxt/issues/395.

### DIFF
--- a/sim/state/misc.ts
+++ b/sim/state/misc.ts
@@ -65,6 +65,14 @@ namespace pxsim.control {
         // TODO mode?
         board().bus.queue(id, evid)
     }
+
+    export function eventTimestamp() {
+        return board().bus.getLastEventTime()
+    }
+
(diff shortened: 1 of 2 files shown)


sim/state/misc.ts
```diff
@@ -65,6 +65,14 @@ namespace pxsim.control {
         // TODO mode?
         board().bus.queue(id, evid)
     }
+
+    export function eventTimestamp() {
+        return board().bus.getLastEventTime()
+    }
+
+    export function eventValue() {
+        return board().bus.getLastEvent()
```
pelikhan: should this be getLastEventValue?

caitlin-hennessy: Yes, actually...it probably should.

PR #485, opened 2017-08-02
Add 'user-select: none' to sim elements that shouldn't be draggable

caitlin-hennessy:
Resolves https://github.com/Microsoft/pxt/issues/1557.


microsoftsam:
Can we apply this in pxt-adafruit as well.

PR #498, opened 2017-08-07
[Accessibility] Updated the documentation to make the heading MAS compliant.

veler:
Related issues : [https://github.com/Microsoft/pxt/issues/2547](https://github.com/Microsoft/pxt/issues/2547), [https://github.com/Microsoft/pxt/issues/2546](https://github.com/Microsoft/pxt/issues/2546)

untitled


docs/courses/csintro/booleans/project.md
```diff
@@ -11,14 +11,14 @@ Remind the students of all the different inputs available to them through the mi
  
 ## Project Ideas
 
-### Sunscreen Monitor
```
ganicke: This is a sub-heading of ``## Project ideas`` and should remain as a 3rd level heading. This wouldn't break the ``##`` to ``###`` heading progression for acc would it?

microsoftsam:
@ganicke this change will have to be applied everywhere in ours docs that we skip the h2 heading. 

Could you take a final look and approve?

PR #500, opened 2017-08-09
Colour namespace blocks (pxt-microbit changes)

microsoftsam:
Add respective namespaces to the authoring of inline blocks and buttons

Associated with Microsoft/pxt#2686

### DIFF
--- a/docs/projects/flashing-heart.md
+++ b/docs/projects/flashing-heart.md
@@ -2,7 +2,7 @@
 
 ## Step 1
 
-Place the ``||show leds||`` block and paint a heart.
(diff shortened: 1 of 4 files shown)


ganicke:
Would be nice if we could do some styling like:

blockns

ganicke:
Sorry, give me a chance to pull your ```V0``` and see what it looks like.

microsoftsam:
I think there are mixed opinions about the zero border on the right, we can bring this up at our next UI meeting.
As for the authoring, it's a simpler regex and less error prone for it to be like so: 
``||basic:show leds||``

PR #508, opened 2017-08-16
Updated the call to enableKeyboardInteraction.

veler:
Updated the call to enableKeyboardInteraction. BEFORE merging this PR, we MUST merge the following PR from PXT-core : [https://github.com/Microsoft/pxt/pull/2736](https://github.com/Microsoft/pxt/pull/2736)
Without it, the current PR will not build.


microsoftsam:
use undefined. null is a value in JS.

veler:
Oops sorry, I know it furthermore. I changed it.

PR #517, opened 2017-08-24
Fixed the color contrast of the side menu

veler:
Fixed the luminosity of the menu items in the side menu.

Related issue : [https://github.com/Microsoft/pxt/issues/2797](https://github.com/Microsoft/pxt/issues/2797)

untitled

### DIFF
--- a/theme/site/globals/site.variables
+++ b/theme/site/globals/site.variables
@@ -41,6 +41,8 @@
 
 @tutorialSelectedMenuColor: @blue;
 
+@invertedUnselectedTextColor : rgba(255, 255, 255, 0.8);
+
 /*-------------------
    Layout
 --------------------*/


microsoftsam:
Awesome!

microsoftsam:
This is in site.variables, does it affect all inverted text colors? ie: inverted buttons, etc.

veler:
No, only the sidebar and the footer in the documentation, or at least I didn't found any change somewhere else.

microsoftsam:
Do we need this change here? if PXT can do it?

microsoftsam:
See where this variable is being used: https://github.com/Semantic-Org/Semantic-UI-LESS/search?utf8=%E2%9C%93&q=invertedUnselectedTextColor&type=

veler:
So, I just did a test : if I do this change in PXT, it applies the change to the side menu in the documentation and the footer of the documentation, but that's all. I didn't found any change while running PXT and PXT-Microbit : I checked the index page of MakeCode, the editor, the modals, the contextual menu, the tutorial and the high contrast.

So as you have more experience, I would say it's up to you. The change has the same visual impact.
But maybe it's a better idea to put it in PXT instead of PXT-Microbit so the other targets will enjoy the change. What do you think about?

veler:
I'm closing this PR as I opened one for the same change but on pxt.

PR #521, opened 2017-08-29
Typo in Experiment section

gbowman85:
Changed press B to press A. Value close to 1000 should be for wet dirt.


pelikhan:
Thank you for this PR!

PR #533, opened 2017-09-11
Fix accessibility sim regression when setting pin ARIA text. 

microsoftsam:
Fixes pxt#2894


sim/visuals/microbit.ts
```diff
@@ -477,10 +477,10 @@ namespace pxsim.visuals {
                 accessibility.makeFocusable(this.pins[index]);
                 accessibility.setAria(this.pins[index], "slider", this.pins[index].firstChild.textContent);
                 this.pins[index].setAttribute("aria-valuemin", "0");
-                this.pins[index].setAttribute("aria-valuemax", "1023");
+                this.pins[index].setAttribute("aria-valuemax", pin.mode & PinFlags.Analog ? "1023" : "100");
```
pelikhan: 100? For digital, it should be 1.

PR #535, opened 2017-09-11
Add a page for MES events.

ganicke:
Collection of MES events with short descriptions.

Addresses https://github.com/Microsoft/pxt/issues/1340.


pelikhan:
Could you add a link to the microbit-dal C++ source file? LGTM.

PR #536, opened 2017-09-12
[Accessibility] Fixed the color contrast issue in the documentation

veler:
Fixed an issue where the color contrast of the green text in the documentation was not correct.

Related issue : [https://github.com/Microsoft/pxt/issues/2528](https://github.com/Microsoft/pxt/issues/2528)
Related PR : [https://github.com/Microsoft/pxt/pull/2928](https://github.com/Microsoft/pxt/pull/2928)

For some reason, this issue was fixed by the past but has been reverted at a moment. I think it's when we moved some changes to PXT instead of PXT-microbit.

untitled

### DIFF
--- a/theme/site/globals/site.variables
+++ b/theme/site/globals/site.variables
@@ -20,6 +20,8 @@
 @grey: #95a5a6;
 @green: #107C10;
 
+@greenTextColor : #126B1B;
+
 /* Microbit.org colors */
 @orange: #ff8b27;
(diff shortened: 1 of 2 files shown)


microsoftsam:
Checking into master? before we've made the change in v0?

microsoftsam:
Oh nevermind, this is pxt-microbit.

microsoftsam:
This needs to be applied in all targets.

PR #538, opened 2017-09-12
Add firmware update note and page.

ganicke:
Addresses https://github.com/Microsoft/pxt/issues/2896.

- [x] Small note in _/device/serial.md_ mentioning that a firmware upgrade might be needed to get the serial port to show up.
- [x] Added _/device/firmware.md_ as a direct to the microbit.org support page for the update.

I can add more specific firmware update instructions to the new **firmware** page if necessary. The microbit.org page explains this fairly well though.

### DIFF
--- a/docs/device/firmware.md
+++ b/docs/device/firmware.md
@@ -1,0 +1,11 @@
+# Firmware
+
+Occasionally, changes and improvements to the system software (firmware) on the @boardname@ are needed. Updates to the firmware are made by the [micro:bit Education Foundation](http://microbit.org/about).
+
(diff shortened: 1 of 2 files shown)


pelikhan:
@jaustin All info about firmware should be under microbit.org/firmware and not maintained in our docs -- otherwise it'll grow out of date.

ganicke:
Pull out the _firmware.md_ and just link generally to https://support.microbit.org/support/solutions?

pelikhan:
We wait till the foundation has a dedicated page IMO

PR #546, opened 2017-10-05
Video and instructions do not match

stephenhowell:
Video shows 3V connected to clip, text instructions say GND.


pelikhan:
Thanks!

PR #553, opened 2017-10-16
'Connect your device' notification

caitlin-hennessy:
Depends on https://github.com/Microsoft/pxt/pull/3256

### DIFF
--- a/editor/extension.ts
+++ b/editor/extension.ts
@@ -222,13 +222,9 @@ namespace pxt.editor {
         })
     }
 
-    export function deployCoreAsync(resp: pxtc.CompileResult, isCli = false): Promise<void> {
+    export function deployCoreAsync(resp: pxtc.CompileResult, d: pxt.commands.DeployOptions = {}): Promise<void> {
         let saveHexAsync = () => {
-            if (isCli) {
-                return Promise.resolve()
-            } else {
-                return pxt.commands.saveOnlyAsync(resp)
-            }
+            return pxt.commands.saveOnlyAsync(resp)
         }
 
         startTime = 0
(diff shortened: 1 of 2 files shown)


editor/extension.ts
```diff
@@ -331,8 +327,16 @@ namespace pxt.editor {
                         return wrap.cortexM.reset(false)
                     })
             })
-            .catch(e => {
-                return saveHexAsync();
+            .catch(e => {  
+                if (e.description === "Device not found" && d.reportError) {
```
guillaumejenkins: In pxtwinrt, when this error is thrown, are we using `lf()`? If so, then we'll need to check for the translated string

It might be cleaner to add a language-independent `code` or `type` property in the error object and use that instead

editor/extension.ts
```diff
@@ -331,8 +327,16 @@ namespace pxt.editor {
...
+                    if (d.reportError) {
+                        d.reportError(Util.lf("Connect your device."))
```
guillaumejenkins: Should we add a `Please` at the start? :)

caitlin-hennessy: Sure :)

editor/extension.ts
caitlin-hennessy: Done.

editor/extension.ts
```diff
@@ -331,8 +327,16 @@ namespace pxt.editor {
...
                     })
             })
-            .catch(e => {
-                return saveHexAsync();
+            .catch(e => {  
+                if (e.type === "devicenotfound" && d.reportError) {
+                    if (d.reportError) {
+                        d.reportError(Util.lf("Please connect your device."))
+                    } else {
+                        console.error(e)
```
pelikhan: Use PXT.reportError

editor/extension.ts
```diff
@@ -331,8 +327,16 @@ namespace pxt.editor {
...
+                    }
+                } else {
+                    saveHexAsync()
```
pelikhan: Missing return?

editor/extension.ts
caitlin-hennessy: Done, will be in next PR

caitlin-hennessy:
Done, will be in next PR

editor/extension.ts
caitlin-hennessy: Done, will be in next PR

PR #561, opened 2017-11-07
Updating a point fix for issues/3399

abchatra:
Changing serial parameter name to serialNumber https://github.com/Microsoft/pxt/issues/3399

### DIFF
--- a/docs/projects/rps-teams.md
+++ b/docs/projects/rps-teams.md
@@ -103,7 +103,7 @@ let players: number[] = [0]
 ## Step 4: Receiving a message (part 1)
 
 In an the ``|on radio received|`` event, we receive the status from another @boardname@. Click on the **gearwheel*
-to add the ``serial`` parameter as we will need it to identify who sent that packet.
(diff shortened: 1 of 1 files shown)


docs/projects/rps-teams.md
```diff
@@ -117,9 +117,9 @@ let player_index = 0
 let players: number[] = [0]
 let weapon = 0
 let found = false
-radio.onDataPacketReceived(({ receivedNumber, serial }) => {
+radio.onDataPacketReceived(({ receivedNumber, serialNumber }) => {
```
pelikhan: Does this compile?

abchatra: Nope needs serial: serialNumber

abchatra: I will fix

abchatra: Users will still have to rename their variable from the gear box. I am going to abandon this fix for the right fix from Richard.

PR #576, opened 2017-11-29
Add sim mentions to the R-P-S HoC project

ganicke:


ganicke:
Take #579 instead of this one now.

PR #591, opened 2017-12-06
Fixing documentaiton of troubleshoot to

abchatra:
Give a logical jump for folks with right cable or firmware


docs/device/windows-app/troubleshoot.md
```diff
@@ -1,4 +1,4 @@
-# Troubleshooting downloads from the Windows 10 App
+ Troubleshooting downloads from the Windows 10 App
```
pelikhan: Title 1 needed

abchatra: That was not intentional

PR #592, opened 2017-12-06
Fix for the documentation firmware download

abchatra:


abchatra:
This doesn't work on Edge

PR #601, opened 2017-12-11
Buffer topics

ganicke:
First cut at some buffer topics. **Buffer** is not mapeped to a namespace so I put the methods in a one-pager like the **Math** page. I could break them out into separate topics if that makes more sense.

Also, i'll need to add the i2cRead/WriteBuffer api pages.

RE: https://github.com/Microsoft/pxt/issues/3594


carlosperate:
Wow, I'm very glad to see the buffers getting documented in so much details, thanks!

DavidWhaleMEF:
This is super work, thanks!

PR #603, opened 2017-12-12
adding turtle examples

pelikhan:
Requires pushing beta to master


thinkberg:
looks like this must be updated to merge

PR #616, opened 2018-01-30
Add sim note to 'reset' api topic

ganicke:
Fixes https://github.com/Microsoft/pxt/issues/3781.


pelikhan:
We should fix the simulator really.

ganicke:
Ok, pick it back to v0?

ganicke:
Merge this one. Then, link to this PR/Issue from the sim fix so the note gets removed.

PR #630, opened 2018-02-06
updated api translations

pelikhan:
Fresh download of API translations from Crowdin. Required for offline work.
Issues: these translations are currently added to the bundled projects, which gets compiled into the target.js file... which is going to grow quite a bit.


abchatra:
Do we need to add all these files to target repo? Why can't this be part of build?

pelikhan:
It takes 10min to download from crowdin

PR #651, opened 2018-02-15
Fix brightness led visuals offset bug in simulator

samelhusseini:
Fix led brightness visuals where it's offset on latest Chrome, Firefox and Edge. 
Fix was to add an extra attribute "transform-box" equal to "fill-box". 

Fixes https://github.com/Microsoft/pxt/issues/3702
Fixes https://github.com/Microsoft/pxt/issues/2700


abchatra:
Works in safari as well?

samelhusseini:
Safari had no issue but tested on: 
- Chrome 64, Edge 16, IE 11, Firefox 58 (Windows) and Safari 11 (Mac).

PR #656, opened 2018-02-16
Add 'soil moisture' lesson to science exp

ganicke:
This lesson has an external data collection worksheet in Google Docs. Convert to MD?


pelikhan:
Drop it for now. We need to think about this.

ganicke:
Right.

Issue #648, opened 2018-02-15
W10 App: Soft-bricking of micro:bit when flashing over Python program


jaustin:
Steps to reproduce:
* Go to python.microbit.org and flash a 'hello world' program to your micro:bit
* Open with Windows 10 app
* Flash the micro:bit using the W10 app, any simple icon will do

Expected behaviour:
* Icon is shown

Actual Behaviour:
* micro:bit appears dead. Flashing other programs with the app doesn't work

You can recover your micro:bit by using mass storage Drag and Drop

Here's a hex file from a 'broken' micro:bit
[out.zip](https://github.com/Microsoft/pxt-microbit/files/1728648/out.zip)

pelikhan:
I can repro.

mmoskal:
This commit https://github.com/Microsoft/pxt-microbit/commit/161444b5d359e6a9c6bf27b7ce6a9de67c7ba78c fails the flashing if we detect problems with UICR; we should be more explicit in asking the user to just drag&drop the file and everything will work the next time.

@jaustin if you have someone who could look into actually fixing this - please go ahead. It seems the problem is writing to the UICR - probably the flash-writing code needs to be updated

mmoskal:
It's also possible that codal require some memory to be erased, when it's not.

jaustin:
@mmoskal thanks for the workaround. I'm guessing we're hitting problems because the UICR can only be erased with a full erase?

8.1 of the nRF51 Reference Manual
Code readback protection of the whole code area, or a part of the code area can be configured and enabled in the UICR. The UICR can only be erased by using ERASEALL.

pelikhan:
The fix is in beta. I've also added a 5th step in the troubleshoot to download a .hex file into the MICROBIT drive.

jaustin:
@pelikhan is there a fix beyond this workaround? If not can we spawn a ticket for a real fix to keep track of it?

pelikhan:
The fix does not appear to be on our side.

mmoskal:
This is a corner case with an easy workaround. 

The fix would he in extension.ts and certainly in reach of someone with DAPLink expertise. PRs are welcome.

mmoskal:
Actually, it would be probably best if the partial flashing code moved into dapjs. Then this is also where the fix would be. 

[...]

PR #659, opened 2018-02-16
Complain if flashing cannot succeed; see #648

pelikhan:


mmoskal:
I don't think anyone ever sees this message. On localhost we pop up download dialog, which I guess might be good enough. Does this also happen in the app?

pelikhan:
Correct, on the app nobody sees the message. Let me see what i can do.

PR #662, opened 2018-02-17
Add 'magnetism' lesson

ganicke:
Looks like Carl hasn't finished this lesson yet. 'Setup and procedure' is still in skeleton form. We can push and leave the front page link cold. BTW, I made the magnet picture so it's 'original'.


pelikhan:
Let's wait.

pelikhan:
@ganicke could you remove it for now?

ganicke:
Yea, we'll drop it. I could check with Carl again and see if he wants to finish his version. If so, I'll add it back in.

PR #667, opened 2018-02-20
include ws2812 ASM to avoid replication in packages

pelikhan:
Adds the necessary ASM to drive a strip of neopixels. Allows 3rd party to wrap it up in their products.
@mmoskal should/could we disable interrupts when we know that BLE is active?


pelikhan:
@finneyj @jaustin

jaustin:
@finneyj do we think this has any new feature interactions we need to be worried about?

jaustin:
...Although it's a bit more work, one advantage of bringing this into C/C++ would be that we could use sd_nvic_critical_region_enter() / _exit() if we're running on top of SD instead of the CPSID/E - I don't *think* we have the tools do to that from TS?

mmoskal:
@jaustin did you test `sd_nvic_critical_region_enter()`? Does it disable all interrupts? If we get any interrupt, the colors are completely out of sync.

jaustin:
@mmoskal no, not tested - but I think it guarantees that the colours get out of sync rather than the radio. Users who have combined BLE with neopixel report colours stuttering, but not getting way out, just distortion or flicker. Peli said you'd seen hard crashes, though?

pelikhan:
I could move this into the radio package. The goal here is to mitigate the proliferation of ws2812 drivers.

abchatra:
I am not sure how ws2812 API belongs to radio (other than both can't work with ble). I also can't think of a better solution  😞.

pelikhan:
It prevents it from Interrferrig with BLE, allows vendors to use and avoids adding a new package in the “add package” dialog.

pelikhan:
We're going to move this code into a github repo

pelikhan:
ASM moved to https://github.com/microsoft/pxt-ws2812

PR #669, opened 2018-02-21
Clearer message for devices blocks

pelikhan:
Send docs from 'devices' apis to https://microbit.org/guide/mobile/


pelikhan:
@jaustin FYI

PR #678, opened 2018-02-26
override hid deploy if can hid

pelikhan:
Don't try hid if no hid avaiable.

### DIFF
--- a/editor/extension.ts
+++ b/editor/extension.ts
@@ -116,18 +116,21 @@ namespace pxt.editor {
             })
     }
 
-    function initAsync() {
-        let canHID = false
+    function canHID(): boolean {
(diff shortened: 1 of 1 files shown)


editor/extension.ts
```diff
@@ -401,7 +404,9 @@ namespace pxt.editor {
                         .then(text => project.overrideTypescriptFile(text))
```
guillaumejenkins: I know it's not related to your changes but `text => project.overrideTypescriptFile(text)` is never called since the previous promise is always rejected

The rejection should come after the call to `project.overrideTypescriptFile(text)`

pelikhan: Can you open another PR for this?

PR #680, opened 2018-02-27
Update README.md

gbaman:
Update the readme.md file to include current docs for local development after long conversation with @pelikhan.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,34 +15,74 @@ Please add an issue if you discover an (unreported) bug.
 
 The local server lets you to run the editor and serve the documentation from your own computer.
 
-### Setup
+### Setup   
+This section explains how to setup your development environment for pxt-microbit, whether that be to run a copy locally or to make source changes.    
+Note that pxt-microbit can not be run without its main depandancy, [pxt](https://github.com/Microsoft/pxt). Below explains how to setup both.      
 
 The following commands perform a one-time setup after synching the repo on your machine.
+* Note for any editing of the .cpp files, Yotto must be installed. Do do so, follow the instructions on [their site](http://docs.yottabuild.org/).
+* Install node.js 8.9.4 or higher.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,84 +15,85 @@ Please add an issue if you discover an (unreported) bug.
...
-```
-* install the microbit target
-```
-pxt target microbit
-```
-
-### Running
-
-Run this command to open a local web server (add ``sudo`` for Mac/Linux shells)
+* Note for any editing of the .cpp files, Yotto must be installed. Do do so, follow the instructions on [their site](http://docs.yottabuild.org/).
```
pelikhan: yotta

README.md
```diff
@@ -15,84 +15,85 @@ Please add an issue if you discover an (unreported) bug.
...
+
+Run this command from inside pxt-microbit to open a local web server (add ``sudo`` for Mac/Linux shells)
+```
+pxt serve --cloud
 ```
+If the local server opens in the wrong browser, make sure to copy the URL containing the local token. 
+Otherwise, the editor will not be able to load the projects.
 
-Both of those repoes are now ready to go. To start your local server, run
+If you need modify the `.cpp` files (and have installed yotta), enable yotta compilation by removing the ```--cloud``` flag (add ``sudo`` for Mac/Linux shells):
```
pelikhan: Use ``--cloud`` to skip the local yotta compilation.

gbaman: Is that not what the above is saying, but the other way around?
As in, probably better for most folks to be starting with --cloud, then remove it if you do need to modify the .cpp files?

README.md
gbaman: Haha, that one is a mixture of Yotta and Yocto apparently!
Used to do some Yocto stuff with the Intel Galileo/Edison boards.

gbaman: Fixed

README.md
pelikhan: Right my bad.

PR #701, opened 2018-03-08
Update scroll-image.md

punker76:
Just change the documentation here, cause the underlying API doesn't work correctly yet https://github.com/lancaster-university/microbit-dal/blob/2cff906f019944ae15d532157c1301c940f67862/source/drivers/MicroBitDisplay.cpp#L931
So negative values are not working for now.


pelikhan:
Negative scrolling is supported on the hardware. We need to fix the simulator to support that as well.

pelikhan:
Actually you're right, the C++ implementation does not support scrolling images backward.

PR #708, opened 2018-03-14
Data analysis topics

ganicke:
This is a group of pages to replace the _data-logging.md_ page. Let me know what features need mention. I'm calling the chart and console view as the "Data Viewer". What are we calling it for reals?


docs/device/data-analysis/generating.md
```diff
@@ -0,0 +1,111 @@
+# Generating data
+
+## Sensor values
```
pelikhan: Thé simplest to generate data is ´´led.plotBarGraph’´.

ganicke: I got it at the end of the 'writing data' topic. I'll move it up to the top if you think it should go there.

ganicke:
Oh, yea. I can do something for remote data collection.

docs/device/data-analysis/generating.md
pelikhan: It should be their go to block for Data logging.

ganicke: Ok, I'll re-cast it as the first method to use.

PR #710, opened 2018-03-14
updated Tinkertanker repo names

microbit-mark:


pelikhan:
I noticed that 'TinkerTanker' casing is different in targetconfig.json.

PR #773, opened 2018-05-14
Add radio methods to replace the radio mutator

samelhusseini:
The radio (onDataReceived) method has a mutator dialog (cog wheel) and we're moving away from that in v3 of PXT.

and to give you an idea of how bad it is (using the object destructuring model, ie the current model), if you look at the generated code: 

```
radio.onDataPacketReceived( ({ receivedNumber }) =>  {
	
})
```
Kids struggle with knowing what to put between the { } in the handler. 

The proposal is to move away from all that and just create three new blocks / functions: 

onReceivedString (string)
onReceivedNumber (number)
onReceivedValue (key, value)

Looks like this: 
screen shot 2018-05-14 at 11 03 27 am

and the JS is simply:
```
radio.onReceivedNumber(function (receivedNumber) {
	
})
radio.onReceivedValue(function (name, value) {
	
})
radio.onReceivedString(function (receivedString) {
	
})
```

In order to retrieve the last packet's time, serialnumber or signal.
we add a new block to retrieve the last's packets data: 

screen shot 2018-05-14 at 11 17 28 am

I think this is much more intrusive to use, and matches what we do when sending the value (ie: there's three functions sendNumber, sendValue and sendString)

@pelikhan @jaustin thoughts?


libs/radio/radio.ts
```diff
@@ -33,6 +33,16 @@ namespace radio {
         public signal: number;
     }
 
+    export enum PacketProperty {
+        //% blockIdentity=radio._packetProperty
+        time,
```
pelikhan: capitalize TS, lower case blocks.

libs/radio/radio.ts
```diff
@@ -33,6 +33,16 @@ namespace radio {
...
+        //% block="serial number"
+        //% blockIdentity=radio._packetProperty
+        serialNumber,
+        //% blockIdentity=radio._packetProperty
+        signal
```
pelikhan: let's name it signalStrength

libs/radio/radio.ts
```diff
@@ -52,63 +62,111 @@ namespace radio {
...
+    //% blockId=radio_on_buffer block="on radio received" blockGap=8
+    export function onReceivedBuffer(cb: (buffer: Buffer) => void) {
         onDataReceived(() => {
             receiveNumber();
             const packet = new Packet();
             packet.receivedNumber = receivedNumber();
             packet.time = receivedTime();
             packet.serial = receivedSerial();
-            packet.signal = receivedSignalStrength();
             packet.receivedString = receivedString();
```
pelikhan: not needed

pelikhan:
Could we add some patching rules?

PR #778, opened 2018-05-17
Update miniproject.md

inrhapsody:


pelikhan:
Translation should be done in Crowdin.

PR #835, opened 2018-06-01
Implement idiv for v1

samelhusseini:
Is this the right way to do this? @pelikhan


mmoskal:
This isn't needed - the operation is already on integers. It should work as is.

PR #854, opened 2018-06-04
buffer data vs payload app compat fix

pelikhan:
Adding macro to access "data" field.

### DIFF
--- a/libs/core/pxtbase.h
+++ b/libs/core/pxtbase.h
@@ -7,6 +7,11 @@
 #pragma GCC diagnostic ignored "-Wformat"
 #pragma GCC diagnostic ignored "-Warray-bounds"
 
+// cross version compatible way of access data field
+#ifndef PXT_BUFFER_DATA
+#define PXT_BUFFER_DATA(buffer) buffer->data
+#endif
+
 // needed for gcc6; not sure why
 #undef min
 #undef max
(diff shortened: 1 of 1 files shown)


libs/core/pxtbase.h
```diff
@@ -7,6 +7,11 @@
 #pragma GCC diagnostic ignored "-Wformat"
 #pragma GCC diagnostic ignored "-Warray-bounds"
 
+// cross version compatible way of access data field
+#ifndef PXT_BUFFER_DATA
+#define PXT_BUFFER_DATA(buffer) buffer->data
+#endif
+
```
mmoskal: This you will also need to add to `v0`

pelikhan: i can add one in each package that needs it

mmoskal: Right. When you're at it, maybe also add `#define PXT_MICROBIT_V1 true` so you have something to test on in the future

pelikhan: Should I do something like #define PXT_MICROBIT_MAJOR_VERSION 1 ?

mmoskal: well, then you need in the package:
```
#if PXT_MICROBIT_MAJOR_VERSION >= 1
...
```
which again, requires adding `#define PXT_MICROBIT_MAJOR_VERSION 0` to `v0`; but maybe that's the way to go?

pelikhan: adding ```PXT_MICROBIT_TAGGED_INT```

PR #893, opened 2018-06-25
Update flashing.md (soft experiment)

ikunovic:
Just trying to see when and how it will be changed on public page, I would like to write some projects in the near future

### DIFF
--- a/docs/projects/flashing-heart.md
+++ b/docs/projects/flashing-heart.md
@@ -17,7 +17,7 @@ basic.showLeds(`
 
 ## Step 2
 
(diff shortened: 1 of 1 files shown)


docs/projects/flashing-heart.md
```diff
@@ -17,7 +17,7 @@ basic.showLeds(`
 
 ## Step 2
 
-Click ``|Download|`` to transfer your code in your @boardname@!
+Click ``|Download|`` to transfer your code into your @boardname@!
```
ikunovic: just this little experiment

PR #950, opened 2018-07-13
fixed radio signal range.

shaoziyang:


shaoziyang:
How can I update the branch? I am not proficient with git.

PR #975, opened 2018-07-27
Add serial support over webusb

mmoskal:
Depends https://github.com/Microsoft/pxt/pull/4554


mmoskal:
This needs a small fix for microbits with older firmware

mmoskal:
should be good to go - test in Chrome with `?webusb=1` and new firmware

PR #988, opened 2018-07-27
Dal points to dal-integration-1 branch

abchatra:
This is for 1.5 hardware changes to support new sensors.


abchatra:
@jaustin @ finneyj - FYI. /beta will point to dal-integration-1 branch

Issue #942, opened 2018-07-09
Music blocks silent on alpha editor.


xmeow:
All music blocks won't output any sound on Alpha editor after downloading the hex to microbit, but works fine in simulator. 

It is also strange that `pins.analogPitch` works which implement the music blocks.

abchatra:
@xmeow this is a nice find. 

Following code plays gibberish in alpha, but plays fine in the live editor.
```
music.beginMelody(music.builtInMelody(Melodies.Dadadadum), MelodyOptions.ForeverInBackground)
```

riknoll:
Update: The issue is actually in `parseInt()` which the melody blocks use. It always returns NaN on the hardware and it's messing up the parsing of the melody strings

riknoll:
Will be released soon

PR #1022, opened 2018-08-01
Fixing parseInt()

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/942

`parseInt()` was always returning `NaN` for integer strings because our implementation of `strtod` wasn't updating the end pointer


riknoll:
I found it while debugging the melody blocks; I noticed that `NaN` was creeping in somewhere so I suspected it was `parseInt()`

Issue #733, opened 2018-04-10
Bug in radio received signal strength in simulator


DavidWhaleMEF:
The docs for RSSI (signal) for the radio state:

```
signal - How strong the radio signal is from -128 (weak) to -42 (strong).
```

When I run code in the simulator, the signal is returned as 10.

Reproducible test case - press button A and the number 10 is scrolled on screen.

let signal = 0
radio.onDataPacketReceived( ({ receivedString, signal }) =>  {
    basic.showNumber(signal)
})
input.onButtonPressed(Button.A, () => {
    radio.sendString("TEST")
})
basic.showString("?")

![screen shot 2018-04-10 at 11 24 07](https://user-images.githubusercontent.com/25079586/38551671-b55c4e22-3cb1-11e8-9fb3-aa99c5b16000.png)

I would like this to return -42 instead please (strongest signal).

Thanks.

riknoll:
Will be fixed in next release of beta

PR #1024, opened 2018-08-02
Set the sim rssi to a value that makes sense

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/733

This is safe to merge but it won't actually fix the issue until this PR is merged in pxt-core:
https://github.com/Microsoft/pxt/pull/4569


riknoll:
Whoops, meant to make this change in v1. Will redo

Issue #1067, opened 2018-08-09
Beta Makecode Download fails with no errors in output.txt (possibly Trig related)


DaveAtKitronik:
Pressing the download button causes a brief flash of the Compilation Failed, please check your code for errors box, and the download does not proceed. However the Output.txt file reports:

Everything seems fine!

Initially seen in a very long script, but now reproduced in a couple of lines of code.
Enter the following code in the Java Script view (probably also works in blocks but not specifically tried)

let complicatednumber = 0
complicatednumber = Math.cos(1)

Press download, see a pop up message, then no errors.
Using Win10, fails on both Chrome and Edge. 

In the more complex script I have managed to get it to produce an output.txt with some error in (not sure how) that has the following possibly useful info:
error: main.ts(204,32): error TS9200: function not found: cos(...) (shim=Math_::cos)
error: main.ts(205,14): error TS9200: function not found: asin(...) (shim=Math_::asin)
error: main.ts(207,12): error TS9200: function not found: atan(...) (shim=Math_::atan)
error: main.ts(226,19): error TS9200: function not found: atan(...) (shim=Math_::atan)
error: main.ts(230,29): error TS9200: function not found: atan(...) (shim=Math_::atan)
error: main.ts(235,57): error TS9200: function not found: cos(...) (shim=Math_::cos)
error: main.ts(242,58): error TS9200: function not found: cos(...) (shim=Math_::cos)
error: main.ts(250,23): error TS9200: function not found: acos(...) (shim=Math_::acos)
error: main.ts(255,32): error TS9200: function not found: asin(...) (shim=Math_::asin)
error: main.ts(259,33): error TS9200: function not found: asin(...) (shim=Math_::asin)
error: pxt_modules/core/pxt-core.d.ts(362,14): error TS9200: function not found: cos(...) (shim=Math_::cos)
error: pxt_modules/core/pxt-core.d.ts(376,14): error TS9200: function not found: asin(...) (shim=Math_::asin)
error: pxt_modules/core/pxt-core.d.ts(383,14): error TS9200: function not found: acos(...) (shim=Math_::acos)
error: pxt_modules/core/pxt-core.d.ts(390,14): error TS9200: function not found: atan(...) (shim=Math_::atan)

I looks like the Math Trig functions have gone awol for some reason during the download compile stage, but not the run the simulator stage?

guillaumejenkins:
Compile diagnostics:
```
[
    {
        "code": 9200,
        "start": 46,
[...]

PR #1069, opened 2018-08-09
Add missing math functions to cpp layer

guillaumejenkins:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1067


mmoskal:
This has significant impact on code size.

PR #1079, opened 2018-08-13
Remove unwanted Math ops

guillaumejenkins:
Follow up to https://github.com/Microsoft/pxt-microbit/pull/1069/files

Only sin, cos & tan were needed, so remove the rest to minimize code size


mmoskal:
Would be good to measure size impact. Many of these share code, so some might be much cheaper than others.

PR #1082, opened 2018-08-13
Enable "allowPackageExtensions" field in pxtarget.json

rustynymph:
It would be great if the micro:bit target could allow for package extensions in the same way that the adafruit target does.


pelikhan:
We're focusing on migrating v0 to v1. This will have to wait.

pelikhan:
We are working on a feature to enable experimental features. Stay tuned.

pelikhan:
The feature can now be enabled via the Experiments dialog in beta.

Issue #1080, opened 2018-08-13
1st party categories are too close in weight, leaving no room for package categories


CleoQc:
**Describe the bug**
Package loads three categories but they don't show up in the same order with the Beta editor

**To Reproduce**
Load https://github.com/dexterind/pxt-giggle

This is the order that appears in the current editor:
![defaultorder](https://user-images.githubusercontent.com/4965628/44045003-586b99d8-9ef5-11e8-9d66-e80e52eb2465.png)

This is the order in the Beta editor:
![betaorder](https://user-images.githubusercontent.com/4965628/44045054-6e9310b0-9ef5-11e8-9115-1f5992253306.png)

**Expected behavior**
I expected the order to stay the same.

As an aside, can I control this order? I'm happy with the order that shows up in the current editor but did nothing special to get it.

guillaumejenkins:
Category order is determined using the `weight` attributes of the namespaces. It's possible we changed the default logic for ordering categories that are missing a `weight` attribute, but that shouldn't matter if you add `weight` to all your categories so they appear in a deterministic relative order.

`weight` is described in this page: https://makecode.com/defining-blocks, in the **Category** section

Let me know if you run into issues!

CleoQc:
They all have the same weight, 99, so they will fit between Basic (with a weight of 100) and Input (with a weight of 99)
There isn't much room to play with, if I want them visually grouped together.

guillaumejenkins:
Gotcha. I think in that case the best approach is to change the weights of our own categories, to give more room for package categories.

CleoQc:
thank you. That would be a swell idea :)

guillaumejenkins:
@CleoQc this is fixed and will be in the next /beta update (you can test locally with the master branch)

PR #1091, opened 2018-08-14
Slightly adjust non-advanced category weights

guillaumejenkins:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1080

@jaustin @microbit-mark this might impact some existing packages, in that their category might move up or down a few categories in the toolbox. Nothing dramatic, but it might be worth notifying them. In the end, this is a desirable change, as it will give package authors more control on where their categories are placed in the toolbox.

Only the top 4 non-advanced categories were changed.

Category | Weight before | Weight after
---|---|---
Basic | 100 | 116
Input | 99 | 111
Music | 98 | 106
LED | 97 | 101
Bluetooth | 96 | 96 (no change)
Radio | 96 | 96 (no change)

### DIFF
--- a/libs/core/basic.cpp
+++ b/libs/core/basic.cpp
@@ -4,7 +4,7 @@
 /**
  * Provides access to basic micro:bit functionality.
  */
-//% color=#0078D7 weight=100 icon="\uf00a"
+//% color=#0078D7 weight=116 icon="\uf00a"
 namespace basic {
 
     /**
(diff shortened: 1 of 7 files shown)


CleoQc:
Can I ask why Music is that high, higher than LEDs? It might just be my personal experience teaching with the micro:bit but we used LEDs all the time, and music only once. 
Irrelevant to current PR, I'm just being curious.

Issue #1002, opened 2018-07-31
Web USB can hang without alerting the user


microbit-rosslowe:
**Describe the bug**
When transferring a script over Web USB it can hang, causing an infinite spin on the 'Download' button. There appears to be no timeout and the user isn't alerted that the connection has failed, requiring a refresh of the page.

**To Reproduce**
Connect a micro:bit over Web USB and ensure the connection works. Try either of the following:
- Pressing the reset button on the back of the micro:bit
- Dragging & dropping a script manually to the micro:bit
Next time 'Download' is clicked Web USB will hang.

**Expected behavior**
It might be useful to have a notification for this disconnect or a timeout function in the browser. Otherwise the page requires a refresh, which might not be immediately obvious.

**Desktop:**
 - OS: macOS
 - Browser: Chrome 68.0.3440.75
 - Version: 1.1.38

jaustin:
I've just had this happen too - the LED keeps flashing so it seems like there's constant activity. My old (Python) script is still running.

jaustin:
@mmoskal I've just seen another side to this which is that the flash update works but that the LED keeps flashing

For example:
https://www.dropbox.com/s/jzfzz0j3runz7h4/WIN_20180801_12_14_33_Pro.mp4?dl=0

So:
* Pair with webUSB
* Flash the micro:bit
* Code is updated, new code runs
* System LED continues to flash
* Either un/replugging the micro:bit or reloading MakeCode can fix this

https://www.dropbox.com/s/jzfzz0j3runz7h4/WIN_20180801_12_14_33_Pro.mp4?dl=0

mmoskal:
It is just reading the serial.

jaustin:
Why does reading serial toggle the LED? This script doesn't do any serial comms and the behaviour doesn't continue when the editor reloads, so it seems odd? If the micro:bit was streaming some serial data I'd expect that... It also doesn't seem to happen all the time (for example, I don't think @microbit-rosslowe sees this behaviour. Ross?) and the app doesn't do the same either?

microbit-rosslowe:
@jaustin Mine actually does the same, I didn't notice because when the browser window isn't active the rapid flashing stops (but the status LED continues to blink around once per second). Is this the same for Windows?

microbit-rosslowe:
[...]

PR #1108, opened 2018-08-20
Fix flashing after microbit reset

mmoskal:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1002


mmoskal:
It worked for me with hidbridge (localhost) and webusb. I'm still a bit puzzled about the whole thing, but it seems to work.

Issue #1119, opened 2018-08-22
Importing legacy HEX file to new editor loses 'digital pin write' values


microbit-giles:
**Describe the bug**
If you create a program in the old editor that writes a value of 1 to a digital pin, when you download the .hex file and import it into the new beta editor, the value you wrote to the pin is lost.

**To Reproduce**
1. Go to https://makecode.microbit.org/# 
2. Add blocks: **on start / digital write pin P0 to 1**
3. Download .hex file
4. Go to https://makecode.microbit.org/beta#editor
5. Drag .hex file into editor.
6. See that  **on start / digital write pin P0 to 1** has become  **on start / digital write pin P0 to 0**

**Expected behavior**
I expected  **on start / digital write pin P0 to 1** to appear in the beta editor, not  **on start / digital write pin P0 to 0**

**Screenshots**
![digital-pin-fail](https://user-images.githubusercontent.com/42380137/44461501-0266ad00-a609-11e8-97d7-39eb6bb7411c.gif)

**Desktop (please complete the following information):**
 - OS: OS X 
 - Browser Chrome Version 68.0.3440.106 and Safari Version 11.1 (13605.1.33.1.2).
 - OS X Version 10.13.4 - High Sierra

**Smartphone (please complete the following information):**
not tested

**Additional context**
- Bug reproduced by @jaustin who provided screen recording.
- If you use different PIN numbers, eg P1, P2) these **are** retained when opening legacy .hex file in beta editor.

guillaumejenkins:
The issue is that we changed the field name from `NUM` in v0 to `SLIDER` in v1. When opening the project, Blockly looks for a `SLIDER` field in the XML but can't find one, so it adds one with the default number of 0.

guillaumejenkins:
Will be fixed in pxt-microbit v1.1.46

PR #1153, opened 2018-08-30
Upgrade math_number_minmax and other shadow upgrades

guillaumejenkins:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1119

### DIFF
--- a/editor/extension.ts
+++ b/editor/extension.ts
@@ -500,7 +500,6 @@ namespace pxt.editor {
 
         // showleds
         const nodes = U.toArray(dom.querySelectorAll("block[type=device_show_leds]"))
-            .concat(U.toArray(dom.querySelectorAll("shadow[type=device_show_leds]")))
             .concat(U.toArray(dom.querySelectorAll("block[type=device_build_image]")))
             .concat(U.toArray(dom.querySelectorAll("shadow[type=device_build_image]")))
             .concat(U.toArray(dom.querySelectorAll("block[type=device_build_big_image]")))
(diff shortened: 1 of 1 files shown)


editor/extension.ts
```diff
@@ -663,8 +663,22 @@ namespace pxt.editor {
...
                 if (b) b.setAttribute("name", "ARG1");
-            })
+            });
 
+        // device_set_digital_pin
+        U.toArray(dom.querySelectorAll("block[type=device_set_digital_pin]"))
+            .forEach(node => {
+                // Change the name of the NUM field to SLIDER
+                const valueNode = getValue(node, "value");
+                const shadowNode = U.toArray(valueNode.children).filter((n) => n.nodeName === "shadow")[0];
```
riknoll: Doesn't have to be shadow, could also have a `<block>` over the shadow. Best to check for both. Also, check the type attribute of the `<block>` and `<shadow>` tags to make sure that they are actually numbers; The decompiler didn't used to check the types of the shadow blocks so some very old scripts could have some weird ones.

editor/extension.ts
```diff
@@ -500,8 +500,11 @@ namespace pxt.editor {
 
         // showleds
         const nodes = U.toArray(dom.querySelectorAll("block[type=device_show_leds]"))
+            .concat(U.toArray(dom.querySelectorAll("shadow[type=device_show_leds]")))
```
riknoll: Don't need to test shadows for device_show_leds (its a statement not an expression) but it won't hurt anything either.

Issue #628, opened 2018-02-06
set Brightness issue when using alongside show Icon


microbit-mark:
Micro:bit-support: 3716, 7103

LED brightness remains at a dimmed level despite declaring the level in the program. This occurs when using a combination of the basic.showIcon and plot.LED functions. I have tried a variety of blocks to reproduce and results are the same.

The LEDs seem to be dimmer than 100.
![image](https://user-images.githubusercontent.com/31242877/35872210-2f0dfdcc-0b5e-11e8-8e27-3a6f7f9bd2e5.png)
![image](https://user-images.githubusercontent.com/31242877/35872217-357fa8cc-0b5e-11e8-9aef-1893275082fd.png)
![image](https://user-images.githubusercontent.com/31242877/35872228-3c1eda86-0b5e-11e8-9b9c-1921c46ee2a5.png)
![image](https://user-images.githubusercontent.com/31242877/35872233-3ffd73d8-0b5e-11e8-867b-6c0a9a20ae75.png)
![image](https://user-images.githubusercontent.com/31242877/35872240-43a66828-0b5e-11e8-9bde-7fec77c2faf0.png)
This occurs in the simulator and on the hardware and maybe related to #3196

From @pelikhan: Our image parser needs to emit 0xff values for on pixels instead of 0x01. Otherwise this happens when switching to grey scale mode.

From  @nielsswinkels. I just ran into what I think is the same bug. It behaves correct on the simulator (firefox) but not on the hardware. Very weird bug to run into.

Even setting a pixel to brightness 255 does not provide a workaround this bug:
https://makecode.microbit.org/_F0b1hPKYJPT6

pelikhan:
When you use "plot brightness", you implicitly switch the screen in greyscale mode. In that mode, our show icon block and puts 1s instead of 255s in the pixel intensity.

slylockfox:
I used the following workaround to change the brightness from 1 to 255 in the image:
```javascript
ImageArch = images.createImage(`
    . # . # .
    # . . . #
    # . . . #
    # . . . #
    # . . . #
    `)
for (let i = 0; i <= 4; i++) {
    for (let j = 0; j <= 4; j++) {
        if (ImageArch.pixelBrightness(i, j) > 0) {
            ImageArch.setPixelBrightness(i, j, 255)
        }

    }
}
```

chewi:
Instead of the above workaround, you can just do this:

```js
[...]

PR #1160, opened 2018-08-31
Sim: greyscale display now respects brightness

guillaumejenkins:
https://github.com/Microsoft/pxt/pull/4637 is needed to fix https://github.com/Microsoft/pxt-microbit/issues/628#issuecomment-417247530, but it also introduces a second problem.

Images are now always at brightness 255 in greyscale mode in the simulator. This makes it so the sim respects the max brightness always.

### DIFF
--- a/sim/state/ledmatrix.ts
+++ b/sim/state/ledmatrix.ts
@@ -134,6 +134,7 @@ namespace pxsim.ImageMethods {
         let cb = getResume();
         let first = true;
 
+        clampPixelBrightness(leds);
         board().ledMatrixState.animationQ.enqueue({
             interval,
             frame: () => {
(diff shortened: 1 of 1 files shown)


sim/state/ledmatrix.ts
```diff
@@ -230,6 +233,19 @@ namespace pxsim.ImageMethods {
             whenDone: cb
         })
     }
+
+    function clampPixelBrightness(img: Image) {
+        if (led.displayMode() === DisplayMode.greyscale) {
```
pelikhan: and brightness < 0xff

guillaumejenkins: Why? We want the brightness from `setBrightness()` to be respected even if the image was at 255, no?

guillaumejenkins: Oh, you mean current screen brightness! Will add the check

PR #1165, opened 2018-09-03
Pull common files from pxt-common-packages

mmoskal:
Depends on https://github.com/Microsoft/pxt-common-packages/pull/407 and a bump

This removes file that are shared with pxt-common-packages and adds npm reference for it


pelikhan:
We need to update ``README.md``

Issue #1075, opened 2018-08-12
Pitch and roll values are reversed In real machine.


mbitfun:
**Describe the bug**

In simulator the pitch and roll values are correct.
But, In real machine pitch and roll values are reversed

**To Reproduce**

Please check in real machine.

![microbit-screenshot](https://user-images.githubusercontent.com/33863428/44006308-6aff08ec-9ebd-11e8-8c74-c95618919812.png)

**Desktop (please complete the following information):**
 - OS: Windows10
 - Browser Chrome
 - Version 68

guillaumejenkins:
Good catch, this is a regression as it only happens in /beta

guillaumejenkins:
@jaustin has something changed in the micro:bit framework around rotation? Our `rotation` API hasn't changed at all since v0. The functions are exactly the same:

v0: https://github.com/Microsoft/pxt-microbit/blob/master/libs/core/input.cpp#L310
v1: https://github.com/Microsoft/pxt-microbit/blob/v1/libs/core/input.cpp#L321

```
enum class Rotation {
    //% block=pitch
    Pitch = 0,
    //% block=roll
    Roll = 1,
};

int rotation(Rotation kind) {
    switch (kind) {
        case Rotation::Pitch: return uBit.accelerometer.getPitch();
        case Rotation::Roll: return uBit.accelerometer.getRoll();
    }
    return 0;
}
```

guillaumejenkins:
Maybe @finneyj or @jamesadevine would have some info as well?

jaustin:
Looks strongly like we broke this in the DAL, as the 'dal-integration' instance of v0 gets this wrong too.

@finneyj we should fix our axes!

DAL bug https://github.com/lancaster-university/microbit-dal/issues/371

(PS this reminds me of #799 - which would have helped a lot here!)

finneyj:
Thanks @guillaumejenkins, @jaustin. @mbitfun.

Curious - we definitely checked the axes were consistent after the patch. Will check again...

finneyj:
Just built up two fresh C++ native test cases for the current HEAD revision of microbit-dal master, and the HEAD revision of dal-integration (which should translate to the versions compared above).

The accelerometer data coming out of the device driver via getX(), getY() and getZ() and this is consistent in terms of magnitude and axis alignment, so the issue must be constrained to the pitch and roll computation a little further up the stack... will dig some more.

abchatra:
@finneyj any update here?

[...]

PR #1174, opened 2018-09-05
dal-integration-2

guillaumejenkins:
~~Added `#define MICROBIT_FULL_RANGE_PITCH_CALCULATION 0` in `pxt.h`, not sure if this is the right place~~

~~[View this PR without whitespace](https://github.com/Microsoft/pxt-microbit/pull/1174?w=1)~~

Now uses the new `dal-integration-3` tag, which has the fix and the config flag

### DIFF
--- a/libs/core/pxt.h
+++ b/libs/core/pxt.h
@@ -2,6 +2,7 @@
 #define __PXT_H
 
 //#define DEBUG_MEMLEAKS 1
+#define MICROBIT_FULL_RANGE_PITCH_CALCULATION 0
 
 #pragma GCC diagnostic ignored "-Wunused-parameter"
 
@@ -63,7 +64,7 @@ namespace pxt {
   int templateHash();
   int programHash();
   uint32_t programSize();
-  uint32_t afterProgramPage();     
+  uint32_t afterProgramPage();
   int getNumGlobals();
   RefRecord* mkClassInstance(int vtableOffset);
 
@@ -169,7 +170,7 @@ namespace pxt {
   };
 
   class Segment {
-  private:    
+  private:
       uint32_t* data;
       uint16_t length;
       uint16_t size;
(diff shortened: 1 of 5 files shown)


libs/core/pxt.h
```diff
@@ -2,6 +2,7 @@
 #define __PXT_H
 
 //#define DEBUG_MEMLEAKS 1
+#define MICROBIT_FULL_RANGE_PITCH_CALCULATION 0
```
pelikhan: I am not sure that is how it works. Typically yotta flags are specified in confit.json.

guillaumejenkins: Right, I was going with what Joe said here: https://github.com/Microsoft/pxt-microbit/issues/1075#issuecomment-418883498

pelikhan:
He is missing the glue code
“there should be an option to allow you to flip this in your config.json file, but I'm afraid I forgot to add some glue in dal-integration-2 to let this happen.”

pelikhan:
Try it out but we should an option eventually.

libs/core/pxt.h
pelikhan: Let's drop this. Joe said to let it go and fix it in a later release. It is just a small discreptancy... and honestly it's more consistent if both behave the same.

abchatra:
Changes for v1 as well?

pelikhan:
Note that making v0 and v1 inconsistent is Wrong. We should use the new behavior in both editors and call it a big fix.

libs/core/pxt.json
```diff
@@ -54,7 +54,8 @@
                     "pairing_mode": 1,
                     "whitelist": 1,
                     "security_level": "SECURITY_MODE_ENCRYPTION_NO_MITM"
-                }
+                },
+                "full_range_pitch_calculation": 0
```
pelikhan: This is marked as optional option (that can be override ) - it should be in the mandatory option section.

Issue #1176, opened 2018-09-06
The simulator‘s frame is not clear on high contrast on mode


Helen09:
**Repro steps:**
1. Navigate to https://makecode.microbit.org/beta#editor
2. Click "More..." -> "High Contrast On" to open high contrast mode
3. Click "Music" and drag "Play Tone" and put it under block "forever"
4. Observe the simulator

**Expected result:**
The simulator‘s frame should be clear on high contrast on mode
![issue906com](https://user-images.githubusercontent.com/23466737/45144513-f413b600-b1f0-11e8-8b92-9a862d215e86.PNG)

**Actual result:**
The simulator‘s frame is not clear on high contrast on mode
![issue90601](https://user-images.githubusercontent.com/23466737/45144565-14437500-b1f1-11e8-9d4c-99051c631def.PNG)

PR #1184, opened 2018-09-06
Add white stroke to sim border

guillaumejenkins:
Use a stroke instead of a border, which looks much better and works when other sim parts are present

Fixes https://github.com/Microsoft/pxt-microbit/issues/1176

image

image


pelikhan:
Needs cross-browser test.

Issue #1150, opened 2018-08-30
WebUSB: No warning if a non-MakeCode program is on the device when attempting to flash


pmcalabrese:
**Describe the bug**
Even if the microbit is paired over WebUSB the firmware will be downloaded on my machine instead of be flashed on the device.

**To Reproduce**
Go to https://makecode.microbit.org/beta?webusb=1#editor
pair the microbit by clicking "pair device"

![image](https://user-images.githubusercontent.com/1509241/44854906-1d20cd80-ac6a-11e8-8920-8eb478a84b4e.png)

a prompt will confirm the microbit is paired successfully

![image](https://user-images.githubusercontent.com/1509241/44854981-4c373f00-ac6a-11e8-83b5-29c38e091dae.png)

create a sketch and click Download button

The firmware (.hex) will be downloaded on my machine instead of be flashed over WebUSB

![image](https://user-images.githubusercontent.com/1509241/44855068-83a5eb80-ac6a-11e8-90e2-e79875085124.png)

**Expected behavior**
The firmware (.hex) should be flashed over WebUSB into the microbit

**Screenshots**
Add screenshots to help explain your problem. You can copy paste the screenshot in the github report. The .gif screen recording is very useful as well. 

**Desktop (please complete the following information):**
 - OS: OSX 10.13.6
 - Browser Chrome
 - Version 68.0.3440.106 (Official Build) (64-bit)

**Additional context**
i followed this guide:
https://support.microbit.org/support/solutions/articles/19000084059-beta-testing-web-usb

Microbit DETAILS.TXT
```
# DAPLink Firmware - see https://mbed.com/daplink
Unique ID: 9900000048074e45001810150000005e0000000097969901
HIC ID: 97969901
Auto Reset: 1
Automation allowed: 0
Overflow detection: 0
Daplink Mode: Interface
Interface Version: 0247
Git SHA: 69d1e077da73dbf0fc1db599bf35ce2a7b016aa8
Local Mods: 0
USB Interfaces: MSD, CDC, HID, WebUSB
Interface CRC: 0xc651b9dd
Remount count: 0
URL: https://microbit.org/device/?id=9900&v=0247
```

guillaumejenkins:
What happens if you flash a MakeCode (.hex) program manually via drag & drop, and then try the WebUSB download again?

pmcalabrese:
[...]

PR #1208, opened 2018-09-11
Web usb error handling

guillaumejenkins:
This is the new behavior of WebUSB for microbit with these changes:

- If micropython program, fall back to flashing entire file
  - 60 sec timeout
- If makecode program, perform a quick flash (same as right now)
  - 25 sec timeout
- Both cases have a timeout message asking to reconnect device and manually flash
- Both cases have a catch-all generic error message asking to manually flash

Fixes https://github.com/Microsoft/pxt-microbit/issues/1150


libs/core/_locales/core-jsdoc-strings.json
```diff
@@ -159,6 +159,12 @@
   "String.concat": "Returns a string that contains the concatenation of two or more strings.",
   "String.concat|param|other": "The string to append to the end of the string.",
   "String.fromCharCode": "Make a string from the given ASCII character code.",
+  "String.includes": "Determines whether a string contains the characters of a specified string.",
```
pelikhan: you need to bump pxt and pxt-common-packages for this.

pelikhan:
Read to merge.

PR #1217, opened 2018-09-12
Pair device dialog

guillaumejenkins:
Let's not merge until design is final

Requires https://github.com/Microsoft/pxt/pull/4669

Here's the proposed flow:

## 1. WebUSB unavailable
Normal download happens and normal instructions are shown:
image

### 2.a Device already paired
Flash happens automatically, user sees no dialog

### 2.b Device not paired (or no device connected)
User sees the following dialog:
image

#### 2.b.i User chooses to pair
Pairing prompt is shown, and after the user selects the device, flash happens immediately without further action. From that point on, we flash automatically every time, same as 2.a.

#### 2.b.ii User chooses "Don't enable one-click"
Normal download and instructions happen (same as 1.), and from that point on in that **session**, we never prompt for WebUSB again (user has to refresh the page to enable WebUSB again)


editor/extension.ts
```diff
@@ -432,39 +432,12 @@ namespace pxt.editor {
...
-        let saveHexAsync = () => {
-            return pxt.commands.saveOnlyAsync(resp)
-        }
-
+    function flashAsync(resp: pxtc.CompileResult, d: pxt.commands.DeployOptions = {}): Promise<void> {
         startTime = 0
         let wrap: DAPWrapper
         log("init")
 
+        d.showNotification(U.lf("Flashing device..."));
```
riknoll: We don't really use the term "flashing" anywhere else

editor/extension.ts
```diff
@@ -832,29 +883,94 @@ namespace pxt.editor {
         valueNode.appendChild(s);
     }
 
+    function webUsbDialogAsync(confirmAsync: (options: any) => Promise<number>): Promise<number> {
+        const boardName = pxt.appTarget.appTheme.boardName || "???";
+        const htmlBody = `
+        <div class="ui grid stackable">
```
riknoll: confirmAsync has a `jsx` option in addition to the `htmlBody` one, can we use that? I know this code came from another target but it would be nice to stop inflating HTML strings wherever possible.

pelikhan:
We should always display the "new to one-click download" option if available.

abchatra:
2.b buttons should be same as 1. Otherwise there will be cognitive overload.

2.b first time here can be changed to learn more which takes you to second window which has these options.

editor/extension.ts
```diff
@@ -508,29 +481,83 @@ namespace pxt.editor {
...
+            return pxt.commands.saveOnlyAsync(resp);
+        };
+        return Promise.resolve()
+            .then(() => {
+                const isUwp = !!(window as any).Windows;
+                if (isUwp) {
+                    // Go straight to flashing
+                    return flashAsync(resp, d);
+                }
+                if (!pxt.usb.isAvailable) {
```
pelikhan: i think you need isEnabled here. isAvailable() is a function.

PR #1220, opened 2018-09-13
Edit 'tools' projects

ganicke:
A tiny edit to the _Analog Pin Calibrator_ page.

The _Analog Pin Calibrator_ and _Light Level Calibrator_ don't really calibrate anything. They just show the values read. How about calling them _Analog Pin Tester_ and _Light Level Meter_?

RE: #1218


pelikhan:
Sure, go ahead with the renames

Issue #1229, opened 2018-09-13
Block snap sound different in v1


pelikhan:
Try snapping blocks in v0 vs v1. Somehow they are different now.

pelikhan:
Sounds have been restored to v0 sound style.

PR #1231, opened 2018-09-13
enfore v0 sounds

pelikhan:
Copying all sounds from pxt#v0 over to microbit.

Fix for https://github.com/Microsoft/pxt-microbit/issues/1229


riknoll:
Do we need to include a license or something? I

pelikhan:
Added blockly in thirdpartynotices.

PR #1264, opened 2018-09-18
Update plant watering.md

indigo51:
the connection to servo are male not female. Correction to parts listing


pelikhan:
Oops - thanks!

Issue #1263, opened 2018-09-17
game of life opens in blocks


pelikhan:
Open game of life example, you get grey blocks.

Markdown should specify that this is a typescript program.

ganicke:
Sure, I'll take a look at it.

abchatra:
Fixed

PR #1270, opened 2018-09-19
Switch 'game of life' from block to TS

ganicke:
Change from blocks to TS. Functions with parameters don't decompile to blocks.

RE: #1263


abchatra:
Looks good

PR #1289, opened 2018-09-25
Experiments in v1

pelikhan:
* updated pxt references
* added "experiments"


pelikhan:
Requires Experiments branch

PR #1292, opened 2018-09-25
Updating the link to firmware at the right location

abchatra:
Updating the docs for firmware update to the right location


editor/extension.ts
```diff
@@ -870,7 +871,7 @@ namespace pxt.editor {
                 <div style="justify-content: center;display: flex;padding: 1rem;">
                     <img class="ui image" src="./static/download/firmware.png" style="height:100px;" />
                 </div>
-                <a href="/webusbfirmware" target="_blank">${lf("Check your firmware version here and update if needed")}</a>
+                <a href="${docUrl}/webusbfirmware" target="_blank">${lf("Check your firmware version here and update if needed")}</a>
```
pelikhan: the other use of docUrl is guarded (not that it matters here)

PR #1297, opened 2018-09-25
Shim for serial.readBuffer

### LINKED ISSUE #1296: Shim for serial.readBuffer is missing
**Describe the bug**
When using the serial.readBuffer block causes the following simulator error:
Program Error: pxsim.serial.readBuffer is not a function

**To Reproduce**
Add a serial.readBuffer block (see screenshot)

**Expected behavior**
No error. readBuffer an arbitrary filled buffer with correct length.

**Screenshots**
bildschirmfoto 2018-09-25 um 23 08 22

 - OS: Mac
 - Browser chrome
 - Version 69.0.3497.100

philipphenkel:
Fix #1296


sim/state/serial.ts
```diff
@@ -37,6 +37,11 @@ namespace pxsim.serial {
         // TODO
     }
 
+    export function readBuffer(length: number): RefBuffer {
+        // TODO
```
pelikhan: Add this to match C++
```
      if (length <= 0)
        length = DAL.MICROBIT_SERIAL_READ_BUFFER_LENGTH;
```

philipphenkel: This constant seems not to be available. Do you mean DAL.MICROBIT_SERIAL_DEFAULT_BUFFER_SIZE?

pelikhan: Actually just use 64.

philipphenkel: done

PR #1312, opened 2018-09-28
The link was broken

baradaran:
The new naming convention was not linked to this wiki page. Maybe the whole package section should be renamed, but this is a quick fix at the moment.


pelikhan:
Thank you! I've added redirects so that other broken link route correctly.

PR #1320, opened 2018-10-01
render project categories instead of listing

pelikhan:
In v1, projects are split in subpages so the project page only points to this sub-pages to avoid duplicating cards.


ganicke:
I like. I'll need to do some sub-headings in the individual card pages. We have "cool projects" repeated quite a bit.

PR #1324, opened 2018-10-01
Update font size of blocks

samelhusseini:
Make the text more readable on all browsers. 

Also: set Flyout button color to the secondary color. 

Chrome: 
screen shot 2018-10-01 at 2 34 30 pm

Safari: 
screen shot 2018-10-01 at 2 34 37 pm

### DIFF
--- a/theme/blockly.less
+++ b/theme/blockly.less
@@ -11,6 +11,14 @@
 /* Reference import */
 @import (reference) "semantic.less";
 
+
+text.blocklyText {
+    font-weight: 600;
+    text-rendering: optimizeLegibility;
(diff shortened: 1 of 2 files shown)


theme/blockly.less
```diff
@@ -11,6 +11,14 @@
 /* Reference import */
 @import (reference) "semantic.less";
 
+
+text.blocklyText {
+    font-weight: 600;
```
pelikhan: is this a change that should go in all editors?

samelhusseini: Possibly, let's try it here first.

PR #1327, opened 2018-10-02
Bluetooth project settings to disable services, enable memory logging (experimental)

pelikhan:
New flags to selectively disable default BLE services
image


jaustin:
Does the debug output to serial also turn on memory debugging? Are we sure we really want to let people disable these services yet? Happy to see a PR but not sure we've got far enough in investigation to commit to them being there long term?

pelikhan:
I've tagged this as "do not merge"

pelikhan:
Waiting for DAL update that allows less BLE services.

PR #1331, opened 2018-10-02
support for MIDI simulator in v0

pelikhan:
Adds "control.__midiSend" function (implemented in v1) to allow package v0/v1 to work. v1 PR is https://github.com/Microsoft/pxt-microbit/pull/1332


pelikhan:
@abchatra this small fix will allow to get rich MIDI simulator in v1

PR #1333, opened 2018-10-02
Fix notification of specific event values and MICROBIT_EVT_ANY for same event source

philipphenkel:
The current implementation triggers registered handlers twice per event when listening for both ANY and a specific value.


pelikhan:
@philipphenkel could you add a test case as a comment here?

pelikhan:
```
control.onEvent(DAL.MICROBIT_ID_BUTTON_A, DAL.MICROBIT_BUTTON_EVT_CLICK, function () {
    console.log('A click')
    basic.showString("C")
})
control.onEvent(DAL.MICROBIT_ID_BUTTON_A, DAL.MICROBIT_EVT_ANY, function () {
    console.log('A any')
    basic.showString("A")
}) 
```

pelikhan:
@tballmsft can you take a peek? Our simulator eventbus does not handle the ANY flag.

philipphenkel:
@pelikhan The test case is described in issue #1334 and I'm wondering why didn't show up here (I did not notice that - sorry!)

philipphenkel:
The problem still persits. Please have a look at my simplified example:

bildschirmfoto 2018-11-26 um 21 08 20

https://makecode.microbit.org/_CLjbK0asz6zx

When an event is triggered I expect two notifications. One notification for ANY and one for the specific event.

pelikhan:
We need to integrate tom’s multi-listener fix I think.

philipphenkel:
@pelikhan I could have a look at the fix. Is there a PR or branch?

pelikhan:
See https://github.com/Microsoft/pxt/pull/4844 . FYI @tballmsft

philipphenkel:
@pelikhan It looks like https://github.com/microsoft/pxt/pull/4844 did not fix this issue (I assume the Codal changes are included since Feb 22).
As of today, the problem is still there and it is easily producible with my example of 26 Nov 2018 (see above).

philipphenkel:
@pelikhan I retested and can confirm that the problem is fixed with the latest MakeCode version.

I tested the following config:
makecode.microbit.org Version: 2.0.6
Microsoft MakeCode Version: 5.15.12

PR #1335, opened 2018-10-02
updated to v0.14.54

pelikhan:
Moving index-ref to v0.14.54


abchatra:
Can we release this on Friday? We might have other bugs to fix this week on v0 as finish v1.

pelikhan:
Sure

PR #1336, opened 2018-10-02
Update Docs Reference grammar

billbogaiv:
Standardizes title and parameter grammar (capitalization and periods mostly).


billbogaiv:
There's a lot more (at least with the reference pages) that don't follow the camel-case title convention than do. So, the expected convention for titles is camel-case with spaces?

ganicke:
Yes, that's correct (camel-case with spaces). Thanks for scanning through them and helping with a cleanup!

billbogaiv:
@ganicke: I went through all the ref. pages and made the titles consistent.

A few pages were left alone since they were regular docs. pages and not for specific methods.

ganicke:
@pelikhan, OK with the risk level on this merge? It's a lot of files but a lot good minor cleanups.

pelikhan:
I don't think it is worth it. We need to assess how many translations we are destroying.

ganicke:
OK. In the minimum it would at least break all those page titles.

pelikhan:
We can enfore capitalization via CSS.

PR #1337, opened 2018-10-03
Doc improvements

leesei:


leesei:
This command for checking for wrong challenge number maybe useful when porting to v1
`grep -r "Challenge 2" -L | grep challenge | xargs -I@ code @`

PR #1339, opened 2018-10-03
Use SYNC_SPIN_WAIT for serial write and explicitely set SYNC_SLEEP for read

### LINKED ISSUE #1157: Serial not concurrency safe?
**Describe the bug**

Mulitiple forever loops writing to serial produces erroneous output.

**To Reproduce**
Run the simple program at https://makecode.microbit.org/_WEga5JYiX1x7

Three forever loops run in parallell. The body of each loop repeats three times the sequence of pausing one second and writing to serial. Each time a different string is written - loop 1 writes SL1a, SL1b, SL1c, loop 2 writes SL2a, SL2b... and so on. Output on Teraterm is incorrectly SL1a, SL2b SL3c, SL1a, SL2b, SL3c, SL1a...

Picture of block program:
serial output - wrong execution

Producing this incorrect output:
bild

**Expected behavior**
For each loop, all the text should be output, and in the correct order (though it could be interleaved with the output from the other loops). E.g. SL1a, SL2a, SL3a, SL1b, SL2b, SL3b.... and so on.

When a "gate" function is added to make sure that serial is entered only by one forever loop at a time, correct output is produced, e.g. this program: https://makecode.microbit.org/_hDo7Ea59THbY
Writing to the LED matrix instead of to serial also works corretly (without any "gate")

Picture of program with "gate":
thisworks

Producing this correct output:
bild

Could this be an error in restoring an execution pointer in a context switch? Or an overwrite problem in some serial buffer?
The problems occurred in a rather complex program that I was debugging. The more debugging output I put into the program, the more confused I got and it took a lot of time.  Seems rather unfortunate that the tool used for debugging introduces bugs itself.
After suspecting that the strange behavious actually wasn't my fault, I was eventually able to reproduce the problem in the simple programs above.

--Tomas

**Desktop :**
 - OS: Windows 7
 - Browser Firefox
 - Version 61.0.2 (64-bitars)

philipphenkel:
Fix #1157


pelikhan:
Pending @finneyj comment https://github.com/Microsoft/pxt-microbit/issues/1157#issuecomment-426666968

philipphenkel:
@abchatra I just noticed that https://github.com/microsoft/pxt-microbit/issues/1157 was closed with a comment that's DAL issue. This is not entirely true as we can just set SYNC_SLEEP for send. My PR addresses #1157 on the MakeCode side by the setting the SYNC_SLEEP mode. 

What is the plan going forward?

My initial PR for reference:
https://github.com/philipphenkel/pxt-microbit/commit/b72ad9a860825939fad75ae3764f24a9d68fdae8#diff-bf9dc4e8ed88be69ae8bd7dc5e347102

PR #1348, opened 2018-10-04
fixed links in SUMMARY.md

vinayak42:


pelikhan:
You need to run our server locally to render those markdown files. The link do not render properly in GitHub.

PR #1353, opened 2018-10-05
BLE UART Buffer methods

pelikhan:
read/write buffers through BLE UART


libs/bluetooth/bluetooth.cpp
```diff
@@ -98,6 +98,29 @@ namespace bluetooth {
...
+    /**
+    * Reads buffered UART data into a buffer
+    */
+    //%
+    Buffer uartReadBuffer() {
+        startUartService();
+        int bytes = uart->rxBufferedSize();
+        auto buffer = mkBuffer(NULL, bytes);
+        uart->read(buffer->data, buffer->length);
+        // todo: handler read return value
```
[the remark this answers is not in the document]
pelikhan: Updated, do I need to use macros for the reference counting?

libs/bluetooth/bluetooth.cpp
```diff
@@ -98,6 +98,30 @@ namespace bluetooth {
...
+    /**
+    * Reads buffered UART data into a buffer
+    */
+    //%
+    Buffer uartReadBuffer() {
+        startUartService();
+        int bytes = uart->rxBufferedSize();
+        auto buffer = mkBuffer(NULL, bytes);
+        int read = uart->read(buffer->data, buffer->length);
+        if (read != buffer->length)
```
mmoskal: yeah, it should be: `tmp = mkBuffer(buffer->data, read); decrRC(buffer); buffer = tmp;`

`decrRC()` is essentially equivalent to `free()` here (since there is only one reference)

mmoskal: also I guess when `read<0` you could return `null` or empty buffer

libs/bluetooth/bluetooth.cpp
```diff
@@ -98,6 +98,33 @@ namespace bluetooth {
...
+    /**
+    * Reads buffered UART data into a buffer
+    */
+    //%
+    Buffer uartReadBuffer() {
+        startUartService();
+        int bytes = uart->rxBufferedSize();
+        auto buffer = mkBuffer(NULL, bytes);
+        int read = uart->read(buffer->data, buffer->length);
+        if (read != buffer->length) {
```
mmoskal: so `read` is never negative?

libs/bluetooth/bluetooth.cpp
```diff
@@ -117,7 +117,7 @@ namespace bluetooth {
         int bytes = uart->rxBufferedSize();
         auto buffer = mkBuffer(NULL, bytes);
         int read = uart->read(buffer->data, buffer->length);
-        if (read != buffer->length) {
+        if (read >= 0 && read < buffer->length) {
```
mmoskal: Sorry for being picky, but how about `if (read < 0) { decrRC(buffer); return mkBuffer(NULL, 0); }`

pelikhan: picky is good

PR #1379, opened 2018-10-11
Tilt support in breadboard + display of accelerometer values

pelikhan:
Simulator improvements:
- [ ] tilting with breadboard
accsim
- [ ] display accelerometer values
accsimbreadboard

### DIFF
--- a/sim/visuals/boardview.ts
+++ b/sim/visuals/boardview.ts
@@ -3,7 +3,6 @@ namespace pxsim.visuals {
         return new visuals.MicrobitBoardSvg({
             runtime: runtime,
             theme: visuals.randomTheme(opts.highContrast),
-            disableTilt: false,
             wireframe: opts.wireframe
         });
     }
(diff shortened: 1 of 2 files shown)


samelhusseini:
Would you get all these verbose values when a simple 'on shake' block appears. If so, I'm not sure these values are relevant for most kids that are wanting to use the accelerometer for shaking. 

Maybe we can think of a way to display these values in a non-intrusive way.

pelikhan:
Only showing values that are being read now.

pelikhan:
@jaustin FYI

pelikhan:
The visual changes are positioned in the simulator SVG so form factor does not affect this. For the tilt, we use to do the computation but we did not apply it to the correct DOM element.

pelikhan:
So not risky. Very useful to see accelerometer values live.

abchatra:
Looks good.

Issue #1375, opened 2018-10-10
"Take me to the old editor" link may be too specific


jaustin:
When opening a program that contains an extension that doesn't upgrade well, we get the following dialogue

![image](https://user-images.githubusercontent.com/147113/46741599-4b4bf100-cc9d-11e8-94ac-a2d57dcfb0c7.png)

The 'go to the old editor' button takes me to
https://makecode.microbit.org/v0.14.54#editor

This exposes direct version selection in a way that could leave users saving URLs to older versions

is it better to send it to just /v0 ?

pelikhan:
Good one we should use the major version only.

abchatra:
We are going to fix this to redirect to v0.

abchatra:
Fixed. Will be available in /beta soon.

PR #1384, opened 2018-10-11
Redirect logic to v0

abchatra:
Redirect to latest v0 if the upgrade fails. We don't want to redirect to internal builds.


abchatra:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1375

PR #1417, opened 2018-10-15
Home courses page

pelikhan:
image

Listing courses from the home page


ganicke:
How about something else for _discover-javascript-blocks.png_:

image

pelikhan:
Sure

PR #1431, opened 2018-10-15
Add more desc to on-received-buffer

ganicke:
RE: #1430 

Add some extra description to _on_received_buffer_.

Technically, what does "set, synchronized, before the radio..." mean? Is there a period of time (500ms, etc.) before the group numbers are known between stations?


pelikhan:
Synchronized means they must have the same group.

PR #1447, opened 2018-10-17
A few touches to 'js writing code'

ganicke:
I'll add another gif for SHIFT-ALT-F later. My new version ScreenToGif isn't getting dpi correct for some reason.


ganicke:
Got it...
dummy

PR #1521, opened 2018-10-25
Adding Beacon Hill course card

pelikhan:
image


ganicke:
Ah, ok, education term for activity inclusion.

Issue #1463, opened 2018-10-19
An error shows in example "Karel the LED" tutorial


AimeeCao18:
**Repro steps:**
1. Navigate to https://makecode.microbit.org/projects/karel
2. Scroll down to "Coding Karel the LED" step 
3. Copy the code to the "JavaScript" editor page

**Expected result:**
Error should not show when copying the example code to "JavaScript" editor

**Actual result:**
An error shows in example "Karel the LED" tutorial
![1](https://user-images.githubusercontent.com/44188776/47201520-bb98f780-d3ac-11e8-80d6-082157c029f7.png)

abchatra:
Nice find. Thanks @AimeeCao18

abchatra:
Although I am not able to repro. Can you try again?

AimeeCao18:
I verified this issue again and it still can repro. Below is the published project
https://makecode.microbit.org/_3puJVFePu4yJ

pelikhan:
TypeScript 2.0 is probably more strict that 1.0 (or the other way around.

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

abchatra:
@riknoll can you take a peek?

pelikhan:
I suspect the old version is still cached. I did rewrite that sample.

ganicke:
Moving ``const board = new Board();`` (in the old karel) below the class seems tricksy enough to satisfy both versions in case of cache confusion.

pelikhan:
Make sure to pull the source from https://makecode.microbit.org/v1/projects/karel

PR #1523, opened 2018-10-25
Update 'karel' to use typescript source from v1

ganicke:
Upgrade the typescript source with fixes from the `v1` version of the project.

RE: #1463


pelikhan:
Oopos, this should have gone in v0. @ganicke master is frozen now.

ganicke:
OK, oops then. See #1529

PR #1524, opened 2018-10-25
Updating hero banner

samelhusseini:
hero


Jaqster:
Can you move the blocks to be more centered (move to the right a bit)?  Also, did you see what the other colored background looked like?

samelhusseini:
- [ ] Center the blocks
- [ ] Connect the red pin to the micro:bit

samelhusseini:
How's the latest version @Jaqster?

Jaqster:
Did you move the blocks over to the right?  I can't tell...  Maybe if you could leave the servo motor on the left, and move the micro:bit guy with the blocks so it's in the center, that would help spread the elements out a bit more.

samelhusseini:
Okay how about this?

hero1

Jaqster:
Perfect!  Ship it!

samelhusseini:
Awesome!

PR #970, opened 2018-07-26
added DF robot Oblog

microbit-mark:
On request from DFRobot adding this package and creating iOT category in packages

PR #1527, opened 2018-10-25
Adding missed dfrobot

abchatra:
Please double check everything: https://github.com/Microsoft/pxt-microbit/pull/970

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -136,6 +136,15 @@ Check out [microbit.org](http://microbit.org/resellers/) for more information on
 }]
 ```
 
+## Internet of Things
+```codecard
+[{
+  "name": "DF Robot Obloq",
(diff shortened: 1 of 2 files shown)


pelikhan:
Roll it back.
- [ ] no icon on package
- [ ] no release

abchatra:
Oh was it intentional? Why is it in v0 ?

pelikhan:
I expected Mark to check these things. We need to file bugs against that repo to get them to produce a release and get an icon.

abchatra:
I am not reverting this now. Here is the scenario: On Friday 26th user can use dfrobot package but Monday due to update he can't

Lets keep it sync and if we want to roll back we need to do for both v0 & v1.

pelikhan:
If it was already in v0, it's legacy now.

PR #1530, opened 2018-10-26
Add 'Writing Functions' lesson to Blocks2JS

ganicke:
- [x] Add new 'Writing Functions' lesson
- [x] Fix SUMMARY for Blocks2JS links
- [x] Add a card for 'Writing Functions' in card page

I reused the card image from 'Writing JavaScript'. I could try to make a new one in the same style 😐


pelikhan:
Could we have an intro section that shows how it goes in blocks? (it is a blocks -> javascript course after all :))

ganicke:
Sure, I'll add a section starting out the page by showing the simpler parmless, returnless function in blocks. Lesson will then transition to the more complex forms that are written in JS.

PR #1540, opened 2018-10-26
radio double support

pelikhan:
Support for TNumber on the C++ layer. This protocol should be migrated in TS.

### DIFF
--- a/docs/reference/radio/packet.md
+++ b/docs/reference/radio/packet.md
@@ -10,6 +10,45 @@ A packet that was received by the radio.
 * `serial` - The serial number of the @boardname@ that sent this packet or `0` if the @boardname@ did not include its serial number.
 * `signal` - How strong the radio signal is. The exact range of values varies, but it goes approximately from `-128` dB (weak) to `-42` dB (strong).
 
+## Packet layout
+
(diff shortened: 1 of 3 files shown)


abchatra:
This should go to next release, not today. Too late.

docs/reference/radio/packet.md
```diff
@@ -10,6 +10,45 @@ A packet that was received by the radio.
...
 * `signal` - How strong the radio signal is. The exact range of values varies, but it goes approximately from `-128` dB (weak) to `-42` dB (strong).
 
+## Packet layout
+
+This section documents the data layout of the packet if you need to interpret the data outside of MakeCode.
+
+    Packet byte layout
+    | 0              | 1 ... 4       | 5 ... 8           | 9 ... 28
+    ----------------------------------------------------------------
+    | packet type    | system time   | serial number     | payload
```
mmoskal: Always sending time and serial seems somewhat wasteful for strings and buffers.

mmoskal:
I'm also wondering, why don't we do all this stuff in TS? We could just have a function to send and receive a raw buffer and do the encoding in TS.

pelikhan:
Tested on hardware. This code should definitely go to TS.

Issue #1164, opened 2018-09-03
RPS teams doc


microbit-mark:
micro:bit-support: 6884

**Describe the bug**
https://github.com/Microsoft/pxt-microbit/blob/master/docs/projects/rps-teams.md

The line "the team with the most players wins the game" is confusing. Does it mean the team that has the greatest number of rocks, paper or scissors wins, or the highest tally of either rocks, paper or scissors? Perhaps it just needs clarifying a bit further.

ganicke:
The project provides description for just one team. Maybe say that: "players with the largest roster are the winners"?

PR #1576, opened 2018-11-02
Rps teams update

ganicke:
- [x] Change events to function parms.
- [x] Set range bound for random.
- [x] Clarify what a "team" is.
- [x] A few other edits.

RE: #1164 

A game duration is a one shake or does the game continue for many shakes?


pelikhan:
Game never really stops, it constantly updates as players shake.

Issue #1314, opened 2018-09-28
Improve documentation for I2C blocks


ppelleti:
**Is your feature request related to a problem? Please describe.**
This came up when someone asked a question on the micro:bit Slack.  He was trying to use the I2C read and write blocks to read and write an I2C EEPROM.  In helping him, I was a bit annoyed that the documentation on the read and write blocks was minimal.  In particular, I had to go to the [Runtime documentation](https://lancaster-university.github.io/microbit-docs/ubit/i2c/) to fully understand what the "repeated" parameter meant.

**Describe the solution you'd like**
1. It would be helpful to have a little more text to explain to the uninitiated that these are I2C bus reads and writes, and not memory reads and writes.
2. Document the "repeated" parameter.
3. Explain the difference between Int8LE and Int8BE.  (How does a single byte have an endianness?)

**Describe alternatives you've considered**
An alternative to documenting the "repeated" parameter would be to rename it "send stop bit" (or perhaps "end of transaction") and invert the sense of it.  I think that would be clearer to those familiar with I2C.  (And I'm not sure the I2C blocks will be all that useful in any case to those unfamiliar with I2C.)

If there's no difference between Int8LE and Int8BE, then just replace them with a single "Int8" choice.

**Additional context**
None.

ppelleti:
Just a followup: I noticed that the repeated parameter *is* documented [here](https://makecode.microbit.org/reference/pins/i2c-read-number), so I just didn't RTFM enough.  The "documentation" I'm referring to in the report above is the mouse-over documentation for the block.  (i. e. the doc comment)

Also, I found the documentation for [NumberFormat](https://makecode.microbit.org/types/buffer/number-format), but that still didn't answer my question of why single-byte numbers are available in both big endian and little endian.

abchatra:
@ppelleti if you want to help feel free to give a pull request with updated documentation. There is an edit button at the bottom of these pages where you can find the source of file in github.

PR #1589, opened 2018-11-05
I2C ref doc 'repeated' description

ganicke:
Try to clarify what the **repeated** parameter does, re: #1314.

Question: When **repeated** is `true`, will this NOT send the stop condition during the operation of the individual API call (when numbers or buffers are formed from multi-byte reads/writes but send the stop on the last byte read/write)? Or, will it NOT send the stop condition per API call at all (stop condition is not used at any time during byte transfers in the call)?

The use of **repeated** is not well described in any of the codal/mbed stuff:

https://github.com/lancaster-university/microbit-dal/blob/master/source/drivers/MicroBitI2C.cpp
https://lancaster-university.github.io/microbit-docs/ubit/i2c/
https://os.mbed.com/docs/v5.10/mbed-os-api-doxy/classmbed_1_1_i2_c.html


ganicke:
Also, does `UInt8BE` an `Int8BE` really mean anything? Such a thing might indicate an MSB and LSB reversal of bit order in the byte, but really?

pelikhan:
Big Endian or Little Endian

mmoskal:
The BE and  LE don't mean anything in one byte case - they are there for consistency, though I guess that can be argued. The UInt and Int are differnt.

pelikhan:
They are staying for legacy.

pelikhan:
@ganicke see answer above.

ganicke:
Thanks all for the info. It looks like the actual answer is down in here:

https://github.com/ARMmbed/mbed-os/blob/master/targets/TARGET_NORDIC/TARGET_MCU_NRF51822/i2c_api.c

where the stop is written once after all the byte writes/reads.

ganicke:
RTM

PR #1619, opened 2018-11-09
Removing v0 settings and adding electron version

abchatra:
2  changes
- Remove the v0 specific settings as they are defined in v0 targetConfig
- Add electron update settings.

Waiting for the build before I merge this.


abchatra:
@pelikhan removing v0 config from here are ok right?

pelikhan:
Yes it is served per branch.

PR #1633, opened 2018-11-13
stop melody block

pelikhan:
Requested block to stop melodies.


pelikhan:
@samelhusseini ready to roll

PR #1641, opened 2018-11-14
Update targetconfig.json

microbit-mark:
add extensions from @shaoziyang who is going to update readmes to include typescript rather than code images. 

Going to process extensions.md seperately as we are reviewing our list


pelikhan:
Mark can you resolve the conflicts?

PR #1652, opened 2018-11-19
Spellcheck docs

samelhusseini:
Ran a spellchecker on our docs.


pelikhan:
Any way to automate this?

samelhusseini:
Yeah looking into that. Problem is there are a bunch of valid words that we use like mbed. so I'm curating that list first.




samelhusseini:
Yeah looking into a good way to configure the things to skip. Otherwise it will get repetitive if you constantly have to skip over the same words.

PR #1694, opened 2018-12-03
Add game state boolean ref docs

ganicke:
- [x] Add doc pages for **isRunning**, **isPaused**, **isGameOver**
- [x] Link up pages in _game_ card page
- [x] Fix api help paths and descriptions
- [x] Remove the `?` in the block text. Do we use this for conditions normally???
- [x] Add ``!_paused`` as a condition for  **isRunning**.

Would ``isRunning()`` not mean:

```typescript
export function isRunning(): boolean {
        return !_isGameOver && !_paused;
}
```

Block look:

image

RE #1677

PR #1695, opened 2018-12-03
Remove punctuation for isTouching, isTouchingEdge

xuefengedu:
Remove punctuation for isTouching, isTouchingEdge
To be consistent with other condition blocks, as we discussed in #1694 , thanks.


pelikhan:
Thanks for following up!

PR #1706, opened 2018-12-06
Update targetconfig.json

microbit-mark:
Remove PiSupply org and add extensions.

As part of reviewing extension approvals we will be moving away from approved organisations. The extensiosn are already listed under extensions.md

### DIFF
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -5,7 +5,7 @@
             "microbit-foundation",
             "KitronikLtd",
             "tinkertanker",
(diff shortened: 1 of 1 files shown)


targetconfig.json
```diff
@@ -6,7 +6,6 @@
             "KitronikLtd",
             "tinkertanker",
             "Seeed-Studio",
```
pelikhan: extra comma

microbit-mark: ack. done

PR #1714, opened 2018-12-07
Releasing  v1.2.12 to live (just the dal changes)

abchatra:
Releasing dal changes for microbit.


pelikhan:
Looks good. Good thing we create tons of releases :)

PR #1736, opened 2018-12-12
Releasing 1.2.13 which contains DAL changes

abchatra:
1.2.12 didn't contain dal changes even though in the history it looks. Likely gittag is from a wrong branch,


abchatra:
Nope we figured out that issue is only for the script which are generated from /beta site in Aug

Issue #1769, opened 2018-12-29
Extensions do not have right license


abchatra:
Following approved extensions don't have right license:

No license:
Seeed-Studio/pxt-grove null
Microsoft/pxt-max6675
mbitfun/pxt-katakana null
chevyng/pxt-ucl-junkrobot null
LaboratoryForPlayfulComputation/pxt-BlockyTalkyBLE null
pizayanz/pxt-linebeacon null
jdarling/pxt-pca9685 null
sparkfun/pxt-weather-bit null
k8robotics/pxt-k8 null
sparkfun/pxt-moto-bit null
sparkfun/pxt-gamer-bit null

Microsoft/pxt-sonar { key: 'other',
  name: 'Other',
  spdx_id: 'NOASSERTION',
  url: null,
  node_id: 'MDc6TGljZW5zZTA=' }
srs/pxt-bitbot { key: 'apache-2.0',
  name: 'Apache License 2.0',
  spdx_id: 'Apache-2.0',
  url: 'https://api.github.com/licenses/apache-2.0',
  node_id: 'MDc6TGljZW5zZTI=' }
 null
PiSupply/pxt-iot-lora-node { key: 'gpl-3.0',
  name: 'GNU General Public License v3.0',
  spdx_id: 'GPL-3.0',
  url: 'https://api.github.com/licenses/gpl-3.0',
  node_id: 'MDc6TGljZW5zZTk=' }
Microsoft/pxt-hacking-stem { key: 'other',
  name: 'Other',
  spdx_id: 'NOASSERTION',
  url: null,
  node_id: 'MDc6TGljZW5zZTA=' }
PaulDFoster/pxt-microbit-GY521 null
Annikken/pxt-Andee null

abchatra:
```javascript
let fs = require("fs")
let rp = require("request-promise-native");
let s = fs.readFileSync("targetConfig.json")
let config = JSON.parse(s);
let repos = config.packages.approvedRepos;
let orgs = config.packages.approvedOrgs;

async function reqAsync(query) {
 var options = {
    uri:  query,
    headers: {
        'User-Agent': 'Request-Promise'
    },
    qs: {
        access_token: '' // -> uri + '?access_token=xxxxx%20xxxxx'
    },
    resolveWithFullResponse: true,
    json: true 
    }

    resp = await rp(options)
        .then(function (response) {
            return {json: response.body, headers: response.headers};
        })
        .catch(function (err) {
            console.log(err)
        });

[...]

PR #1777, opened 2019-01-03
Removing orgrepo and adding few default extensions

abchatra:
Two changes:
1. This change removes blanket approval for organizations. This is to make sure each repo has the right license to serve. See #1769 
2. Add few default extensions, top 10 extensions by usage, in last 90 days which have MIT license and an image are shown. 
image


abchatra:
@Jaqster @jaustin

PR #1784, opened 2019-01-10
ad max-88

microbit-mark:
add max-88 @alankrantas


abchatra:
License, image and tag looks good.

PR #1801, opened 2019-01-22
add 51bit SFC

microbit-mark:
add 51bit SFC/NES controller extension @xuefengedu


xuefengedu:
@microbit-mark  thanks for great help!

PR #1819, opened 2019-02-06
auto-generated projects summary

pelikhan:
Generated from the build.


ganicke:
Just the SUMMARY is generated or the card page too? Cards won't have descriptions if that's OK? Well, it's been that way the whole time I guess.

pelikhan:
It's auto-generated. I guess there is no description on those pages or it would have picked it up.

PR #1851, opened 2019-02-22
V1 handlers

tballmsft:


pxtarget.json
```diff
@@ -221,7 +221,7 @@
         "yottaTarget": "bbc-microbit-classic-gcc",
         "yottaCorePackage": "microbit",
         "githubCorePackage": "lancaster-university/microbit",
-        "gittag": "pxtgc-0",
+        "gittag": "pxtgcevents",
```
abchatra: Why are we changing the gittag?

tballmsft: because Joe did some work on the DAL to support multiple handlers; michal created a branch with that work and the gc work.

abchatra: This needs to be in master otherwise next release will override these changes

tballmsft: Understood. Michal created the pxtgcevents branch of microbit-dal in order to combine the DAL changes. Eventually, everything will go to master of microbit-dal, as with pxt-microbit.

libs/bluetooth/pxt.json
```diff
@@ -18,6 +18,7 @@
     "yotta": {
         "config": {
             "microbit-dal": {
+                "message_bus_concurrency_mode": 1,
```
mmoskal: This needs to go in `core/pxt.json`, otherwise it's only enabled in bluetooth apps

libs/core/pxt.h
```diff
@@ -37,6 +37,8 @@ typedef RefMImage *Image;
 
 extern MicroBit uBit;
 extern MicroBitEvent lastEvent;
+extern void setBackgroundHandlerFlag();
+extern void unregisterFromDal(void *);
```
mmoskal: `extern` not needed (nit-picking ;)

pxtarget.json
mmoskal: I think it's fine to have it in the PR, so the builds go through, and we need to change it before merging

libs/bluetooth/pxt.json
tballmsft: ooh... thanks for that.

mmoskal:
Should we revive this? #2688 adds support in microbit-dal

pelikhan:
Yes we should.

mmoskal:
@tballmsft if you want to test if everything still works as expected, build is here https://makecode.microbit.org/app/3b00a48703e7b5fe39002526c8475c597197a735-2734415fee

libs/core/control.cpp
```diff
@@ -348,6 +348,21 @@ namespace control {
...
+    //%
+    //% help=control/next-handler-to-background
+    void nextHandlerToBackground() {
+        setBackgroundHandlerFlag();
+    }
+
+    /** Remove a handler from the background for all events */
+    //%
+    //% help=control/remove-from-background
+    void removeFromBackground(Action handler) {
```
pelikhan: need to check NULL probbably

libs/core/pxt.json
```diff
@@ -56,6 +56,7 @@
     "yotta": {
         "optionalConfig": {
             "microbit-dal": {
+                "message_bus_concurrency_mode": 1,
```
pelikhan: optionalConfig means someone can turn off this mode. do we want this?

mmoskal: no, should be always on, not sure about the syntax

libs/core/pxt.json
```diff
@@ -54,6 +54,11 @@
     "public": true,
     "dependencies": {},
     "yotta": {
+        "config": {
+            "microbit-dal": {
+                "message_bus_concurrency_mode": 1
+            }
+        },
```
pelikhan: probably fiber_user_data too?

docs/reference/control/next-handler-to-background.md
```diff
@@ -0,0 +1,42 @@
...
+## Example
+
+This program shows how to register multiple background handlers
+for the ``on button pressed`` event (of button A):
+
+```ts
+// after this program runs, a press of button A
+// will print the sequence 1, 2, 3
+control.nextHandlerToBackground();
+input.onButtonPressed(Button.A, () => {
```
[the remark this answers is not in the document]
mmoskal: I think it was my idea to do it like this. Otherwise we would have to modify all calls that register events to take additional arguments etc...

Issue #1852, opened 2019-02-22
Add offline app to offline documentation.


abchatra:
Microbit offline documentation: https://makecode.microbit.org/offline

Please add information about offline app in the above page. Only Microbit and Mindstorms EV3 has this. 
Offline app is here: https://makecode.microbit.org/offline-app 

We want to make sure it is prominent and it is a pre-release app at this stage.

abchatra:
@pelikhan @Jaqster FYI

pelikhan:
CSS looks broken on the offline-app page.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

abchatra:
Yes we broke it today due to microbit update

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

xuefengedu:
it seems this issue has fixed and completed.

PR #1853, opened 2019-02-22
Extend offline page to include app section

ganicke:
Extend the pxt offline base page to include a section for the offline app.

Move this section to the top above **Web application**?

RE: #1852

### DIFF
--- a/docs/offline.md
+++ b/docs/offline.md
@@ -2,4 +2,4 @@
 
 ## Offline app #target-app
 
-The MakeCode editor is available as app which you can install on a computer with Windows or Mac OS. Once installed, the [MakeCode Offline App](offline-app) lets you create, run, and download your projects to the @boardname@. It works the same as the Web application does in your browser but it's a stand-alone application that will work when a connection to the internet is restricted or not available.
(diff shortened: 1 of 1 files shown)


docs/offline.md
```diff
@@ -0,0 +1,5 @@
+# @extends
+
+## Offline app #target-app
+
+The MakeCode editor is available as app which you can install on a computer with Windows or Mac OS. Once installed, the **[MakeCode Offline App](offline-app)** lets you create, run, and download your projects to the @boardname@. It works the same as the Web application does in your browser but it's a stand-alone application that will work when a connection to the internet is restricted or not available.
```
pelikhan: /offline-app?

pelikhan:
Yes we should move the offline app on top of the page and use a full url in the link instead of just a hyperlink imo. The offline page reads a bit funny right now.

docs/offline.md
ganicke: Yea, oops.

ganicke:
Right, this was originally done for the offline browser use.

pelikhan:
Right it evolved :)

abchatra:
@ganicke Can we do the same change for https://makecode.mindstorms.com/offline as well?

ganicke:
I think so. I take a look that one too.

PR #1855, opened 2019-02-22
Fix link format for offline-app page

ganicke:
Add the leading slash '/' for page path.

Should I add a mention that the app is BETA? -- https://github.com/Microsoft/pxt-microbit/issues/1852#issue-413189460

RE: #1853


ganicke:
Want to also invite user comments/suggestions/issues in the pre-release note?

PR #1856, opened 2019-02-23
Split MakerBit extension into stand-alone packages

philipphenkel:
@microbit-mark @abchatra We split our approved MakerBit extension into stand-alone extensions. This way we are more memory efficient and allow for easy re-use of our modules even when users do not have our board. 
The "master" extension still exists and includes our modules as dependencies (BT disables as a compromise).


microbit-mark:
These look good to me @abchatra

abchatra:
Sorry for the bit delay. @philipphenkel Is this good to merge?

philipphenkel:
@abchatra Yes, looks good from my perspective.

pelikhan:
@philipphenkel please check that the examples in your README.md compile.

philipphenkel:
@pelikhan Thanks for the merge! I will check all example code today.

Issue #1929, opened 2019-03-20
setRxBufferSize Missing


ryanteck:
**Is your feature request related to a problem? Please describe.**
The serial buffer has a default of 20 characters set, for many applications more than 20 are required. It seems to be there in some parts of the library but not accessible.

**Describe the solution you'd like**
Implementation of the stRxBufferSize function with a block & JS  / Typescript

**Describe alternatives you've considered**
N/A

**Additional context**
For us its quite a big issue as it prevents our Micro:bit add-on from functioning at its full potential.

AmedeeBulle:
I second this request.

The `serial.onDataReceived()` event handler and `serial.readUntil()`  function are currently unusable when you have lines with more than 20 characters (you will get truncated lines at best).

A workaround is to constantly poll the serial with `serial.readString()` is a separate fiber, but it is not very practical (and you can easily miss characters)

bosnivan:
I third this request. Have same problems.

andybatey:
+1 this request.  This issue limits my usage model.

pelikhan:
This function is in /beta.

whaleygeek:
Great, glad to see this has finally been dealt with, well done.

AmedeeBulle:
Tested in my project, works fine, thanks!

(There are no blocks though, but it doesn't matter for me as it is for a library)

PR #1931, opened 2019-03-21
Adding serial.set(Rx|Tx)BufferSize

pelikhan:
* Fix for https://github.com/Microsoft/pxt-microbit/issues/1929
* use React for dialogs.


ryanteck:
Thanks for this quick fix :+1:

PR #1939, opened 2019-03-26
Update the max length of a name in a sendValue packet

riknoll:
It went from 12 characters to 8 when we started supporting floating points. This PR fixes the simulator, JS doc, and reference for the function. @ganicke feel free to push any changes to this branch!


riknoll:
Actually, looks like there is also a bug in radio.cpp. I'm going to close this and open a new PR

PR #1953, opened 2019-04-04
Remove c++ macros that are moving to common packages

jwunderl:
see https://github.com/Microsoft/pxt-common-packages/pull/782, need to be using a version of common packages that includes that PR before this can be merged

### DIFF
--- a/libs/core/platform.h
+++ b/libs/core/platform.h
@@ -1,28 +1,3 @@
 // helpful define to handle C++ differences in package
 #define PXT_MICROBIT_TAGGED_INT 1
-
-// cross version compatible way of access data field
-#ifndef PXT_BUFFER_DATA
-#define PXT_BUFFER_DATA(buffer) buffer->data
-#endif
-
-// cross version compatible way of access data length
-#ifndef PXT_BUFFER_LENGTH
-#define PXT_BUFFER_LENGTH(buffer) buffer->length
-#endif
-
-#ifndef PXT_CREATE_BUFFER
-#define PXT_CREATE_BUFFER(data, len) pxt::mkBuffer(data, len)
-#endif
-
-// cross version compatible way of accessing string data
-#ifndef PXT_STRING_DATA
-#define PXT_STRING_DATA(str) str->getUTF8Data()
-#endif
-
-// cross version compatible way of accessing string length
-#ifndef PXT_STRING_DATA_LENGTH
-#define PXT_STRING_DATA_LENGTH(str) str->getUTF8Size()
-#endif
-
 #define PXT_POWI 1
--- a/libs/core/platform.h
+++ b/libs/core/platform.h
@@ -3,7 +3,7 @@
 
 // cross version compatible way of access data field
 #ifndef PXT_BUFFER_DATA
-#define PXT_BUFFER_DATA(buffer) buffer->data
+#define PXT_BUFFER_DATA(buffer) buffer->getUTF8Data()
 #endif
 
 #ifndef PXT_CREATE_BUFFER
--- a/libs/core/platform.h
+++ b/libs/core/platform.h
@@ -21,8 +21,8 @@
 #endif
 
 // cross version compatible way of accessing string length
-#ifndef PXT_STRING_LENGTH
-#define PXT_STRING_LENGTH(str) str->getLength()
+#ifndef PXT_STRING_DATA_LENGTH
+#define PXT_STRING_DATA_LENGTH(str) str->getUTF8Size()
 #endif
 
 #define PXT_POWI 1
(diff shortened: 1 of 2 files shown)


jwunderl:
Ah, just saw your comment on the other one, good timing. I'll add a macro for size here too then?

pelikhan:
does the extension build?

jwunderl:
Yes, running on ``pxt serve --local`` and it builds, versus beta which is broken. Don't think I have any hardware of theirs to check, though

pelikhan:
You need to clone the extension repo under /projects and run "pxt build --local" from that folder.

jwunderl:
Oh, so they were relying on the parity between buffer and boxed string which was broken, and this was still correct? That makes sense, I'll switch this to just add macros for string then, and then find where they used it and change that over?

took a bit but got docker set up, had to use the yotta link it gave while building as the pxt github login command wasn't working. It does seem to build fine with this (broken) fix, besides a few warnings:

```
[187/190] Building CXX object source/CMakeFiles/pxt-microbit-app.dir/src/source/core/pins.cpp.o
[188/190] Building CXX object source/CMakeFiles/pxt-microbit-app.dir/src/source/core/codal.cpp.o
In file included from /src/source/SSD1306_OLED/Adafruit_SSD1306.h:28:0,
                 from /src/source/SSD1306_OLED/Adafruit_SSD1306.cpp:24:
/src/source/SSD1306_OLED/Adafruit_GFX.h: In member function 'virtual void Adafruit_GFX::invertDisplay(bool)':
/src/source/SSD1306_OLED/Adafruit_GFX.h:71:37: warning: unused parameter 'i' [-Wunused-parameter]
     virtual void invertDisplay(bool i) {};
                                     ^
/src/source/SSD1306_OLED/Adafruit_SSD1306.cpp: In member function 'virtual void Adafruit_SSD1306::shiftDisplay(int16_t, int16_t)':
/src/source/SSD1306_OLED/Adafruit_SSD1306.cpp:221:11: warning: suggest parentheses around operand of '!' or change '&' to '&&' or '!' to '~' [-Wparentheses]
         !(!buffer[nx + (ny>>3)*_rawWidth] & (1<<(ny&7))) ){
            
[189/190] Building CXX object source/CMakeFiles/pxt-microbit-app.dir/src/source/core/control.cpp.o
/src/source/core/codal.cpp: In function 'void pxt::gcProcessStacks(int)':
/src/source/core/codal.cpp:251:9: warning: unused variable 'cnt' [-Wunused-variable]
     int cnt = 0;
         ^~~
[190/190] Linking CXX executable source/pxt-microbit-app
package built; written to binary.hex
```

pelikhan:
Yes we should add additional macros for strings and potentially patch their repo to use those instead.

libs/core/platform.h
```diff
@@ -3,11 +3,21 @@
 
 // cross version compatible way of access data field
 #ifndef PXT_BUFFER_DATA
-#define PXT_BUFFER_DATA(buffer) buffer->getUTF8Data()
+#define PXT_BUFFER_DATA(buffer) buffer->data
 #endif
```
pelikhan: could you also add a PXT_BUFFER_LENGTH macro for the sake of completeness

libs/core/platform.h
```diff
@@ -3,11 +3,21 @@
...
 
 #ifndef PXT_CREATE_BUFFER
 #define PXT_CREATE_BUFFER(data, len) pxt::mkBuffer(data, len)
 #endif
 
+// cross version compatible way of accessing string data
+#ifndef PXT_BOXED_STRING_DATA
+#define PXT_BOXED_STRING_DATA(boxedString) boxedString->getUTF8Data()
```
pelikhan: just call it PXT_STRING_DATA and PXT_STRING_LENGTH

jwunderl: Assuming ``PXT_STRING_LENGTH`` should grab ``->getLength()`` instead of the size, is there any value in adding size too that grabs ``->getUTF8Size`` ?

pelikhan: callit PXT_STRING_DATA_LENGTH

pelikhan:
Maybe those macros should be moved to pxt-common-packages, next to the C++ declaration of Buffer/String.

jwunderl:
That definitely sounds reasonable to me, but do we have any context on why ``PXT_BUFFER_DATA`` and ``PXT_CREATE_BUFFER`` were added here rather than in common packages in the first place? I'm not seeing any clues from https://github.com/Microsoft/pxt-microbit/pull/1165

pelikhan:
Probably because it requires 1 less bump to do it in microbit.

jwunderl:
Fair. I'll bump the buffer and string macros over to common packages then

jwunderl:
Should be good to go  now, this is now just bumping common packages and removing the macros that got moved

PR #1956, opened 2019-04-04
Switching to draggable radio blocks

riknoll:
Switching to draggable parameters in the radio blocks.
dragging_radio

The new workflow for using radio blocks is to drag the variables out of the event block itself. The variables will not show up in the variable category of the toolbox and cannot be renamed or used outside of the radio event (they are "local").

Here's the impact for existing projects:
1. All scripts that use the old radio blocks will keep using those blocks (they will **not** be automatically upgraded)
2. The old blocks will map to the deprecated functions in the TS and the new blocks will compile to the same non-deprecated TS functions as before
3. Any JavaScript projects using the old functions will get the new blocks if they ever switch to blockly

@ganicke are there any update needed to the docs for this?

### DIFF
--- a/libs/radio/radio.ts
+++ b/libs/radio/radio.ts
@@ -74,8 +74,8 @@ namespace radio {
      * Registers code to run when the radio receives a number.
      */
     //% help=radio/on-received-number blockHandlerKey="radioreceived"
-    //% blockId=radio_on_number block="on radio received" blockGap=16
-    //% useLoc="radio.onDataPacketReceived"
+    //% blockId=radio_on_number_drag block="on radio received" blockGap=16
+    //% useLoc="radio.onDataPacketReceived" draggableParameters=reporter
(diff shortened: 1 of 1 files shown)


pelikhan:
Spec looks good. Do we support multiple types of "on received handler" with this change?

pelikhan:
Never mind, that is unrelated to this change.

pelikhan:
@riknoll  What happens to legacy project who did rename the parameter?

abchatra:
Legacy blocks stay.

riknoll:
If they were using the old blocks, nothing happens (the project still uses the same blocks).

If they renamed via the TS and then decompiled, we pick up the new name. For example:

dragging_radio_rename

pelikhan:
@ganicke do we have any tutorial/guide in the markdown docs that rename radio variables?

ganicke:
I don't think there's a mention ability to change the parm name. I'll have to note that for those blocks.

Issue #1637, opened 2018-11-14
Block to convert number to string


microbit-giles:
**Is your feature request related to a problem? Please describe.**
There is a 'parse to number' block for converting strings to numbers, but apparently no block to do the reverse, convert numbers to strings (like str(number) in Python, for example).

**Describe the solution you'd like**
I would like there to be a matching 'parse to string' block.

**Describe alternatives you've considered**
I'm told you can use the 'join' blocks convert numbers to strings but this is highly unintuitive and could not see any reference to this method in the documentation for 'parse to number'. 

**Additional context**
At the very least could the 'join' block workaround be mentioned in the 'parse to number' documentation? It seems logical to me that in describing the way to convert strings to numbers, there should at least be a link to the reverse operation.

guillaumejenkins:
Can you describe your use case for converting a number to a string?

microbit-giles:
Good question - not off the top of my head! As part of my work I am learning more about MakeCode so I can explain it to educators and prepare learning resources and it occurred to me that most programming languages offer conversion both ways, and that MakeCode should follow suit. It seemed odd to have one but not the other.

guillaumejenkins:
Right, it seems like teaching string conversion is one of the only things this would be useful for, especially since all the micro:bit APIs that take a string (Show String, Radio Send String, etc) already accept number arguments (they'll be automatically converted).

Our stance on teaching more advanced programming concepts has always been that it should happen in the JavaScript editor, and type conversion falls in that category in my opinion. Also note that the Join block compiles down to `"" + <your number here>`, which is exactly what a "To String" block would do.

It seems like the real issue here is that "Join" does not intuitively mean "To String". Unfortunately, changing the text on blocks breaks translations and any existing screenshots / course instructions that refer to the old block text, so I'm afraid we're stuck with "Join"...

That being said, mentioning the Join block in the Parse to Number documentation is certainly a good idea!

microbit-giles:
Thanks, would be great to mention the workaround in the docs.

There are other blocks that perform what I regard as highly advanced functions and it just seemed odd to me that there's a specific block for going one way but not the other, but a reference in the documentation would be very helpful!

pelikhan:
The text category should have a “to String” block that takes any expression block, just like JavaScript.

abchatra:
@pelikhan are you taking a look at this?

pelikhan:
Not at the moment.

[...]

PR #1958, opened 2019-04-04
Add toString block

jwunderl:
Have to duplicate this function from https://github.com/Microsoft/pxt-common-packages/pull/781 as microbit overrides / doesn't use that file

fixes https://github.com/Microsoft/pxt-microbit/issues/1637


docs/reference/text/to-string.md
```diff
@@ -0,0 +1,7 @@
+# to string
+
+Converts a value into a text representation.
+
+```sig
+toString(123)
```
pelikhan: Does this generate the correct sig?

libs/core/control.ts
```diff
@@ -53,3 +53,14 @@ namespace control {
...
 }
+
+/**
+ * Convert any value to a string
+ * @param value value to be converted to a string
+ */
+//% help=text/to-string weight=1
+//% block="convert $value=variables_get(myValue) to string"
+//% blockId=variable_to_string blockNamespace="text"
+function toString(value: any): string {
```
pelikhan: There is no global "toString" function in JavaScript, so does this block decompile to an instance ``.toString()`` call?

docs/reference/text/to-string.md
jwunderl: Screen Shot 2019-04-04 at 11 39 24 AM
yup

PR #1959, opened 2019-04-04
adding led game sprite

pelikhan:
- [ ] we should remove the namespace when displaying the type

image


abchatra:
@Jaqster

Jaqster:
yes, let's add for completeness

pelikhan:
Removing namespace requires changes somewhere in pxt.

PR #1965, opened 2019-04-05
Radio max name length

riknoll:
Setting to 8 and fixing the sim.


pelikhan:
We should consider completely rewriting this in TS

PR #1977, opened 2019-04-05
default to num instead of variable for toString

jwunderl:
Screen Shot 2019-04-05 at 2 22 10 PM

Show number instead of a variable to start with

I'll have to dup this over to pxt-common-packages as well

### DIFF
--- a/docs/reference/text/convert-to-text.md
+++ b/docs/reference/text/convert-to-text.md
@@ -1,7 +1,7 @@
-# to text
+# convert to text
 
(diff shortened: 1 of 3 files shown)


abchatra:
Also this API needs to be called "convert ___ to text". Text is the name of the category and used in other api (text from char code). 

substring API uses string, I guess we can leave it as is for now

jwunderl:
Switched over to toText

Screen Shot 2019-04-05 at 2 41 35 PM

PR #1980, opened 2019-04-07
Fill in 'to-string' ref page

ganicke:
Add some fill into 'toString()' ref page.

Depends on #1977


jwunderl:
Minor note, block is changing slightly with #1977

ganicke:
Well, I'll switch-chanjo then.

ganicke:
Closing this in favor of #2002.

PR #1873, opened 2019-02-27
support for is-gesture

pelikhan:
Support for https://github.com/Microsoft/pxt-microbit/issues/1870
gallery

PR #1983, opened 2019-04-08
Fix shake button

### LINKED ISSUE #1970: Some buttons in simulator require double click
**Describe the bug**

Some elements of the simulator require double clicks on /beta (looks like one to focus then one to input) that don't on the live site. Example with shake: (top is safari, bottom is chrome)

same with light sensor and other simulator button, but only in safari

**To Reproduce**

1. Go to editor in safari
2. add a block that uses a simulator feature (here, on shake with anything in it)
3. see error (need to double click instead of single click on other things)

 - OS: MacOS (and also iOS I think)
 - Browser safari

### LINKED ISSUE #1971: Shake simulator button not working on ipad in /beta
**Describe the bug**

shake button is not working at all on ipad in /beta

**To Reproduce**
1. Go to iOS safari editor (I'm using xcode simulator, but reproed on test iPad)
2. add ``on shake`` block that does anything
3. press shake button

**Expected behavior**

should simulate 'shake'

**Screenshots**

 - Device: iPad pro (simulator)
 - OS:  iOS 12.1
 - Browser safari

jwunderl:
Fixes #1970 
Fixes #1971 

Issue is that the ``sigma`` for the gesture is getting reset to 0 immediately before it is tested, even though it was just set to be a valid sigma in the ``forceGesture`` method.

### DIFF
--- a/sim/state/accelerometer.ts
+++ b/sim/state/accelerometer.ts
@@ -261,7 +261,7 @@ namespace pxsim {
             this.setGesture(g);
         }
 
-        private setGesture(g: number) {
+        private setGesture(g: number, force?: boolean) {
             // Perform some low pass filtering to reduce jitter from any detected effects
             if (g == this.currentGesture) {
                 if (this.sigma < DAL.MICROBIT_ACCELEROMETER_GESTURE_DAMPING)
@@ -273,7 +273,7 @@ namespace pxsim {
             }
 
             // If we've reached threshold, update our record and raise the relevant event...
-            if (this.currentGesture != this.lastGesture && this.sigma >= DAL.MICROBIT_ACCELEROMETER_GESTURE_DAMPING) {
+            if (this.currentGesture != this.lastGesture && (this.sigma >= DAL.MICROBIT_ACCELEROMETER_GESTURE_DAMPING || force)) {
                 this.lastGesture = this.currentGesture;
                 board().bus.queue(DAL.MICROBIT_ID_GESTURE, this.lastGesture);
             }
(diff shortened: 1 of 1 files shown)


abchatra:
Which change introduced this bug?

jwunderl:
https://github.com/Microsoft/pxt-microbit/pull/1873

sim/state/accelerometer.ts
```diff
@@ -273,15 +273,14 @@ namespace pxsim {
...
+            if (force || (this.currentGesture != this.lastGesture && this.sigma >= DAL.MICROBIT_ACCELEROMETER_GESTURE_DAMPING)) {
                 this.lastGesture = this.currentGesture;
                 board().bus.queue(DAL.MICROBIT_ID_GESTURE, this.lastGesture);
             }
         }
 
         forceGesture(g: number) {
-            this.sigma = DAL.MICROBIT_ACCELEROMETER_GESTURE_DAMPING + 1;
-            this.setGesture(g);
+            this.setGesture(g, /* force */ true);
```
abchatra: It took me a while to understand setGesture. I think we could just simply this change by adding 3 lines at forceGesture
```javascript
        forceGesture(g: number) {
            this.sigma = 0;
            this.lastGesture = this.currentGesture = g;
            board().bus.queue(DAL.MICROBIT_ID_GESTURE, this.lastGesture);
}
```

jwunderl: Yeah, I wasn't 100% happy with adding it as an optional parameter, but wanted both to funnel through the same ``bus.queue`` call as that seems like pretty specific logic to duplicate. I'll give it a second pass to see how it can be refactored more clearly.

jwunderl:
Tested with mac chrome and iOS safari and it works with latest changes, which I think are a bit easier to parse

ezgif com-crop (1)

abchatra:
Thanks for making it better

Issue #1765, opened 2018-12-28
All radio receive blocks are permisive


rhubarbdog:
**Describe the bug**
All radio receive blocks are permissive, well they assume each others identity \
without warning. Block `radio.onReceivedValue(name, value)` acts as a va\
riant block. It receives messages (19 character), integers, doubles and value, \
number pair blocks as 8 chars preceded by an int 32 or a 8 byte double.
Likewise a `radio.onReceivedString(receivedString)` receives the name portion o\
f a sent value and `radio.onReceivedNumber(receivedNumber)` receives the numeri\
c porion of a sent value.

Send value truncates the name to 8 character for and integer and 11 characters \
for a double precision value without warning and only receives 8 regardless of \
what was sent and the `name length byte` is ignored.  When I send a double prec\
ision value/name pair with an over long	name 11	characters are sent but	the
 `name length byte` is incorrectly set to 12

**To Reproduce**
On 2 different microbits, <a href="https://makecode.microbit.org/_9LoATg7jLRtD">on one load </a>these blocks and <a href="https://makecode.microbit.org/_EVELdi6iFhcu">on the second</a> these blocks
on microbit 1 press A this sends a string; `hello world` `0` received
on microbit 1 press B that sends a  number; `` (blank) `-99` received
on microbit 1 press A+B that sends a value pair; `val_pair` `-256` correctly received

**Expected behavior**
Microbit 2 should only receive the data when A+B is pressed on Microbit 1. in the other 2 cases either nothing should happen or the display should scroll an error message.

**Desktop (please complete the following information):**
 - OS: Linux mint 18.2
 - Browser : mozilla firefox
 - Version : 63.0.3

rharbird:
Like @rhubarbdog, I too am experiencing problems with this feature. I am trying to use several name, value pairs for reporting waterflow in ml for a sink in a bathroom :

"this-use-ml" : number
"average-use-ml": number
"total-use-ml": number

I have found that the block radio.onReceivedValue(name, value) does not behave correctly.  I have tried using:

-  3 separate if statements to parse the packet received,
- an if, else if construct and,
- a switch statement.

The code works correctly in the simulator but does not work when uploaded onto 2 boards. I'm guessing now that the problems are likely related to the way in which 'name' is parsed.

This is an extremely useful function for IoT-type applications so I would appreciate it if it could be fixed because I use the micro:bit for teaching non-CS students at undergraduate level and it's really useful!


Kind Regards,


Rae

[...]

PR #2000, opened 2019-04-11
Rewriting the radio in TypeScript

riknoll:
Fixes https://github.com/Microsoft/pxt-microbit/issues/1765

As the title says, this PR takes all of the packet reading/writing and moves it from the C++ layer to TypeScript. This does not change the radio protocol, so old hex files will still work with new ones.

It also tweaks the radio behavior to make it much easier to use multiple message types; currently any `onReceived` block will fire for any packet even if it doesn't correspond to the payload that was sent. In other words, if you received a message sent by `sendString()` then `onReceivedNumber()` would still get called even though you didn't receive a number. This is technically a breaking change for that reason but the impact should be minimal because we don't allow you to mix/match event handlers in the blocks (this PR removes that restriction as well).
 
Also, the old cpp code had some bugs around sending doubles and UTF-8 strings and moving everything to the TS fixes a few hard crashes. The test script I checked in fails on the current hardware.


libs/radio/test.ts
```diff
@@ -0,0 +1,352 @@
+/**
+ * Tests for the radio. Press A on mbit 1 and B on mbit 2 to run the tests.
+ * Sends random ints, doubles, strings, and buffers and checks them on
+ * the other side
+ */
+
+class FastRandom {
```
mmoskal: is our built-in random so slow that this makes a difference?

riknoll: No, it just needs to be deterministic. I needed a random number generator that I could seed so that I could generate the test cases on each side as the come in and verify that they match

mmoskal: oh, cool

mmoskal:
Also, score! 2000!

libs/radio/radio.cpp
```diff
@@ -419,7 +81,7 @@ namespace radio {
     //% blockId=radio_datagram_rssi block="radio received signal strength"
     //% deprecated=true
     int receivedSignalStrength() {
-        if (radioEnable() != MICROBIT_OK) return 0;
+        if (radioEnable() != MICROBIT_OK || packet == NULL) return 0;
```
pelikhan: Move to TS?

sim/state/radio.ts
```diff
@@ -115,70 +108,17 @@ namespace pxsim.radio {
...
-        });
-    }
-
-    export function receiveNumber(): number {
-        const packet = board().radioState.datagram.recv();
-        return receivedNumber();
-    }
-
-    export function receiveString(): string {
+    export function _takePacket() {
```
pelikhan: __ to have it ignored by Monaco I think

libs/radio/radio.ts
```diff
@@ -16,182 +16,124 @@ enum RadioPacketProperty {
...
+    const PACKET_TYPE_STRING = 2;
+    // payload: buffer length (9), buffer (10 ... 28)
+    const PACKET_TYPE_BUFFER = 3;
+    // payload: number (9 ... 16)
+    const PACKET_TYPE_DOUBLE = 4;
+    // payload: number (9 ... 16), name length (17), name (18 ... 26)
+    const PACKET_TYPE_DOUBLE_VALUE = 5;
+
+    interface RadioCallbacks {
+        numberListeners?: ((receivedNumber: number) => void)[];
```
pelikhan: We only support 1 of each. I would allocate a new component id for the radio callbacks and use a different event id for each type. Then it would fit nicely in the bus.

libs/radio/shims.d.ts
```diff
@@ -8,80 +8,24 @@ declare namespace radio {
...
-     * Reads the next packet from the radio queue and returns the packet's number
-     * payload or 0 if the packet did not contain a number.
-     */
-    //% help=radio/receive-number
-    //% weight=46
-    //% blockId=radio_datagram_receive block="radio receive number" blockGap=8
-    //% deprecated=true shim=radio::receiveNumber
-    function receiveNumber(): number;
+    //% advanced=true shim=radio::_sendRawPacket
+    function _sendRawPacket(msg: Buffer): void;
```
pelikhan: Some users might like this, is there a “readRawPacket”?

libs/radio/radio.cpp
```diff
@@ -81,303 +35,27 @@ namespace radio {
...
-    */
-    //% help=radio/write-received-packet-to-serial
-    //% weight=3
-    //% blockId=radio_write_packet_serial block="radio write received packet to serial"
-    //% advanced=true
-    void writeReceivedPacketToSerial() {
-        if (radioEnable() != MICROBIT_OK) return;
-        writePacketAsJSON(type, ivalue, dvalue, (int) serial, (int) time, msg, bufMsg);
-    }
+        uint8_t buf[msg->length];
```
pelikhan: Can we send the msg->data directly? Need to check that the DAL makes a copy.

libs/radio/radio.cpp
```diff
@@ -81,303 +35,27 @@ namespace radio {
...
-
-        setPacketPrefix(buf, PACKET_TYPE_STRING);
-        int stringLen = copyStringValue(buf + PACKET_PREFIX_LENGTH, msg, MAX_PAYLOAD_LENGTH - 1);
-
-        uBit.radio.datagram.send(buf, PACKET_PREFIX_LENGTH + stringLen);
+    //% help=radio/received-packet
+    Buffer _takePacket() {
+        if (radioEnable() != MICROBIT_OK) return mkBuffer(NULL, 0);
+        packet = uBit.radio.datagram.recv();
+        return mkBuffer(packet.getBytes(), 32);
```
pelikhan: I think we should use https://github.com/lancaster-university/microbit-dal/blob/2cff906f019944ae15d532157c1301c940f67862/inc/types/PacketBuffer.h#L242 . The radio might returns an empty packet (of size 1)

darzu:
Are there any measurable changes to user program compile time since there's more STS now?

libs/radio/radio.cpp
mmoskal: I don't think we can use codal types directly anymore. They would not be gc tracked.

sim/state/radio.ts
riknoll: No, single works

libs/radio/radio.cpp
riknoll: I don't think I can; iIt is part of the CODAL class and not in the buffer itself.

libs/radio/radio.cpp
riknoll: [yep](https://github.com/lancaster-university/microbit-dal/blob/2cff906f019944ae15d532157c1301c940f67862/source/drivers/MicroBitRadioDatagram.cpp#L130)

libs/radio/shims.d.ts
riknoll: Renamed takePacket to readRawPacket() and removed the underscores

libs/radio/radio.cpp
riknoll: Switched to length. @mmoskal I think peli just meant the length function, not sending the packet buffer to the TS

libs/radio/radio.ts
riknoll: How do I allocate a new component ID? Do I need to do a PR on the DAL?

riknoll:
@darzu do we have any hooks for measuring compile time? I certainly didn't notice any slowdown.

libs/radio/radio.ts
pelikhan: Use the existing MICROBIT_Radio_ID value

libs/radio/radio.cpp
pelikhan: correct, use length instead of 32.

riknoll: Done

libs/radio/radio.ts
riknoll: Done

riknoll:
@pelikhan refactored to use event bus, please review

Issue #1752, opened 2018-12-19
Infection project is different on hardware


microbit-mark:
microbit-support:19134

**Describe the bug**
https://makecode.microbit.org/projects/infection displays assigned user letters in simulator but not on the device

**To Reproduce**
1. Go to 'https://makecode.microbit.org/projects/infection'
2. Select JS and copy to makecode or click on makecode link
3. Press A+B in simulator and programme is initiated, showing number of connected users and letter on additional devices.
4. Flash same program to micro:bit
4. Press A+B, number of connected users is shown on master device, but additional micro:bits show smiley face and not assigned letter. 

**Expected behavior**
Docs says "When paired, a letter appears on the player’s screen, this letter is the player’s identity." This is expected behaviour

rhl2401:
Anyone who have solved this problem?

abchatra:
@pelikhan

Issue #1640, opened 2018-11-14
Sending serial number via radio converts it to a float


microbit-mark:
support ticket: 18385

**Describe the bug**
When sending a serial number via radio, if the serial number is a positive integer, the output adds additional decimal places.

serial is 0x791b328b  (2031825547 in decimal) is received as 2031825546.9999999
serial is 0x761b118b  (1981485451 in decimal) is received as 1981481450.999999
serial is -0x7292b83e (1922218046 in decimal) is recieved as 1922218046

**To Reproduce**
1. Go to 'https://makecode.microbit.org/_HCb2j14b199k '
2. Download to two micro:bits (making sure one has a positive serial number initially)
3. Press A on one
4. See serial number displayed on the other with added decimal places and over serial USB.

**Expected behavior**
Serial number should display as an integer

This is the behaviour in makecode and beta 

makecode.microbit.org version:  1.2.9

Microsoft MakeCode version:  4.6.2

microbit runtime version:  v2.1.0

**Additional info**
Not sure if this is a duplicate of #1535 and the fix you mentioned on the call @pelikhan @abchatra 

cc @microbit-carlos

pelikhan:
This is a different bug.

microbit-mark:
Nudging this one again @pelikhan @abchatra. Serial numbers with added decimal places is confusing.

microbit-mark:
tagged with next-release as discussed @abchatra

ChaseMor:
Looks like the issue comes from printing a number that is larger than 2^30.

abchatra:
We don't support integers larger than 2^30 as it doesn't fit in tagged int strategy and will be converted to float. Floating point values don't have exact precision. This is a limitation in makecode. 

We won't be able to fix from our end.

martinwork:
@microbit-mark Would it be useful to provide access to the 5 character micro:bit friendly name derived from the received serial number?

osupk8:
Can ou give an example ?

martinwork:
For the host micro:bit, control.deviceName is the friendly name (eg. tovip), derived from control.deviceSerialNumber, which corresponds to the A+B+Reset LED pattern used for Bluetooth pairing. I meant the equivalent of deviceName for the received serial number. I'm imagining a "device name" choice on the received packet block dropdown.

abchatra:
Convert to string and send as a string? Even that has 8 character limit, but you can send multiple parts of as the string.

[...]

PR #2003, opened 2019-04-12
Fixes for infection game

ChaseMor:
Fixes Microsoft/pxt-microbit#1752

The previous infection game relied on sending the devices' serial number through the field parameter as a string. Since this string is limited to only a few characters, the serial number was getting cut off so when the player compared their serial number with the one sent, it didn't match.

The problem is made worse since the toString on serial numbers results in adding more significant figures as shown in Microsoft/pxt-microbit#1640.

Since the values it sends during these steps are the icon/player number and health (as a value of 0 to 3), it is simple enough to append the value to the field name and send the serial numbers as the value.

### DIFF
--- a/docs/projects/infection.md
+++ b/docs/projects/infection.md
@@ -147,52 +147,48 @@ const GameIcons = {
 
 class Message {
 
-    static decrypt(input: Buffer): Message {
-        if (input.length != 13) {
-            return null;
-        }
-        return new Message(input.getNumber(NumberFormat.Int8LE, 0),
-            input.getNumber(NumberFormat.Int32LE, 1),
-            input.getNumber(NumberFormat.Int32LE, 5),
-            input.getNumber(NumberFormat.Int32LE, 9));
+    private _data: Buffer;
+
+    constructor(input?: Buffer) {
+        this._data = input || control.createBuffer(13);
     }
 
-    private data: Buffer;
+    get kind(): number {
+        return this._data.getNumber(NumberFormat.Int8LE, 0);
+    }
 
-    constructor(kind: number, serialNumberFrom: number,
-        value?: number, serialNumberTo?: number) {
-        this.data = control.createBuffer(13);
-        // 1 byte for kind
-        this.data.setNumber(NumberFormat.Int8LE, 0, kind);
-        // 4 bytes for from
-        this.data.setNumber(NumberFormat.Int32LE, 1, serialNumberFrom);
-        // 4 bytes for value
-        this.data.setNumber(NumberFormat.Int32LE, 5, value);
-        // 4 bytes for to
-        this.data.setNumber(NumberFormat.Int32LE, 9, serialNumberTo);
+    set kind(x: number) {
+        this._data.setNumber(NumberFormat.Int8LE, 0, x);
+    }
+
+    get fromSerialNumber(): number {
+        return this._data.getNumber(NumberFormat.Int32LE, 1);
+    }
+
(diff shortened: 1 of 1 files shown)


docs/projects/infection.md
```diff
@@ -69,7 +69,7 @@ As a result, it will not convert back to blocks.
 
 ### ~
 
-https://makecode.microbit.org/_gymCJCWPbiDu
+https://makecode.microbit.org/_5dfL87Wa11so
```
pelikhan: Let’s get rid of that. People can copy it.

docs/projects/infection.md
```diff
@@ -137,6 +145,56 @@ const GameIcons = {
     Healthy: IconNames.Happy
 }
 
+class Message {
+
+    static decrypt(input: Buffer): Message {
```
pelikhan: This is not needed - simply wrap the buffer in a message class.

docs/projects/infection.md
```diff
@@ -137,6 +145,56 @@ const GameIcons = {
...
+            input.getNumber(NumberFormat.Int32LE, 9));
+    }
+
+    private data: Buffer;
+
+    constructor(kind: number, serialNumberFrom: number,
+        value?: number, serialNumberTo?: number) {
+        this.data = control.createBuffer(13);
+        // 1 byte for kind
+        this.data.setNumber(NumberFormat.Int8LE, 0, kind);
```
pelikhan: use a set function?

docs/projects/infection.md
```diff
@@ -137,6 +145,56 @@ const GameIcons = {
...
+        // 4 bytes for from
+        this.data.setNumber(NumberFormat.Int32LE, 1, serialNumberFrom);
+        // 4 bytes for value
+        this.data.setNumber(NumberFormat.Int32LE, 5, value);
```
pelikhan: Use a set function

docs/projects/infection.md
```diff
@@ -137,6 +145,56 @@ const GameIcons = {
...
+        // 4 bytes for to
+        this.data.setNumber(NumberFormat.Int32LE, 9, serialNumberTo);
+    }
+
+    send() {
+        radio.sendBuffer(this.data);
+        basic.pause(250)
```
pelikhan: needed?

ChaseMor:
Do you mean like this?

```typescript
class Message {

    private data: Buffer;

    constructor(input?: Buffer) {
        this.data = input || control.createBuffer(13);
    }

    get kind(): number {
        return this.data.getNumber(NumberFormat.Int8LE, 0);
    }

    set kind(x: number) {
        this.data.setNumber(NumberFormat.Int8LE, 0, x);
    }

    get fromSerialNumber(): number {
        return this.data.getNumber(NumberFormat.Int32LE, 1);
    }

    set fromSerialNumber(x: number) {
        this.data.setNumber(NumberFormat.Int32LE, 1, x);
    }

    get value(): number {
        return this.data.getNumber(NumberFormat.Int32LE, 5);
    }

    set value(x: number) {
        this.data.setNumber(NumberFormat.Int32LE, 5, x);
    }

    get toSerialNumber(): number {
        return this.data.getNumber(NumberFormat.Int32LE, 9);
    }

    set toSerialNumber(x: number) {
        this.data.setNumber(NumberFormat.Int32LE, 9, x);
    }

    send() {
        radio.sendBuffer(this.data);
    }
}
```

pelikhan:
Yup

Issue #2017, opened 2019-04-22
Radio default group


Jaqster:
I'm proposing this idea - Radio Group should be implicitly set by default to group 1.  
So, in a classroom, the default behavior is everyone is broadcasting to each other over the same channel.
When people use the Radio Set Group block in their code, then we assume they are explicitly going to be using different groups to communicate.

abchatra:
+1 @pelikhan I am up for this change. What do you think?

pelikhan:
It is a breaking change. It was designed to avoid random noise from other groups so unless we have a strong reason to change it, i would leave it as is.

abchatra:
How is it breaking change?

pelikhan:
In a classroom, script that used to be mostly independent will now clash on the same radio group.

abchatra:
No one knows that we assign a magic group consists of hash to radio group.

abchatra:
I believe it is not a breaking change. Default group 1 is broadcast group.

pelikhan:
It is a breaking change. Students will start to receive messages from the entire classroom -- this did not happen in the past.

microbit-mark:
I can see the case for this as well as leaving it as is. It's a good starter for a conversation about wireless communication, to think about why we would set a channel to avoid noise and when everyone broadcasts to each other you get the equivalent of a Denial Of Service, so we've used this in cybersecurity lessons before. Be interested in @microbit-giles's perspective.

microbit-giles:
In my own teaching I have used the 'noise' / 'denial of service' effect to introduce the need for groups, using the fact that everyone gets everyone else's message as a learning point, but this is because I have mostly done radio classwork in Python where this, I believe, is the default behaviour. Like Mark I can see cases for both sides of this - a benefit of changing to a default group would be that the experience for users of both MakeCode and Python to be similar.

abchatra:
What is the default group in python?

microbit-giles:
I believe the Python default group is 0: https://microbit-micropython.readthedocs.io/en/latest/radio.html

martinwork:
The automatic group behaviour is such a good idea! If the default is changed, a new block or setting to choose the automatic group behaviour might be helpful.

microbit-mark:
Couple of options: 
- Can we set group to something like 'auto' by default, so people know there's something going on that sets the group automatically
[...]

PR #2037, opened 2019-04-30
Changing the default group to zero

abchatra:
Fixes https://github.com/Microsoft/pxt-microbit/issues/2017

### DIFF
--- a/libs/radio/radio.cpp
+++ b/libs/radio/radio.cpp
@@ -15,7 +15,9 @@ namespace radio {
             return r;
         }
         if (!radioEnabled) {
-            uBit.radio.setGroup(pxt::programHash());
+            // Changing default group to zero, it was program hash before (pxt::programHash());
+            // Python has default group zero as well.
+            uBit.radio.setGroup(0);
             uBit.radio.setTransmitPower(6); // start with high power by default
             radioEnabled = true;
         }


mmoskal:
I don't have a strong opinion, but it is a breaking change.

abchatra:
Yes it is, but no one understood the old behavior. So hoping no one would have taken dependency.

pelikhan:
The current behavior is designed to avoid clashes in a classroom. This is a breaking change.

abchatra:
I understand. It is not working as no one knows a special magical hash is generated. Even mark from foundation didn't know about it.

pelikhan:
Yes it just works. If we make this change, we will have more collisions in a classroom setting so we need to bump up the position of "set group" and document thoroughsly -- otherwise radio programs will trigger randomly once you have a classroom of device streaming packets.

pelikhan:
@finneyj

abchatra:
Sorry I disagree. Currently everyone assumes they have to setGroup to communicate with others. They don't understand that if they flash the same program across multiple microbits it would magically work for those microbits. 

It is a clever trick. Reality is broadcasting on group zero is better. Python does the same.

pelikhan:
This change needs to be documented and "set group" should be moved up in the first set of "sender blocks".

jaustin:
Can we do some specific in-classroom testing of both behaviours before putting this change live please? I've seen real confusion over collisions happening with the Python Editor where the default 'declash' isn't there. Totally agree this needs work, though, as the 'magic radio group' is clever but also invisible... In some ways it just punts the problem down the line or makes it seem like 'radio' isn't working.

The biggest issue with a broadcast group is that people are often building their own communications protocol (for example, '1' means 'heart', '2' means dog) and if you get input you're not expecting it's really hard to separate logic errors from miscellaneous radio activity. I've seen a classroom where people send around the string "STOP USING RADIO".

I wonder though, if *just* moving the 'radio set group' up might do the trick? If you make that more visible it'll help people get to it when they need to be in the same group.

@microbit-giles and @finneyj might be able to help work out how/where to test.

pelikhan:
I am in favor of putting “set group” as the first block in the radio category —and— add “set group” as a first step in all tutorials/projects.

Choosing the channel is logically the first thing you do when you use any kind of talkie-walker. We are no different here.

PR #2042, opened 2019-05-07
Update extensions.md

kaku111:
@microbit-mark 
Hello!!
Because this extension will be used by more than two companies at the same time, it cannot add the company name. So we want to be able to delete the company name, leaving only the name of the robot in the Extension Gallery website. thanks!!


pelikhan:
It will get updated on the next bump.

kaku111:
@pelikhan 
Thanks pelikhan, but in the Extension Gallery the  name "prokit" seems to have not been removed. What should I do for this?

The website of the Extension Gallery is below:
       https://makecode.microbit.org/extensions
In the ROBOTICS project.  thanks a lot!!

image

PR #2067, opened 2019-05-22
add Keigan Motor extennsion

microbit-mark:
cc @keigan-motor


keigan-motor:
Thank you so much all.
I have fixed the issues.
Please tell me if we should do more action.

Issue #2085, opened 2019-05-29
Turtle Spiral example blocks not loading properly 


jwunderl:
**Describe the bug**

Turtle Spiral isn't decompiling properly.

Also minor grammar error: This project uses **an** microturtle Extension

**To Reproduce**
1. Go to turtle spiral example
2. see error

**Expected behavior**

decompiles properly

**Screenshots**

![Screen Shot 2019-05-29 at 11 30 40 AM](https://user-images.githubusercontent.com/5615930/58582183-08c42e00-8206-11e9-8803-714c2a34a6a4.png)

abchatra:
Good to fix, but I am assuming this is not a regression from live. Right?

jwunderl:
Correct, not a regression

PR #2094, opened 2019-05-29
Fix Turtle Spiral example decompilation

jwunderl:
fixes https://github.com/microsoft/pxt-microbit/issues/2085

I'm assuming it's intentional that the following decompiles to gray blocks, as it's not necessarily safe to just turn them into regular loops

```typescript
let index = 0;
for (index = 0; index <= 4; index++) { }
for (index = 0; index <= 4; index++) { }
```


riknoll:
Yep, this is working as intended

PR #2115, opened 2019-06-01
Fix webusb in microbit

darzu:
See: https://github.com/microsoft/pxt/pull/5517


darzu:
NVM, I'm updating https://github.com/microsoft/pxt/pull/5626 to not need target changes

Issue #2025, opened 2019-04-25
serial baud rate 300 supported?


microbit-mark:
microbit support: 21787

**Describe the bug**
 
Using https://makecode.microbit.org/_gvXFiKAim9Th and timing bit interval measures ~840us on an oscilloscope,  Working this figure backwards means that the operational baud rate is close to 1200 bps.
![DSCN1817](https://user-images.githubusercontent.com/31242877/56730269-6c62c380-674f-11e9-9931-b65b004f4f80.JPG)

If one has access to a TTL serial interface and Putty or equivalent, setting the receive Baud rate to 300 results in gibberish being displayed, changing Putty speed to 1200, results in clear text as transmitted being displayed.

**Expected behavior**
The bit width at 300 Baud should be 3.333’ms.

**Additional context**
This shows how the baud rate is configured from MBED
https://github.com/lancaster-university/mbed-classic/blob/master/targets/hal/TARGET_NORDIC/TARGET_MCU_NRF51822/serial_api.c#L108-L113

and this shows the baud request is sent to the Mbed RawSerial object directly:
https://github.com/lancaster-university/microbit-dal/blob/9164922a519aa213a565cbf2abaa2cf4debf4977/source/drivers/MicroBitSerial.cpp#L819

so it should obey rates listed in:
https://github.com/lancaster-university/mbed-classic/blob/master/targets/hal/TARGET_NORDIC/TARGET_MCU_NRF51822/serial_api.c#L32-L50

As the lowest available rate is 1200, Does this mean that having the option MakeCode to set the baud rate to 300 should not be available, or is there another reason that it is?

abchatra:
fixed

PR #2124, opened 2019-06-04
remove baud rate as it is not support on HW

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/2025.

This is a breaking change however, it never worked on HW so i think it does not make sense to try do a workaround aside from removing the entry.


pelikhan:
@abchatra ping

Issue #1762, opened 2018-12-27
Expose message bus listener flags in control.onEvent 


philipphenkel:
**Is your feature request related to a problem? Please describe.**
Currently, control.onEvent uses MESSAGE_BUS_LISTENER_QUEUE_IF_BUSY by default. Up to 10(?) events are queued. While this is a good default behavior I would like to have more control.

I missed configuration options like DROP_IF_BUSY or IMMEDIATE already several times when working on projects with high time sensitivity. This time I'm working on IR signal detection. I'm waiting for the first high pulse and read subsequent signals directly via digitalReadPin. Furthermore, it would reduce memory pressure if we don't queue events always. Therefore both DROP_IF_BUSY or IMMEDIATE would be useful.

**Describe the solution you'd like**
Please add a third parameter to control.onEvent so that the callback behavior can be specified like in the CPP layers.

- MESSAGE_BUS_LISTENER_REENTRANT
- MESSAGE_BUS_LISTENER_QUEUE_IF_BUSY
- MESSAGE_BUS_LISTENER_DROP_IF_BUSY
- MESSAGE_BUS_LISTENER_IMMEDIATE

**Describe alternatives you've considered**
"Drop if busy" can be realized on caller side (this is not efficient). I cannot think of a replacement of "immediate" except polling.

**Additional context**
See control.cpp control.onEvent. It would be a small change. We would need to add a parameter and pass it as the 4th argument to registerWithDal.
```
  /**
     * Registers an event handler.
     */
    //% weight=20 blockGap=8 blockId="control_on_event" block="on event|from %src=control_event_source_id|with value %value=control_event_value_id"
    //% help=control/on-event
    //% blockExternalInputs=1
    void onEvent(int src, int value, Action handler) {
        registerWithDal(src, value, handler);
    }
```

abchatra:
Fixed

PR #2126, opened 2019-06-04
support flag argument

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/1762


pelikhan:
IMMEDIATE is not compatible with our runtime (at least currently). Leads to 914 error (PANIC_CALLED_FROM_ISR)

abchatra:
Lets take this. Block works with new default argument and backward compatible right?

pelikhan:
Tested on hardware.

pelikhan:
Yes old blocks load without issues.

Issue #1886, opened 2019-02-28
How to extend the package from the offline app without internet connection?


lionyhw:
How to extendthe package from the offline app without internet connection?

pelikhan:
@lionyhw could you give more details about your request?

lionyhw:
Now I have downloaded the offline app and installed it, and then I cut off my internet, 
![](https://i.imgur.com/57GY4zm.png)
 I want to extend the package into the project, but it shows I failed to connect the internet and appears to be offline. 
![](https://i.imgur.com/EFYNmxB.png)
I found the "packages" in the files of offline app, will it do any favor to extend the package?
![](https://i.imgur.com/01R1bAh.png)
How to extend the package into the project without the internet connection?

lionyhw:
@pelikhan Looking forward to your reply!

pelikhan:
This is not supported yet. Could you explain further your scenario?

lionyhw:
The students in the classroom have no access to the internet, how can we do to help the students to extend the package in this condition?

pelikhan:
What packages are you missing?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

lionyhw:
Look，I want to extend the package "Tinkercademy" and "Ring:bit car " , but the only way to achieve is to download through the internet connection.   
 ![](https://i.imgur.com/zrh6ocS.png)
But how can I do to extend these two packages to the offline App without the Internet connection? 
 ![](https://i.imgur.com/OL7A90b.png)

abchatra:
The request the way I understand it. If they had the local github repo of the extension, can they somehow load it as an extension.

lionyhw:
Yes. You understanding is correct. 
One computer is off-line. I copy the "makecode for micro:bit off-line app" to this computer by USB drive,but I can't extend any packages on this app when my computer is offline. So I want to try load local github repo. Will it works? If it don't work, is there any solutions?

pelikhan:
In the next version of makecode, you will be able to import a .hex as an extension. So save your package as a .hex; then share it with your offline users.

PR #2127, opened 2019-06-04
import extension files

pelikhan:
image

Fix for https://github.com/microsoft/pxt-microbit/issues/1886
@riknoll Make sure this works in Electron.


abchatra:
Changes in the commit doesn't look relevant? Why is 
    getUint8(off: int32): int32;?

pelikhan:
They appear for me if you run “pxt buildtarget” but that might be a problem with my local setup

pelikhan:
Will fix

Issue #2104, opened 2019-05-30
set_pull_pin has no impact on the simulator


vijairaj:
**Describe the bug**
`set_pull_pin` `P1` to `up` doesn't seem to have any impact on the simulator, so the `digital_read_pin` reads out `0` instead of `1` on restart.

**To Reproduce**
Please check the [snippet](https://makecode.microbit.org/91397-74083-35314-29491)

abchatra:
Fixed

PR #2133, opened 2019-06-05
Handle setPull in simulator and set pin value

vijairaj:
Fix for #2104

### DIFF
--- a/sim/state/edgeconnector.ts
+++ b/sim/state/edgeconnector.ts
@@ -48,7 +48,7 @@ namespace pxsim.pins {
         let pin = getPin(pinId);
         if (!pin) return;
         pin.pull = pull;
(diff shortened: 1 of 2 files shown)


sim/state/edgeconnector.ts
```diff
@@ -48,6 +48,7 @@ namespace pxsim.pins {
         let pin = getPin(pinId);
         if (!pin) return;
         pin.pull = pull;
+        pin.value = pull > 0 ? 1023 : 0;
```
pelikhan: you should check pull against the microbit-dal value definitions.

vijairaj: Updated PR. Thank you.

pelikhan:
Thank you for your contribution!

Issue #2041, opened 2019-05-05
Typo in enum value BackAndWhite


editicalu:
The micro:bit block for the LED mode "Black and white" uses the enum value `BackAndWhite`. However, the visual editor shows the 'black and white' option.

## Visual 
![Black and white](https://user-images.githubusercontent.com/22838908/57192249-a3468f80-6f2e-11e9-9d8a-c0f3ac5f5b51.png)

## JavaScript

```ts
led.setDisplayMode(DisplayMode.BackAndWhite)
```

## Reference

The reference talks about `BlackAndWhite`, but the user gets an error when he tries to add this.

![The error the user recieves](https://user-images.githubusercontent.com/22838908/57192309-3b447900-6f2f-11e9-90f8-fa13074c6dcb.png)

Helen09:
This issue doesn't repro as I tried
![Issue51301](https://user-images.githubusercontent.com/23466737/57593399-d19b1f00-756d-11e9-9902-aa971fdc873b.gif)

editicalu:
It clearly states `BackAndWhite` (without an L in bLack) in your screencast. The error occurs when entering `BlackAndWhite` manually.

PR #2138, opened 2019-06-05
Fix Black and White Typo

ChaseMor:
Switches the `DisplayMode.BackAndWhite` to the correct `DisplayMode.BlackAndWhite`

Fixes https://github.com/microsoft/pxt-microbit/issues/2041


ChaseMor:
@riknoll @pelikhan We had talked about just adding an additional corrected spelling to the enum, but I can't seem to get it to hide the previous option. Not including `//%` just defaults to upgrading the name from the camelCase and `//% blockHidden=true` doesn't seem to work.

Issue #2024, opened 2019-04-25
SerialWriteLine() – sends out text, padded with spaces at the end in multiples of 32


microbit-mark:
microbit support:  21786 

**Describe the bug**
"Padding is added at the end of text by additional spaces (ASCII 32) when using serialWriteLine() to make up difference to a length in multiples of 32, less \r\n , irrespective of  TxBufferSize() for variables and literals.

One can set the Tx buffer size away from default, either lower or higher, it makes no difference to the number characters output.

**To Reproduce**
Makecode Beta was used with serial re-directed to P0, P1 at 9600 Baud, with serial TTL to USB converter (3.3V) connected to computer running Putty.

The following is from a Putty log, A short string of 5 characters gets padded with 25 trailing spaces, followed by \r\n making a total of 32 characters output. Note, I added the “<” for clarity of spaces.
The hex dump is of the same string, clearly showing ASCII character “20” is output.

TestA                         <
54657374412020202020202020202020
20202020202020202020202020200D0A
"

** This also reporoduces over USB**
1. Go to 'https://makecode.microbit.org/_Y7MYCxUMH3Wx '
2. DOwnload to micro:bit and view serial output as HEX. I use Coolterm for this
3. Press A, then B, see on B serial write line adds a bunch of '20's.

**Expected behavior**
new line is created and character is printed without extraneous spaces

**Additional context**
This functionality appears to be deliberate 'padding' see:
https://github.com/Microsoft/pxt-microbit/blob/bc4b4445b0050433bbf0ac62c5436bbee1db4165/libs/core/serial.ts#L22 
What is the reason for this @abchatra ?

abchatra:
Winrt layer (Windows app) fails to parse the output otherwise. Done on purpose by this change:
https://github.com/Microsoft/pxt-microbit/pull/676

abchatra:
@pelikhan

whaleygeek:
This means it is impossible to use the UART to communicate with external modules connected to pins, because the link is no longer binary transparent, so this is a breaking change.

Do we need padding to be an optional feature that we can turn off? Or fix the bug in the right place in the Winrt layer instead?

whaleygeek:
I think I would tolerate padding on by default when pins routed to USB, padding off by default when routed to pins and a javascript only method call to enable/disable padding for special cases // @jaustin 

But making the UART non binary transparent makes it impossible now to communicate with modules attached to pins that require full binary transparency.

[...]

PR #2145, opened 2019-06-07
add API to disable serial padding.

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/2024
image


abchatra:
I don't like this api text.

How about
```serial set writeLine pad to 0```
```setWriteLinePad```

It only impacts writeLine and nothing else.

pelikhan:
@abchatra ready for review.

Issue #2147, opened 2019-06-07
Shake button on simulator doesn't work more than once


abchatra:
**To Reproduce**
1. Add a 'on shake' block
2. Click on simulator on shake button
3. Works only on first time.

**Desktop (please complete the following information):**
 - OS: iPad
 - Browser: Safari

abchatra:
Is this related to https://github.com/microsoft/pxt-microbit/issues/820?

Issue #820, opened 2018-05-30
On Shake can only be triggered once with the mouse in the simulator


Helen09:
**Repro steps:**
• Navigating to https://makecode.microbit.org/beta#editor
• Drag blocks "on shake", "show number" and "pick random 0 to 4"
• Shake the simulator and observe
**Expected Result:**
Shaking simulator should be same as clicking "SHAKE" button, block "pick ramdom" always works
![2](https://user-images.githubusercontent.com/23466737/34331731-5b4ea59c-e964-11e7-9575-6737c3b8c1a9.gif)
**Actual Result:**
When shaking the simulator, after block "pick random" works once and the number doesn't change anymore
![1](https://user-images.githubusercontent.com/23466737/34331737-622ced10-e964-11e7-9505-dc872304e765.gif)

nbogie:
Experimenting here, It seems to be possible to trigger a shake event ONCE in the simulator this way but then not again until the "shake" button is pressed, after which time a shake can again be triggered ONCE (sometimes), and so on.

So this is not to do with the "pick random" block, rather the unintuitive way shake events are triggered, and not triggered, by manual movement of the simulated microbit.  

From my perspective, it would be better if such manual movement of the simulated microbit either NEVER triggered a shake event (just use the button instead), or ALWAYS triggered a shake event given the accelerometer data actually matched the gesture.

(experimented on chrome 67.0.3396.62 on mac os 10.13.4)

Helen09:
This issue still reproes

abchatra:
Fixed

PR #2161, opened 2019-06-07
always shake when button is pressed

jwunderl:
fixes https://github.com/microsoft/pxt-microbit/issues/2147
re https://github.com/microsoft/pxt-microbit/issues/820

I had actually fixed this in https://github.com/microsoft/pxt-microbit/pull/1983 but changed it to retain the behavior as it seemed like an intentional change (to match the behavior when you try to continuously shake it otherwise).

this makes it so when you press shake it always shakes, instead of waiting for you to start a different gesture like it normally would.

### DIFF
--- a/sim/state/accelerometer.ts
+++ b/sim/state/accelerometer.ts
@@ -267,7 +267,7 @@ namespace pxsim {
                 ++this.sigma;
             }
 
-            if (this.sigma >= DAL.MICROBIT_ACCELEROMETER_GESTURE_DAMPING) {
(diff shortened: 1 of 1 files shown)


jwunderl:
Actually, this doesn't quite fix 820 after re-reading it, as it's relating to just shaking the simulator which should eventually retrigger the event like shaking an actual microbit does (rather than waiting for a separate gesture to occur first). I'll try fix the logic here to fix that too, shouldn't be too bad

Issue #2162, opened 2019-06-07
Stop melody all does not stop melodies playing in the background.


kimprice:
**Describe the bug**
Stop melody all does not stop melodies that are playing in the background.

**To Reproduce**
See picture.

**Expected behavior**
All melodies should stop playing.

**Screenshots**
![stopMelodyAll](https://user-images.githubusercontent.com/15070078/59136173-c3ef7400-8936-11e9-96a7-4695bbecaf5e.png)

abchatra:
@pelikhan Expected?

pelikhan:
This is the behavior of the new music library. We need to keep a list of melodies playing in order to stop them all.

________________________________

pelikhan:
This might be a regression; a feature interaction with the microbit background sounds and the new sound engine.

PR #2174, opened 2019-06-10
stop background before foreground

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/2162


pelikhan:
The foreground melody has priority so it's best to first cancel foreground, then background. There's probably more dragons here but I tested various combinations and everything looks good.

Issue #2215, opened 2019-06-17
OLED12864_I2C extension not work in MakeCode V2


shaoziyang:
In MakeCode V2, when add OLED12864_I2C extension and download to microbit, microbit will show error code 022.

Same program is work in MakeCode V1.

pelikhan:
Use the hex literal to encode the font into a buffer storred in flash.

   const font = hex `02abc....`;

________________________________

shaoziyang:
I tried this method, but it didn't work. It seems that extension have use a big buffer (1025 bytes) cause memory error.

pelikhan:
Did you use a single hex buffer?

shaoziyang:
Yes, I use single hex buffer for **font**, but get same error. When I reduce buffer size for **_screen** (it will loss of data), it will work.

https://github.com/makecode-extensions/OLED12864_I2C

pelikhan:
@mmoskal

mmoskal:
I'm looking into fixing the buffer size issue.

However, you should still use a hex buffer for font - this will save 0.5k of memory for other uses. Let me know if you need more info on how to use the hex buffer for font.

shaoziyang:
I test it in pxt version 2.1.1, it still doesn't work and show error code 022.

pelikhan:
The fix is not live yet.

________________________________

darzu:
I've confirmed this appears fixed in /beta https://makecode.microbit.org/_FsxCaU8tKbK8

@shaoziyang , if you have the time, could you try your project as well to make sure the issue is gone in /beta?

shaoziyang:
@darzu Yes, it works now, I will do more test soon.

HyperDuino:
I just tried but still get 022 error.  https://makecode.microbit.org/_hvLA8yV1PfAY

shaoziyang:
@HyperDuino You need run it in MakeCode beta now. 

https://makecode.microbit.org/beta

I am working in new version of OLED12864, with 6x8 font, and quick display speed.

abchatra:
Fixed

PR #2216, opened 2019-06-17
Use gcPreAllocateBlock() to fix #2177, #2215

### LINKED ISSUE #2177: Panic 022 on beta (only) with this program
**Describe the bug**
This program causes panic 022 on beta after 176 iterations but continues for over 1000 on live.
https://makecode.microbit.org/_bWcdFXavrdqc

**To Reproduce**
1. Go to 'https://makecode.microbit.org/_bWcdFXavrdqc'
2. Click on 'Edit'
3. Download to micro:bit
4. Click Show Console Device
5. Let the count go above 176
5. Change to beta, Download and Show Console 
6. Wait for panic

**Expected behavior**
Behaviour should be similar between live and beta.

 - OS: Windows
 - Browser Chrome
makecode.microbit.org version:  1.4.16
Microsoft MakeCode version:  5.15.6
microbit runtime version:  pxtgc-0

mmoskal:
Requires https://github.com/microsoft/pxt-common-packages/pull/866

This allocates additional 8k of memory in one segment on non-BLE builds

Needs testing with BLE enabled.

fixes #2177, #2215

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@types/web-bluetooth": "0.0.4"
   },
   "dependencies": {
-    "pxt-common-packages": "6.9.3",
+    "pxt-common-packages": "6.12.6",
     "pxt-core": "5.15.10"
   }
 }
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@types/web-bluetooth": "0.0.4"
   },
   "dependencies": {
-    "pxt-common-packages": "6.12.6",
+    "pxt-common-packages": "6.9.4",
     "pxt-core": "5.15.10"
   }
 }
--- a/libs/core/codal.cpp
+++ b/libs/core/codal.cpp
@@ -18,14 +18,18 @@ extern "C" void target_reset() {
     microbit_reset();
 }
 
+uint32_t device_heap_size(uint8_t heap_index); // defined in microbit-dal
+
 namespace pxt {
 
 MicroBit uBit;
 MicroBitEvent lastEvent;
 
 void platform_init() {
-    microbit_seed_random();
+    microbit_seed_random();    
     seedRandom(microbit_random(0x7fffffff));
+    if (device_heap_size(1))
+        gcPreAllocateBlock(device_heap_size(1) - 4);
 }
 
 void platform_init();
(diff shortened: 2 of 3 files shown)


mmoskal:
the pxt-common-pkgs PR is already on stable6.9, so just needs a bump there and dep update here or on relevant stable branch

darzu:
> the pxt-common-pkgs PR is already on stable6.9, so just needs a bump there and dep update here or on relevant stable branch

Awesome, thanks, I can drive the change into the live release (after testing of course).

It does look like this PR has errors?
```
/home/build/prj2/source/core/codal.cpp
/home/build/prj2/source/core/codal.cpp: In function 'void pxt::platform_init()':
/home/build/prj2/source/core/codal.cpp:32:51: error: 'gcPreAllocateBlock' was not declared in this scope
         gcPreAllocateBlock(device_heap_size(1) - 4);
                                                   ^
/home/build/prj2/source/core/codal.cpp: In function 'void pxt::gcProcessStacks(int)':
/home/build/prj2/source/core/codal.cpp:255:9: warning: unused variable 'cnt' [-Wunused-variable]
     int cnt = 0;
         ^
ninja: build stopped: subcommand failed.
```

mmoskal:
this is because p-c-p is not bumped

package.json
```diff
@@ -45,7 +45,7 @@
     "@types/web-bluetooth": "0.0.4"
   },
   "dependencies": {
-    "pxt-common-packages": "6.9.3",
+    "pxt-common-packages": "6.12.6",
```
mmoskal: I believe you need to stay on 6.9 for microbit; I don't think we'll change that until next year

darzu: Oh I see, thanks

darzu: Oh so the other PR went straight into stable6.9, shouldn't it have gone into master then cherry-picked into stable6.9? I'm still figuring this out..

I'll create a PR to cherry-pick it into pcp master

mmoskal: it should have generally went into master first, and be cherry picked to stable; however I think it needs changes before it can go to master, since things have changed in that region of code in dev - which is why I did on stable6.9 as it's kind of microbit specific; I'll handle the pcp PR sometime later (we don't need it until next major microbit release)

darzu: https://github.com/microsoft/pxt-common-packages/pull/884

PR #2242, opened 2019-07-02
Bump stable2.0 to include: Use gcPreAllocateBlock() to fix #2177, #2215 from #2216

### LINKED ISSUE #2177: Panic 022 on beta (only) with this program

darzu:
Cherry-picks #2216 

Bumps microbit stable2.0 index-ref to 2.0.4

I checked that this fix in /beta appears to fix: #2177 and #2215

StillTODO is to check that this fix works off the stable branch (which doesn't have everything beta has)


docs/index-ref.json
```diff
@@ -1,3 +1,3 @@
 {
-    "appref": "v2.0.3"
-}
+    "appref": "v2.0.4"
```
mmoskal: I think this should go to beta first and the main app should be only updated in a later PR

darzu: Yes, I've pushed it to beta with 2.1.2. I'm going to try to test the neopixel fix and stability there. I'll still want to test off the release branch I think because it doesn't have all the same commits as beta

mmoskal:
BTW you can view all

darzu:
> BTW you can view all versions at /v1.2.3 etc don't have to be tagged as beta

Oh thanks, I forgot about that. I was able to test /v2.0.4, the RC

docs/index-ref.json
pelikhan: where is 2.1.2 built from?

Issue #2245, opened 2019-07-02
Panic 020 with basic radio usage after gcPreAllocBlock() fix


darzu:
**Describe the bug**
A panic ":( 020" occurs with this program: https://makecode.microbit.org/_JjAcytiDu5zC

**To Reproduce**
1. Open: https://makecode.microbit.org/_JjAcytiDu5zC in makecode.microbit.org/v2.0.4
2. Download to hardware device
3. Shake device, wait, press buttons, shake again, press buttons
4. Should see ":( 020" error scroll across

**Other repros:**
- https://makecode.microbit.org/_UMChbu6cYVuV

Seen on Edge on Windows and Chrome on Mac.

This only happens after this change:
https://github.com/microsoft/pxt-common-packages/pull/866
https://github.com/microsoft/pxt-microbit/pull/2216

FYI: @mmoskal 

This is a ship-blocker for the fix to https://github.com/microsoft/pxt-microbit/issues/2215 and https://github.com/microsoft/pxt-microbit/issues/2177

mmoskal:
In non-microbit targets most of codal allocations happen in our GC heap, except for ones made in interrupt service routines. In microbit DAL allocates everything in its own heap so we have to decide on a split between DAL and GC. Long-term we should use the strategy from other targets in microbit as well.

pelikhan:
Tested as fixed in v2.0.5

PR #2246, opened 2019-07-03
Decrease size of GC heap to allow more DAL allocs

mmoskal:
This PR:
 * decreases the size of initial GC heap to 256 from 1024
 * leaves 1024 bytes from the second chunk of heap for microbit-dal allocations (in particular fibers and their stacks)

fixes https://github.com/microsoft/pxt-microbit/issues/2245


pelikhan:
@darzu are pushing this fix to beta?

pelikhan:
Tested on HW as good. Pushing to beta.

pelikhan:
Deployed to /beta and cherry-picked into master.

PR #2249, opened 2019-07-04
add 4tronix minibit

microbit-mark:
cc @4tronix


4tronix:
Can we have an update on this please? Product expected to be released very shortly

pelikhan:
I have opened two issues.

PR #2266, opened 2019-07-16
don't calibrate when reading field strength

pelikhan:
When measuring the magnetic field strength, don't calibrate.
- [ ] test on hardware


pelikhan:
@jamesadevine

PR #2314, opened 2019-07-22
Fix remove life animation causing microbit stuck

Leo1003:
On the real microbit board, if the program execute other game blocks while the remove life animation is playing, it would cause strange behavior or even make the game stuck. 

This pull request make `removeLife()` to lock the `_backgroundAnimation`, thus, it should behave correctly as `addScore()`.


pelikhan:
Thanks!

Issue #2247, opened 2019-07-03
Update https://makecode.microbit.org/device/error-codes 


microbit-mark:
https://makecode.microbit.org/device/error-codes The list is outdated.

_Originally posted by @pelikhan in https://github.com/microsoft/pxt-microbit/issues/2225#issuecomment-505046012_

We're in the process of updating https://microbit.org/guide/hardware/error-codes/ to include new information eg garbage collector errors. Are there any errors here https://makecode.com/js/errorcodes that ought to be included here https://makecode.microbit.org/device/error-codes

ganicke:
@microbit-mark - we have an error codes page for the CODAL erros at https://github.com/microsoft/pxt-common-packages/blob/master/libs/base/docs/device/error-codes.md

Some/most of the codes are common probably but I can take the current defines from the Lancaster repo and update the page at https://makecode.microbit.org/device/error-codes.

ganicke:
Update with GC error codes. #2327

Issue #2225, opened 2019-06-24
Error code 848


microbit-pauline:
**Describe the bug**
A user has contacted micro:bit through support as they have an error code and cannot locate information on the error.

Can you provide some information?

Error code -  sad face 848

pelikhan:
https://makecode.microbit.org/device/error-codes The list is outdated.

martinwork:
https://makecode.com/js/errorcodes

> The error codes in the 800-899 are even more internal:
> 
> 84X - garbage collector errors

martinwork:
Where does this panic occur in the source code?

microbit-pauline:
I asked the user the question and the response was "When trying to display an image on the scrollbit".
I'm not sure if that is helpful.

martinwork:
Sorry @microbit-pauline , if that was my question. I was hoping @pelikhan  or @mmoskal  might point me towards where panic 848 is generated in the MakeCode source code.

mmoskal:
The GC errors 84x are not something users can diagnose. If you really want, you can search for oops(48) or check macro with 48 argument in gc.cpp

To diagnose it we need the user program

martinwork:
Thanks @mmoskal . Found it in
https://github.com/microsoft/pxt-common-packages/blob/master/libs/base/gc.cpp

@microbit-pauline Can the user supply the program that causes the error?

microbit-pauline:
@martinwork I've made the request

ThScho:
I get error code 848 when displaying "Hello World" on a scroll:bit (https://shop.pimoroni.de/products/scroll-bit). micro:bit crashes (?) after displaying "Hel". Sometimes error code 848 is shown, sometimes not.

ThScho:
After reinstalling the scroll;bit software onto the device the error didn't occur any longer.

PR #2327, opened 2019-08-01
Error codes page update

ganicke:
Update the [Error Codes](https://makecode.microbit.org/device/error-codes) page.

* What additional error codes beyond the GC errors should we add? A broader list is in https://github.com/microsoft/pxt-common-packages/blob/master/libs/base/docs/device/error-codes.md
* Seems that the GC errors have multiple meanings, like **49** for example - ``thread context not found`` in one instance and ``null pre-alloc pointer`` in another.

Re: #2225 and #2247

### DIFF
--- a/docs/device/error-codes.md
+++ b/docs/device/error-codes.md
@@ -40,7 +40,7 @@ Memory error codes range from **800** - **909**.
 
 #### Report errors!
 
-If you ever see an error within the range of **800** - **899**, please report an issue at [GitHub](https://github.com/microsoft/pxt-microbit/issues) or on the [support](https://support.microbit.org/) page.
+If you ever see an error within the range of **800** - **909**, please report an issue at [GitHub](https://github.com/microsoft/pxt-microbit/issues) or on the [support](https://support.microbit.org/) page.
 
 ### ~
 
--- a/docs/device/error-codes.md
+++ b/docs/device/error-codes.md
@@ -34,7 +34,7 @@ The @boardname@ system errors range between **01** - **99**. For a full list of 
 
 ## Memery errors
 
-Memory error codes range from **800** - **899**.
+Memory error codes range from **800** - **909**.
 
 ### ~alert
 
--- a/docs/device/error-codes.md
+++ b/docs/device/error-codes.md
@@ -1,39 +1,86 @@
 # Error codes
 
-Your @boardname@ may encounter a situation that prevents it from running your code. When this happens, a frowny face will appear on your @boardname@ screen (see picture) followed by an error number. These are called _panic_ codes. For a full list of  these codes, what they mean and what you can do to resolve them visit the [micro:bit guide to error codes](https://microbit.org/guide/hardware/error-codes/).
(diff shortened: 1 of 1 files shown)


docs/device/error-codes.md
```diff
@@ -13,6 +14,16 @@ Your @boardname@ may encounter a situation that prevents it from running your co
 * **51** (`MICROBIT_HARDWARE_UNAVAILABLE_MAG`): an error occurred with the micro:bit's magnetometer component.
 * **98** (`MICROBIT_ASSERTION_FAILED`): assertion failed, the condition in an [assert](/reference/control/assert) was false.
 
+## Garbage collector errors
+
```
mmoskal: please add something along the lines of: If you ever see any of these, please report an issue on github or through your support channel. In fact this applies to all errors between 800 and 899.

mmoskal: in fact microbit can also get most of these: https://makecode.com/js/errorcodes
in particular the invalid cast errors

ganicke: Ah, thanks.

docs/device/error-codes.md
```diff
@@ -1,28 +1,86 @@
...
-```sim
-basic.showLeds(`
-    # . . . #
-    # # . # #
-    . . . . .
-    . # # # .
-    # . . . #
-    `)
-```
+## Memery errors
```
ganicke: This is more like"Object Management" errors?

mmoskal: these are generally internal errors; I'm not sure we need to explain the values to the users...

docs/device/error-codes.md
```diff
@@ -1,28 +1,86 @@
...
+
+Memory error codes range from **800** - **899**.
```
mmoskal:
```suggestion
Memory error codes range from **800** - **909**.
```

it looks like that 900-909 are also quite internal, there is no user action to take

mmoskal: though I guess it's good to list them, so they are googlable

docs/device/error-codes.md
```diff
@@ -1,28 +1,86 @@
...
-```
+## Memery errors
+
+Memory error codes range from **800** - **909**.
+
+### ~alert
+
+#### Report errors!
+
+If you ever see an error within the range of **800** - **899**, please report an issue at [GitHub](https://github.com/microsoft/pxt-microbit/issues) or on the [support](https://support.microbit.org/) page.
```
mmoskal:
```suggestion
If you ever see an error within the range of **800** - **909**, please report an issue at [GitHub](https://github.com/microsoft/pxt-microbit/issues) or on the [support](https://support.microbit.org/) page.
```

docs/device/error-codes.md
microbit-mark:
```suggestion
## Memory errors
```

microbit-mark: The more information the better, i think. It's certainly a useful reference for me and more users are reporting garbage collector errors and cast errors

PR #2390, opened 2019-08-13
add OLED12864_I2C_V2

microbit-mark:
cc @shaoziyang


pelikhan:
How is it different from https://github.com//makecode-extensions/OLED12864_I2C ?

pelikhan:
I am going to close this. I think we should update the "V1" rather duplicate it.

Issue #2292, opened 2019-07-18
WCAG 1.4.3: Elements must have sufficient color contrast (.text)


jwunderl:
#### Issue
`Elements must have sufficient color contrast` ([color-contrast](https://dequeuniversity.com/rules/axe/3.2/color-contrast?application=msftAI))
#### Target application
[Microsoft MakeCode for micro:bit](https://makecode.microbit.org/#editor)
#### Element path
.approve.red.labeled > .text
#### Snippet
`<span class="ui text">Reset</span>`
#### How to fix
    Fix any of the following:
      Element has insufficient color contrast of 2.78 (foreground color: #ffffff, background color: #ef767a, font size: 12.0pt, font weight: bold). Expected contrast ratio of 4.5:1
#### Environment
Chrome version 75.0.3770.142
This accessibility issue was found using Accessibility Insights for Web 2.7.1 (axe-core 3.2.2), a tool that helps find and fix accessibility issues. Get more information & download this tool at [http://aka.ms/AccessibilityInsights](http://aka.ms/AccessibilityInsights).

reset button in reset dialogue

![Screen Shot 2019-07-18 at 2 38 22 PM](https://user-images.githubusercontent.com/5615930/61493946-b7b0fa80-a969-11e9-90e5-9ea62d00e4e1.png)

jwunderl:
This will be fixed by https://github.com/microsoft/pxt-microbit/pull/2457, another commit that used to contain the change closed this a bit too early

BhuwanNaswa:
Labels
#A11yMAS, #HCL, #HCL-MakeCode-E2E, #MAS1.4.3, #Win10-Chrome, #Win10-Edge, #accessibility, #K4W, #closed

PR #2457, opened 2019-10-03
more accessible red buttons

jwunderl:
fix microsoft/pxt-microbit#2292; pulled out of https://github.com/microsoft/pxt/pull/5968 as the other targets didn't actually need to darken this button as much, just microbit

66072296-a360b080-e509-11e9-9b58-91cbda9ee44c

### DIFF
--- a/theme/site/elements/button.overrides
+++ b/theme/site/elements/button.overrides
@@ -1,3 +1,10 @@
 /*******************************
          Site Overrides
 *******************************/
+
+.ui.red.button {
+    background: darken(@red, 20%);
+    &:hover, &:focus {
+        background: darken(@red, 30%);
+    }
+}
--- a/theme/site/elements/button.overrides
+++ b/theme/site/elements/button.overrides
@@ -1,10 +1,3 @@
 /*******************************
          Site Overrides
 *******************************/
-
-.ui.red.button {
-    background: darken(@red, 20%);
-    &:hover, &:focus {
-        background: darken(@red, 30%);
-    }
-}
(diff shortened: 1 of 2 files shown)


theme/site/elements/button.overrides
```diff
@@ -1,3 +1,10 @@
 /*******************************
          Site Overrides
 *******************************/
+
+.ui.red.button {
```
pelikhan: should this fix go into pxt itself?

jwunderl: Issue is pretty specific to the red chosen for microbit (it's very pale); looking at it again there aren't that many places @red actually seems to be used, I'll go through and see if just changing that to make it slightly darker is a better idea

jwunderl:
Hm, so only other explicit use was in the toolbox when deleting, which would be a bit darker:

before:


this:


Do we particularly care about that / should I add a rule to lighten it back up a bit, or does it seem fine?

pelikhan:
still feel like this belong to pxt, not specific to microbit

jwunderl:
The issue is specific to how light the chosen sui red variable is for microbit; normal sui red / etc is perfectly fine. the current one is closer to salmon / a light pink:

Screen Shot 2019-10-17 at 4 08 24 PM


arcade
Screen Shot 2019-10-17 at 3 59 47 PM

adafruit
Screen Shot 2019-10-17 at 4 00 15 PM

lego
Screen Shot 2019-10-17 at 4 00 55 PM

all pass / are fairly dark,

only other problematic one is minecraft which doesn't quite pass (about .5 off, vs about 2 off for microbit)
Screen Shot 2019-10-17 at 4 01 42 PM

PR #2459, opened 2019-10-04
bump pxt 5.24.7

pelikhan:
- [ ] invesgitage xml generation difference: new blockly? non-pretty serialization?


libs/core/blocks-test/image.blocks
```diff
@@ -1,46 +1,43 @@
...
+    <block type="pxt-on-start" x="81" y="246">
+        <statement name="HANDLER">
+            <block type="device_show_image_offset">
+                <value name="sprite">
+                    <block type="builtin_image">
+                        <field name="i">IconNames.Heart</field>
+                    </block>
+                </value>
+                <value name="offset">
+                    <shadow xmlns="http://www.w3.org/1999/xhtml" type="math_number">
```
pelikhan: spurious xmlns namespace here. @shakao

libs/core/blocks-test/image.blocks
```diff
@@ -1,46 +1,43 @@
-<xml xmlns="http://www.w3.org/1999/xhtml">
-  <variables>
```
pelikhan: looks like blockly is not serializing those anymore

shakao: hmm, thanks for the heads up! investigating

PR #2460, opened 2019-10-04
Adding link to stable refs

abchatra:


abchatra:
Microbit /beta to latest pxt. We need a channel to test point fixes. Creating /stable for that purpose

PR #2462, opened 2019-10-05
Fix radio parsing race

pelikhan:
integration of https://github.com/microsoft/pxt-microbit/commit/36b8e5d661cefe73f9019c8bbf6aa013bf7a790a


abchatra:
How risky is this change? I am unable to decide.
Does this require testing of all the radio or is it safe? 
Should we wait for /beta feedback before merge?

pelikhan:
Let's put it this way: radio is broken right now and gives incorrect results if the packets come at a sufficient rate.

PR #2466, opened 2019-10-10
Inf 367 fix ci dir

wryun:


wryun:
I will try to stop doing this - sorry.

PR #2480, opened 2019-10-17
micro:bit RSSI fix

pelikhan:
The current RSSI implementation ignore the packet RSSI and uses the current RSSI which will be wrong if packets are queued. We need to read the RSSI value stored in the packetbuffer in CODAL.

- [x] the ``readRawBuffer`` appends the RSSI data into the returned buffer (+4bytes) and the PacketBuffer class was updated to account for that.
- [x] move receivedSignalStrength to TS

Test program. Press B for transmitter, A for receiver. Logs should show constant RSSI for both sources.
```
radio.setGroup(1)
radio.setTransmitSerialNumber(true)
console.log('started');

class Other {
    public rssi: number = 0;
    constructor(public index: number) { }
}

const others: any = {};
    // receiver
    radio.onReceivedNumber(r => handlePacket(r))
})

    // transmitter
    basic.forever(function () {
        led.toggle(0, 0)
        radio.sendNumber(1);
        basic.pause(500)
    })
})

function handlePacket(receivedNumber: number) {
    led.toggle(1, 0)
    const sn = radio.receivedPacket(RadioPacketProperty.SerialNumber);
    let other: Other = others[sn];
    if (!other)
        other = others[sn] = new Other(Object.keys(others).length);
    other.rssi = radio.receivedPacket(RadioPacketProperty.SignalStrength);
    console.log(`${other.index}: ${other.rssi}`)
    led.toggle(other.index, 1)
}
```


libs/radio/radio.cpp
```diff
@@ -40,11 +40,16 @@ namespace radio {
...
-        int size = uBit.radio.datagram.recv(buf, sizeof(buf));
-        if (size <= 0)
+
+        PacketBuffer p = uBit.radio.datagram.recv();
+        if (p == PacketBuffer::EmptyPacket)
             return mkBuffer(NULL, 0);
-        return mkBuffer(buf, size);
+
+        int rssi = p.getRSSI();
+        uint8_t buf[MICROBIT_RADIO_MAX_PACKET_SIZE + sizeof(int)]; // packet length + rssi
```
abchatra: There seems to be lot of copy. recv() function allocates packetBuffer p and copies to it and then it is copied to buf here. and the contents of buf are copied in mkBuffer constructor. We could avoid one copy

pelikhan: I'm not really looking into perf issues in this PR. This is about correctness of the RSSI value.

libs/radio/radio.cpp
```diff
@@ -31,30 +31,38 @@ namespace radio {
...
+     * Internal use only. Sends a raw packet through the radio (assumes RSSI appened to packet)
      */
-    //% advanced=true
     //% async
     void sendRawPacket(Buffer msg) {
         if (radioEnable() != MICROBIT_OK || NULL == msg) return;
-        uBit.radio.datagram.send(msg->data, msg->length);
+
+        // don't send RSSI data; and make sure no buffer underflow
+        uBit.radio.datagram.send(msg->data, max(0, msg->length - sizeof(int)));
```
mmoskal: Does max work on int or unsigned int?

PR #2519, opened 2019-11-13
micro:bit RSSI fix (#2480) [stable 2.0]

pelikhan:
Reads the RSSI stored in the packet rather than the current RSSI (last packet).


abchatra:
Can we wait till the next release? I don't see any customer bugs in this.

pelikhan:
The reported RSSI is (randomly) incorrect when multiple devices are connected.

abchatra:
Is this always been this way or we regressed in the last radio update?

pelikhan:
We regressed on the stack fix (https://github.com/microsoft/pxt-microbit/commit/36b8e5d661cefe73f9019c8bbf6aa013bf7a790a#diff-ff628e776d72b8d8ae97ded769251a16) by leaving the "packet" global around in C++.

The fix https://github.com/microsoft/pxt-microbit/commit/05bf2eec2ad5b597de5824025e200004a67f1b66#diff-24b616897aa6b7cc686169bd2bfae18dR84 is incorrect as it takes the current RSSI rather than the RSSI recorded in the packet.

This fix records the RSS from the packet https://github.com/microsoft/pxt-microbit/commit/f9e9daae183727783506eb15bfd059fc09477f15#diff-24b616897aa6b7cc686169bd2bfae18dR46 .

pelikhan:
I can write a "multi-potato" activity that tests this scenario. Basically we need 1 receiver, 2 transmitters at different distance and their RSSI should stay relavitly constant.

PR #2540, opened 2019-11-27
Docs: Change default value in Javascript.

CleoQc:
Per @microbit-mark 's suggestion, add an example of how to change the default interval value by switching to Javascript.

Also fix stated default value which is not 400 but 600.

Some context @pelikhan https://microbit-community.slack.com/archives/C226FGXQX/p1574799575099200


microbit-mark:
I think it would be a really useful idea to show examples of how to use parameters only available in JS to users if they are switching between blocks and text

PR #2546, opened 2019-12-03
Enable automatic image compression on PR

pelikhan:
Use GitHub action to optimize images on PRs _automatically_. 

This is how it works: the GitHub action runs image optimization libraries on the images in the repo, if any image gets optimized - it adds a commit to the PR with the optimized image.

This PR has a large number of optimized images since it is the first one we run the tools.


ganicke:
Does it make sense to reduce visual quality of images that are already quite small both in physical size and data size? Set some thresholds to determine when to apply optimization? Just a thought.

pelikhan:
Once the tool is in place it won’t modify those images anymore. But yes the first PR is heavy.

PR #2549, opened 2019-12-05
adding radio firefly in approved list

pelikhan:
Enable extensions that enable generic firefly behaviors.


abchatra:
This PR adds microsoft/pxt-apa102 as well. Is this intentional?

pelikhan:
Yes add both.

PR #2562, opened 2019-12-19
Switching to norwegian locale to nb 

abchatra:
Also removing unnecessary codes in targetconfig. These are not necessary in targetconfig.


abchatra:
needs signoff

abchatra:
Older editors take targetconfig from their branch (like v0  & v1) and  not master

PR #2572, opened 2020-01-09
Moving radio to common packages

pelikhan:
Uses the refactored radio apis from common packages.
Requires https://github.com/microsoft/pxt-common-packages/pull/1058

- [x] do not move docs to avoid disruption in translations
- [x] move radio implemention to common packages
- [x] move deprecated apis into "targetoverrides.ts"


abchatra:
Deleting documentation is ok here?  Wouldn't it cause live website to not have docs?

pelikhan:
The live web site will pull it from common packages. I will bring it back here for safety until we upgrade Microbit.

pelikhan:
@mmoskal ready for review.

PR #2579, opened 2020-01-14
Home page cards for 'Science Experiments' videos

ganicke:
Put in the first 4 videos of Shawn's 'Science Experiments' videos.

image

Note: There is a 'Science' category already. A small amount of name collision there.


pelikhan:
It would be nice to add them in each correspoding science lesson as well.

ganicke:
I just did that but yuz was too fast.

PR #2657, opened 2020-03-13
Extensions: Add Cytron edubit

microbit-mark:
@abchatra Do not merge until product listing is live https://www.cytron.io/p-edu-bit

cc @waiweng83 can you let us know on this Pull Request when the product listing is live on your website?

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -305,6 +305,10 @@ Check out [microbit.org](http://microbit.org/resellers/) for more information on
   "url":"/pkg/Seeed-Studio/pxt-grove",
   "cardType": "package"
 }, {
+  "name": "Cytron Edubit",
+  "url":"/pkg/CytronTechnologies/pxt-edubit",
+  "cardType": "package"  
+}, {
     "name": "WS2812B",
     "url": "/pkg/microsoft/pxt-ws2812b",
     "cardType": "package"
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -167,7 +167,8 @@
             "KitronikLtd/pxt-kitronik-clip-detector",
             "DFRobot/pxt-DFRobot-NaturalScience",
             "strawbees/pxt-robotic-inventions",
-            "daferdur/pxt-myHX711"
+            "daferdur/pxt-myHX711",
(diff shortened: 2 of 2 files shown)


abchatra:
@waiweng83 well designed simple set of blocks. I love it. 

Couple of comments: 
1. Can you just point to pxt-neopixel as the dependency instead of exact version? 
"NeoPixel": "github:microsoft/pxt-neopixel"
Is there a need for the exact version? 
2. Remove installedVersion from the pxt.json

>> do not merge until product listing is live https://www.cytron.io/p-edu-bit

Sure, but merging won't do anything other than surface it when someone searches for the extension with the name. As long as it is not advertised anywhere it is a no-op.

waiweng83:
> Can you just point to pxt-neopixel as the dependency instead of exact version?
> "NeoPixel": "github:microsoft/pxt-neopixel"
> Is there a need for the exact version?

We point it to exact version just to prevent future update of pxt-neopixel breaking our extension.
What's the better way of doing that?

abchatra:
>> We point it to exact version just to prevent future update of pxt-neopixel breaking our extension.
Is there a better way of doing that?

Up to you.  If there are bug fixes in neopixel library you won't receive them either.  We try to maintain backward compat. So unlikely future update will break your extension.

waiweng83:
> > > We point it to exact version just to prevent future update of pxt-neopixel breaking our extension.
> > > Is there a better way of doing that?
> 
> Up to you. If there are bug fixes in neopixel library you won't receive them either. We try to maintain backward compat. So unlikely future update will break your extension.

In that case, we prefer to point it to exact version. If there is future update on the neopixel extension, we will test it out and update the version manually.

abchatra:
Ok sounds good @pelikhan fyi

waiweng83:
Did I left out anything else to get it merged?

abchatra:
It is ready to be merged. Waiting for you to give green signal

waiweng83:
I hope that it can be merged now instead of waiting for the product to be launched.
So that we can prepare the handbook with actual user experience.

Thanks.

PR #2658, opened 2020-03-13
Update gesture field for new Blockly renderer

shakao:
Requires pxt-core bump once https://github.com/microsoft/pxt/pull/6697 is merged in


pelikhan:
Bump pxt version?

PR #2679, opened 2020-03-23
added love meter video

pelikhan:


ganicke:
Bad video ID, it's `1IYsy0_9n8g`....oh, you just fixed that.

PR #2685, opened 2020-03-24
Enable new features

abchatra:
- debugger
- new project name first 
- social in share

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -300,7 +300,6 @@
         "experiments": [
             "allowPackageExtensions",
             "instructions",
-            "debugger",
             "debugExtensionCode",
             "bluetoothUartConsole",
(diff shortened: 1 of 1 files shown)


pxtarget.json
```diff
@@ -398,6 +397,12 @@
         "coloredToolbox": true,
         "monacoToolbox": true,
         "hasAudio": true,
+        "socialOptions": {
+            "orgTwitterHandle": "MSMakeCode",
+            "hashtags": "MakeCode",
+            "discourse": "https://forum.makecode.com/",
```
pelikhan: are you sure you want to enable discourse for micro:bit?

abchatra: Yes this is share page and the context is sharing to social platform discourse and not for support.

PR #2686, opened 2020-03-25
Extensions: Add 4tronix Orbit

microbit-mark:
cc @4tronix


abchatra:
Looks good to me. @4tronix
One comment: 
- Do you need hard dependency on the version 0.1.4 of FireLedsBASE? 
If not just remove the version from there and say "github:4tronix/FireLedsBase"

4tronix:
Whenever I have done that in the past, it reverts to loading in a 
previous version. Any idea why this may be so?
I want to remove the version dependency if possible, but I need to have 
"this version or later"

Thanks

------ Original Message ------

pelikhan:
The editor always hardbounds to a specific version.

abchatra:
Editor should load the latest version and then when you share should hard bound to that version. Are you saying when you don't specify 0.1.4 it doesn't load that version (as that is the latest in this package)?

4tronix:
Yes.
I have removed the forcing of the version and bumped the release to 
0.1.3 and now it uses an old, un"released" version of the Fireleds

This is clearly not ideal behaviour. I would always want the latest 
release. Now I have to go through all projects that use this and update 
the version so that they get the bug-fixed release

See the two images attached

Best regards,
Gareth

------ Original Message ------

abchatra:
There is cloud caching which I think stays for 1 hour. So the results are not immediate. It seems to work right now.

4tronix:
Not for me it doesn't - 24 hours later. It is still loading the fireled 
base code v0.1.0 instead of v0.1.4
Are you sure yours is loading the correct fireled base code? - check in 
the github/4tronix/FireLedsBase/pxt.json - it should be "version 0.1.4", 
but the one loaded in the online Makecode editor is "version 0.1.0"

Gareth

------ Original Message ------

abchatra:
Shows the right dependency version for me.
image

Issue #2678, opened 2020-03-23
Blockly rendering bugs


shakao:
**Describe the bug**
Tracking known rendering bugs (P1):
- [x] Scrollbar on context menu
- [x] Right-click help on context menu shows alert
- [x] If-else shape is incorrect
- [x] (microbit) on gesture renders image instead of text
- [x] (arcade) Image field rendering (white instead of gray transparency)

P2 Bugs:
- [ ] Small text field overlap when piano editor is first opened
- [ ] Insertion marker does not consider child blocks (intentional with new renderer)
- [ ] (arcade) Toggle still wonky
- [ ] (minecraft) Dropdown color not set in expandable arguments (on chat)

Non-regression bugs:
- [ ] Duplicating function does not correctly rename
- [ ] (arcade) Color field doesn't highlight selected color

pelikhan:
- [ ] printing

![image](https://user-images.githubusercontent.com/4175913/77801432-14ef6680-7036-11ea-984e-b47842a64ca8.png)

shakao:
closing for now as higher priority bugs have been fixed, will reopen as necessary/when more are discovered

PR #2691, opened 2020-03-27
Render selected text instead of image in gesture handler

shakao:
Fix for image dropdown rendering in the gesture field (https://github.com/microsoft/pxt-microbit/issues/2678)


pelikhan:
Could we move this kind of "image grid" fields to pxt-core? @shakao

PR #2695, opened 2020-03-30
Hotpotato game

pelikhan:
Hot potato game. 

RE:: https://github.com/microsoft/pxt-adafruit/blob/master/docs/projects/hot-potato.md


ganicke:
A real potato this time. The Adafruit version uses a "fake" duct tape potato.

ganicke:
Hey, on an unrelated note, could you use `radio.setTransmitPower()` to add a "social distancing" factor to the "Infection" game?

pelikhan:
yes we could but even with power = 1, you still have a quite a bit of range. We could use RSSI instead.

PR #2706, opened 2020-03-31
Add spy versions of tutorials

pelikhan:
Creating a "text" version of the tutorial. Not changing the existing tutorials for legacy.


ganicke:
`checkdocs` in the build is choking on its snippet collect I think.

pelikhan:
Right, i've fix some paths, should be good now.

ganicke:
Doh, I could've checked those.

ganicke:
Wanna replace `() =>` to `function()` in those?

PR #2725, opened 2020-03-31
Set volume block

pelikhan:
Add rudimentary volume control using the PWM analog value.
- [x] hardware support
- [x] sim support


pelikhan:
Here is a sample program:
```
    vol = Math.max(0, vol - 8)
    led.toggle(1,0)
    music.setVolume(vol)
    music.playTone(Note.C, 200)
})
    vol = Math.min(0xff, vol + 8)
    led.toggle(1,0)
    music.setVolume(vol)
    music.playTone(Note.C, 200)
})
input.onButtonPressed(Button.AB, function () {
    basic.showNumber(vol)
})
let vol = 0
vol = 255
basic.forever(function () {
    led.toggle(0, 0)
    basic.pause(100)
})
```

libs/core/pins.cpp
```diff
@@ -298,6 +298,7 @@ namespace pins {
 
 
     MicroBitPin* pitchPin = NULL;
+    int pitchVolume = 64;
```
mmoskal, line 301:
```suggestion
    uint8_t pitchVolume = 64;
```

libs/core/pins.cpp
```diff
@@ -311,6 +312,26 @@ namespace pins {
...
 
+    /**
+    * Sets the volume the pitch pin
+    * @param volume the intensity of the sound from 0..255
+    */
+    //% blockId=device_analog_set_pitch_volume block="analog set pitch volume $volume"
+    //% help=pins/analog-set-pitch-volume weight=3 advanced=true
+    //% volume.min=0 volume.max=255
+    void analogSetPitchVolume(int volume) {
+        pitchVolume = max(0, min(0xff, volume | 0));
```
mmoskal, line 323:
```suggestion
        pitchVolume = max(0, min(0xff, volume));
```

libs/core/pins.cpp
```diff
@@ -311,6 +312,26 @@ namespace pins {
       pitchPin = getPin((int)name);
     }
 
+    /**
+    * Sets the volume the pitch pin
```
mmoskal, line 316:
```suggestion
    * Sets the volume on the pitch pin
```

libs/core/pins.cpp
```diff
@@ -321,10 +342,12 @@ namespace pins {
     void analogPitch(int frequency, int ms) {
       if (pitchPin == NULL)
         analogSetPitchPin(AnalogPin::P0);
-      if (frequency <= 0) {
+      if (frequency <= 0 || pitchVolume <= 0) {
         pitchPin->setAnalogValue(0);
       } else {
-        pitchPin->setAnalogValue(512);
+        // sound coming out of speaker is not linear, try best match
+        int v = 1 << (pitchVolume >> 5);
```
mmoskal, line 349: this maxes out at `1 << 7 == 128`; I would think that 512 should be audibly lauder. How about:
`pitchVolume * pitchVolume * pitchVolume >> 15` ?

libs/core/pins.cpp
mmoskal: (the `| 0` doesn't do anything in C)

libs/core/pins.cpp
pelikhan: i don't have a buzzer unfortunately; only headphones. I suspect that have vastly different behaviors as far as sound output.

pelikhan: the sound level sounds quickly the same after 64 on headphones.

pelikhan: I'll get this in and we can try a better mapping later.

mmoskal: No problem. Most users will use headphones.

pelikhan:
I figure I optimized for those. Kills your ears a bit less.

PR #2755, opened 2020-04-01
Remove background handlers

mmoskal:
As discussed on call.

This keeps event dispatch in DAL (without map on pxt side).


mmoskal:
@finneyj  https://travis-ci.org/github/microsoft/pxt-microbit/builds/669778901#L730 - it seems that your tag is wrong; I think I was tagging on microbit#master and that worked better

Issue #2694, opened 2020-03-30
Show arrow resets to default when switching back from JS view


microbit-henry:
> micro:bit support: 28409

**Describe the bug**
When a 'show image (arrow image (direction)) at offset 0' block is edited in Javascript to show a different direction, and then the editor is switched back to blocks, all directions are set to north.

**To Reproduce**
1. Go to makecode.microbit.org
2. Create a sequence of blocks with at least 2 'show image (arrow image (direction)) at offset 0' blocks in. Make sure neither block is showing an arrow pointing north.
3. Switch to javascript.
4. Change one of the 'images.arrowImage(ArrowNames.South).showImage(0)' lines to show a different, non-north direction. 
5. Switch back to blocks
6. Note that all 'show image (arrow image (direction)) at offset 0' blocks are now pointing north.

**Expected behavior**
When the direction of the arrows is changed in javascript, these states would be expected to be preserved when moving back to blocks, with the correct directions being represented with the right blocks.

**Screenshots**

![MakecodeBugSmall](https://user-images.githubusercontent.com/53710678/77894930-e6f46700-726d-11ea-9863-416b684e05a7.gif)

**Desktop (please complete the following information):**
 - OS: iOs
 - Browser: Chrome
 - Version: 80

PR #2765, opened 2020-04-02
Remove shadow from arrow image block

shakao:
Fixes https://github.com/microsoft/pxt-microbit/issues/2694


pelikhan:
We should really add those arrows to showicon.

PR #2801, opened 2020-04-08
fixed mispelling of funereal to funeral

### LINKED ISSUE #2193: Funereal/Funeral Melody Mismatch
**Describe the bug**
The melody is called `Funereal` in blocks, but is called `Melodies.Funeral` in JavaScript. 

**Expected behavior**
The two should match

image

image

hristoramirez:
fixes #2193

### DIFF
--- a/libs/core/_locales/core-strings.json
+++ b/libs/core/_locales/core-strings.json
@@ -155,7 +155,7 @@
   "Melodies.Chase|block": "chase",
   "Melodies.Dadadadum|block": "dadadum",
   "Melodies.Entertainer|block": "entertainer",
-  "Melodies.Funeral|block": "funereal",
+  "Melodies.Funeral|block": "funeral",
   "Melodies.Funk|block": "funk",
   "Melodies.JumpDown|block": "jump down",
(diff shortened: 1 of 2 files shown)


libs/core/_locales/core-strings.json
```diff
@@ -155,7 +155,7 @@
   "Melodies.Chase|block": "chase",
   "Melodies.Dadadadum|block": "dadadum",
   "Melodies.Entertainer|block": "entertainer",
-  "Melodies.Funeral|block": "funereal",
+  "Melodies.Funeral|block": "funeral",
```
darzu, line 158: Nice job!

abchatra:
Is there a backward compat issue here? @riknoll ?

riknoll:
It will break translations, that's all

riknoll:
But also, isn't funereal technically correct? Maybe worth fixing just because it *looks* wrong even though it is a word

darzu:
Oh wow, I had no idea funereal was a real word. It was "funeral" in JS, so maybe one was a typo?

PR #2820, opened 2020-04-15
Light packet

pelikhan:
Support for common-packages light packet.
- [x] building from common packages
- [ ] sim support
- [ ] make library neutral, remove "neopixel" brand


libs/light/lightSendBuffer.asm
```diff
@@ -0,0 +1,66 @@
+
+sendBufferAsm:
```
mmoskal, line 2: we should add this function to microbit-dal and nuke it here

pelikhan: Completely agreed. This is more of a placeholder right now.

PR #2865, opened 2020-04-21
Point brightness

pelikhan:
Dual of "plot brightness", reads the current brightness state of an LED.


abchatra:
What rationale for the new block?

pelikhan:
Symmetry. We have a block to set the LED brightness, but we don't have it to read it. We also support this for images.

abchatra:
Symmetric to which block?
image

pelikhan:
plot brightness

libs/core/led.cpp
```diff
@@ -62,18 +62,17 @@ namespace led {
...
      * @param x the horizontal coordinate of the LED
      * @param y the vertical coordinate of the LED
      */
-    //% help=led/point weight=76
-    //% blockId=device_point block="point|x %x|y %y"
+    //% help=led/point-brightness weight=76
+    //% blockId=device_point_brightness block="point|x %x|y %y brightness"
     //% parts="ledmatrix"
     //% x.min=0 x.max=4 y.min=0 y.max=4
     //% x.fieldOptions.precision=1 y.fieldOptions.precision=1
```
abchatra, line 73: //% advanced=true

abchatra:
Ok looks good. Other than it should be in more category right?

pelikhan:
Moved to more.

pelikhan:
image

PR #2873, opened 2020-04-21
Use primary color for details main button

shakao:
i like the purple more than the green but it's still a little hard to read when unfocused/not hovering, i think....

image

image


pelikhan:
Yup still not great contrast. Can we always use a white font?

pelikhan:
Let’s review this at design meetings. Colors....

PR #2879, opened 2020-04-22
Image literal rows fix

pelikhan:
Better support for number of rows != 5 and "imageLiteralRows". The "trick" is to use a shim to send the image to ``images::createImage`` and implement the simulator code in STS.

Test:
```
namespace custom {
    /**
     */
    //% block="foo"
    //% imageLiteral=1
    //% imageLiteralColumns=17
    //% imageLiteralRows=7
    //% shim=images::createImage
    export function foo(i: string): Image {
        const im = <Image><any>i;
        return im
    }
}

let im = custom.foo(`
    # . . . . . . . . . . # # . . . .
    . # . . . . . . . . # . . # . . .
    . . # # . . . . . # . . . # . . .
    . . . # . . . . # . . . . . # . .
    . . . . # . . . # . . . . . # . .
    . . . . # # . . . . . . . . . # .
    . . . . . # # # . . . . . . . . #
    `)
im.scrollImage(1, 200)
```

### DIFF
--- a/libs/core/images.cpp
+++ b/libs/core/images.cpp
@@ -83,7 +83,7 @@ void showImage(Image sprite, int xOffset, int interval = 400) {
 //% parts="ledmatrix"
 void plotFrame(Image i, int xOffset) {
     // TODO showImage() used in original implementation
-    plotImage(i, xOffset * 5);
+    plotImage(i, xOffset * i->img->height);
 }
 
 /**
@@ -178,6 +178,6 @@ bool pixel(Image i, int x, int y) {
 //% weight=70 help=images/show-frame
 //% parts="ledmatrix"
 void showFrame(Image i, int frame, int interval = 400) {
-    showImage(i, frame * 5, interval);
+    showImage(i, frame * i->img->height, interval);
 }
 } // namespace ImageMethods
--- a/sim/state/ledmatrix.ts
+++ b/sim/state/ledmatrix.ts
@@ -86,7 +86,7 @@ namespace pxsim {
     }
 
     export function createImageFromBuffer(data: number[]): Image {
-        return new Image(data.length / 5, data);
+        return new Image((data.length / 5) | 0, data);
     }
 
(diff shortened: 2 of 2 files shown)


MTKilpatrick:
Since these changes were made (at my request for help with creating a 17x7 image editor) the basic,.showIcon() function is causing a crash with Microbit error 020 in certain circumstances. I have no idea why/how. If I revert to the non-beta Makecode editor and download the same code to my Microbit, it functions correctly and showIcon() does not crash.

pelikhan:
Thanks for reporting. Do you have a snippet?

MTKilpatrick:
It's a big game code using the Pimoroni scrollbit.  I don't know if it will help. Because if I insert an additional showIcon() command just before the IF statement that then calls the problems showIcon(), it no longer crashes. The following function is called at every iteration of my FOREVER loop. If remove the show HEART icon, the code will crash at the show SAD icon!

function checkWorm (worm: number) {
    other = 1 - worm
    a = MyWorms[worm].xPos()
    z = MyWorms[worm].yPos()
    // first check target but not if the other worm also
    // there
    if (a == targetx && z == targety) {
        if (MyWorms[other].xPos() != a || MyWorms[other].yPos() != z) {
            MyWorms[worm].incScore()
            MyWorms[worm].grow()
            MyWorms[worm].plotScore()
            destroyTarget()
        }
    }
    if (MyWorms[other].isAnySegmentAt(a, z) || MyWorms[worm].isAtSelf()) {
        basic.showIcon(IconNames.Sad)
        MyWorms[worm].decScore()
        restart = true
        basic.pause(500)
    }
}

MTKilpatrick:
Note: I also tried replacing the basic.showIcon(IconNames.Sad) with a showImage() command after defining a custom image to a variable. That also caused a crash - but again if I inserted an extra showImage before the offending IF statement, the crash did not happen.

Issue #2839, opened 2020-04-17
Filter out deprecated API's from Intellisense


Jaqster:
input.onShake comes up in intellisense, but all our Tutorials use the new onGesture API's

![image](https://user-images.githubusercontent.com/22108954/79599265-11ca2200-809a-11ea-80b7-d29637e8f056.png)

darzu:
Looks like onShake was just missing the "deprecated=true" annotation. Intellisense skips it when it's deprecated.
https://github.com/microsoft/pxt-microbit/pull/2941
Are there any others that we know are missing?

PR #2941, opened 2020-05-01
deprecate onShake

darzu:
We prefer onGesture now I believe. Need confirmation: @abchatra or @riknoll 

Fixes: @https://github.com/microsoft/pxt-microbit/issues/2839

onShake no longer appears in intellisense:
Screen Shot 2020-04-30 at 7 19 03 PM


pelikhan:
We prefer to be able to round trip so yes hide them.

PR #2948, opened 2020-05-01
red-light-green-light project fixes

pelikhan:
A few issues found while running the activity. This should be split into 2 separate guided tutorials...


ganicke:
Want to do a two-page split out of this?

Issue #2971, opened 2020-05-04
Better organization for Music blocks


Jaqster:
Hal suggests adding subcategories to the Music Toolbox drawer for better organization:
**Tone**
•	play tone [note] for [duration]
•	ring tone (Hz) [note]
•	rest (ms) [duration]
•	 [note]
**Melody**
•	start melody [melody] repeating [duration]
•	play melody [sequencer] at tempo [value] bpm
•	stop melody [all]
•	music on [event]
**Volume**
•	set volume [value]
**Tempo**
•	[duration] beat
•	tempo (bpm)
•	change tempo by (bpm) [value]
•	set tempo to (bpm) [value]

![image](https://user-images.githubusercontent.com/22108954/80931024-3d246080-8d6c-11ea-99c8-2160cec281fc.png)

pelikhan:
See https://github.com/microsoft/pxt-microbit/issues/2971

abchatra:
Fixed

microbit-mark:
Not sure I agree that 'start melody' is an advanced feature? It seems simpler than creating a melody sequence from scratch ![Screenshot 2020-05-05 at 10 53 06](https://user-images.githubusercontent.com/31242877/81054389-a87e4780-8ebe-11ea-8062-1b2990559886.png)

The in initial set of melodies is common across MakeCode and MicroPython, so if you move between the two, you have a consistent experience. Could we not address this https://github.com/microsoft/pxt-microbit/issues/2835 by having 'play melody' as the old block and 'play melody sequence' as the new one. @halspeed @jaustin

pelikhan:
We want to encourage kids to craft their own melodies, not just pick one from a list. It is “hard fun”. The other blocks are still there, but lower in the list.

microbit-mark:
Revisiting this as we discussed not calling this section 'Advanced` @abchatra @Jaqster 
Also stop Melody still works with the play melody blocks, so it doesn't seem to make sense to hide it. Could we use a 'more' section for the older block?

Also, resource authors may be referencing the `start Melody` block and if they find `play Melody` first they might mistakenly use this which could cause confusion

abchatra:
Is it ok to rename this to "More"?

microbit-mark:
Re-opening so we can discuss outside the calls @halspeed @jaustin

abchatra:
Fixed

PR #2974, opened 2020-05-04
Music reorg

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/2971 . Same as suggested aside from:

- moving new melody block on top
- moving old melody blocks to bottom

image

### DIFF
--- a/libs/core/music.ts
+++ b/libs/core/music.ts
@@ -173,6 +173,7 @@ enum MusicEvent {
  * Generation of music tones.
  */
 //% color=#E63022 weight=106 icon="\uf025"
+//% groups='["Melody", "Tone", "Volume", "Tempo", "Melody Advanced"]'
 namespace music {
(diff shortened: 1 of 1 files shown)


pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/2971

Jaqster:
Are we trying to deprecate the old melody blocks?  If we are, fine.  But if not, the new Play Melody block is not really a substitute...

abchatra:
One suggestion. Instead of _Melody Advanced_, make sub title just _More_. We can lump un-important blocks there.

abchatra:
_Are we trying to deprecate the old melody blocks? If we are, fine. But if not, the new Play Melody block is not really a substitute..._

Once we introduce play melody, very few users will use start melody blocks.

pelikhan:
Yes we are trying to align melody system across all boards. The new block is more creative.

Issue #3003, opened 2020-05-09
serial.readLine waits for \n and not \r


sjmurdoch:
The [documentation for serial.readLine()](https://makecode.microbit.org/reference/serial/read-line) indicates that the function waits until a \r character before returning. Actually it seems to wait for a \n. At least on MacOS, using screen as recommended, the delimiter sent by pressing enter will be \r and so readLine() will appear to not work. 

As can be seen from the [source of readLine()](https://github.com/microsoft/pxt-microbit/blob/32a7cc0d00f9f384782380c64f345093081c6b7f/libs/core/serial.ts#L91), this function calls readUntil() with a [delimiter of “\n”](https://github.com/microsoft/pxt-microbit/blob/32a7cc0d00f9f384782380c64f345093081c6b7f/libs/core/serial.ts#L105).

For me, calling readUntil() directly with “\r” directly fixed the issue with screen on MacOS. I however have not tested what would happen with other serial terminal software and/or other operating systems.

pelikhan:
Fix deployed in beta. please test if this works for me. You should be able to change NEW_LINE_DELIMITER for your Mac.

sjmurdoch:
If I set `serial.NEW_LINE_DELIMITER` to `Delimiters.CarriageReturn` then `serial.readLine()` works fine using screen on MacOS X. Thanks for sorting this out so quickly.

PR #3008, opened 2020-05-11
use \r as new line delimiter

pelikhan:
Fix for https://github.com/microsoft/pxt-microbit/issues/3003


libs/core/serial.ts
```diff
@@ -102,7 +102,7 @@ namespace serial {
         // (memory) efficient than the C++ implementation, because the
         // strings are statically allocated and take no RAM
         switch (del) {
-            case Delimiters.NewLine: return "\n"
+            case Delimiters.NewLine: return "\r"
```
mmoskal, line 105: I think it would be better to add Delimiters.Enter or Return or  CarriageReturn. There may be already people using that and passing data in from files.

mmoskal: We should also fix the docs.

sjmurdoch: I would also be interested to know what is the behaviour on platforms (e.g. TeraTerm on Windows, which I can't test). I was a bit surprised to find an issue in code that's been present for years, which makes me suspect it works for other people. Perhaps Windows does something different from MacOS here, and changing the delimiter to \r could break other platforms. I couldn't find any specification for what is the right behaviour of serial terminal emulators and suspect there's wide variation.

pelikhan: To follow @mmoskal  suggestion, I have left the existing behavior unchanged, updated the docs and added a carriage return delimiter option. Since this behavior has been in effect for years, changing might break an unknown set of existing programs.

mmoskal: I think you can often change it in the settings of the terminal, what it will send when you press enter. I think a single `\r` is quite common though.

libs/core/_locales/core-jsdoc-strings.json
```diff
@@ -558,6 +558,7 @@
   "pins.pulseIn": "Return the duration of a pulse at a pin in microseconds.",
   "pins.pulseIn|param|name": "the pin which measures the pulse, eg: DigitalPin.P0",
   "pins.pulseIn|param|value": "the value of the pulse, eg: PulseValue.High",
+  "pins.pushButton": "Mounts a push button on the given pin",
```
sjmurdoch, line 561: Is this relevant to the commit?

libs/core/_locales/core-strings.json
```diff
@@ -62,12 +62,15 @@
...
+  "Delimiters.CarriageReturn|block": "carriage return (\r)",
   "Delimiters.Colon|block": ":",
   "Delimiters.Comma|block": ",",
   "Delimiters.Dollar|block": "$",
   "Delimiters.Fullstop|block": ".",
   "Delimiters.Hash|block": "#",
-  "Delimiters.NewLine|block": "new line",
+  "Delimiters.NewLine|block": "new line (\n)",
+  "Delimiters.Pipe|block": "|",
+  "Delimiters.SemiColumn|block": ";",
```
sjmurdoch, line 73: Should be SemiColon (not SemiColumn)

libs/core/serial.ts
```diff
@@ -1,3 +1,24 @@
...
+    //% block="."
+    Fullstop = 5,
+    //% block="#"
+    Hash = 6,
+    //% block="carriage return (\r)"
+    CarriageReturn = 7,
+    //% block="|"
+    Pipe = 8,
+    //% block=";"
+    SemiColumn = 9,
```
sjmurdoch, line 19: Should be SemiColon (not SemiColumn)

libs/core/serial.ts
```diff
@@ -108,6 +129,9 @@ namespace serial {
             case Delimiters.Colon: return ":"
             case Delimiters.Fullstop: return "."
             case Delimiters.Hash: return "#"
+            case Delimiters.CarriageReturn: return "\r"
+            case Delimiters.Pipe: return "|"
+            case Delimiters.SemiColumn: return ";"
```
sjmurdoch, line 134: Should be SemiColon (not SemiColumn)

libs/core/_locales/core-jsdoc-strings.json
pelikhan: No. It's a string from another change ; that hasn't build into the strings. ignore.

libs/core/serial.ts
```diff
@@ -1,3 +1,24 @@
+const enum Delimiters {
+    //% block="new line (\n)"
+    NewLine = 1,
```
mmoskal, line 3: maybe we could do NewLine = 10, Comma = 44, etc. and then have:
```
export function delimiters(del: Delimiters): string { return String.fromCharCode(del as number) }
```
This way you can always cast any number to `Delimiters` if you need to.

pelikhan: I guess we have that function now

mmoskal: (the memory efficiency comment on that function no longer applies due to MSTR())

pelikhan: The delimiters function is mostly a helper for blocks, the API takes a string already.

Issue #2563, opened 2019-12-19
include base/control.ts from common packages in microbit


mmoskal:
**Is your feature request related to a problem? Please describe.**

Folks were asking for certain features that are in common pkgs but not in microbit, in particular:

- [x] https://forum.makecode.com/t/read-free-memory-at-runtime/919 - free memory

- [x] https://forum.makecode.com/t/how-to-go-about-an-interval-extension/812 - intervals

- [ ] https://forum.makecode.com/t/how-can-i-access-the-name-of-my-program/764 - program name (this one was Ev3 but microbit isn't much of a stretch)

abchatra:
@pelikhan will this come as your common-packages plan?

pelikhan:
Yes we should bring this in.

pelikhan:
I did not bring in program name but the rest is there. @mmoskal  close?

mmoskal:
yeah, let's wait until someone asks

PR #3009, opened 2020-05-11
Control support

pelikhan:
Integrate more common control apis. Fix for https://github.com/microsoft/pxt-microbit/issues/2563

- [x] gc stats
- [x] interval
- [ ] program name
- [ ] pauseUntil - requires an API from CODAL

requires https://github.com/microsoft/pxt-common-packages/pull/1106


pelikhan:
@mmoskal do we need programName?

PR #3032, opened 2020-05-13
Move neopixel sendBuffer primitive into core

mmoskal:
Tested as this:

```typescript
//% shim=light::sendNeopixelBuffer
function send(pin: DigitalPin, buf: Buffer): void {}
```

### DIFF
--- a/libs/core/enums.d.ts
+++ b/libs/core/enums.d.ts
@@ -559,7 +559,6 @@ declare namespace led {
     //% block=1200
     BaudRate1200 = 1200,
     }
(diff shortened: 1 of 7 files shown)


pelikhan:
chat about this?

pelikhan:
let's do an upgrade rule that updates the shim

pelikhan:
Rename to Neopixel to ws2812b

mmoskal:
requires https://github.com/microsoft/pxt/pull/7044

PR #3047, opened 2020-05-13
Match send message text in 'microchat'

### LINKED ISSUE #3042: Content inconsistency in "micro chat" python tutorial
Notice we send a ":)"
But later it mentions we should receive a "Yo"

Screen Shot 2020-05-13 at 8 25 14 AM
Screen Shot 2020-05-13 at 8 25 26 AM

ganicke:
Send message in the snippet `":)"` didn't match the description `"Yo"`. Went with the smiley emoticon text. The non-spy version uses "Yo". Maybe I should change that too.

Fixes #3042


pelikhan:
Yes smiley is better, no localization issues.

Issue #3103, opened 2020-05-20
console.log does not support any


pelikhan:
We need to align our console impl with common packages. It does not have the any signature.

![image](https://user-images.githubusercontent.com/4175913/82481682-8d98ff00-9a8a-11ea-8c0f-d7afcba04694.png)

PR #3135, opened 2020-05-21
Support any in console log

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3103

This code is taken from here:

https://github.com/microsoft/pxt-common-packages/blob/master/libs/base/console.ts#L81

I tried to get rid of the console implementation in micro:bit altogether and use the one in common packages but it seems like there are too many differences now in the control APIs (which are used internally). Also, we changed the signature of `console.addListener()` in common packages so we would need upgrade rules

### DIFF
--- a/libs/core/console.ts
+++ b/libs/core/console.ts
@@ -17,7 +17,6 @@ namespace console {
      */
     //% weight=90
     //% help=console/log blockGap=8
-    //% text.shadowOptions.toString=true
     export function log(text: any): void {
         let stringified = inspect(text);
         // pad text on the 32byte boundar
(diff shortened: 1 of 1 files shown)


libs/core/console.ts
```diff
@@ -18,14 +18,15 @@ namespace console {
     //% weight=90
     //% help=console/log blockGap=8
     //% text.shadowOptions.toString=true
```
jwunderl, line 20: I think we wanna drop this line too or else it'll do weird things

riknoll: good catch! these apis don't have blocks so I don't think it matters but good to remove anyways

PR #3138, opened 2020-05-21
Show rounded rope value in 'Tug of LED'

### LINKED ISSUE #3025: The moving step is not accurate in tutorial 'Tug-Of_LED'
**Describe the bug**
The moving step is not accurate in tutorial 'Tug-Of_LED'

1.Navigate to https://makecode.microbit.org/beta#
2.Game->Tug-Of_LED
3.Step one by one

**Expect behavior**
1. The dot should move one step after clicking button A 10 times.
2. Everything should be normal when restarting the simulator

**Actual behavior**
1. The dot move one step after clicking button A 1 time. If changing rope by -1, the situation is same as before.
2. The dot will flash when restarting the simulator
exactly

**Additional context**
1.OS: Windows(rs6)
2.makecode.microbit.org version: 2.3.52
3.Microsoft MakeCode version: 5.37.62

ganicke:
Add a step to 'Tug of LED' tutorial to show the rounded value of `rope` instead of the default truncated.

Fixes #3025

### DIFF
--- a/docs/projects/spy/tug-of-led.md
+++ b/docs/projects/spy/tug-of-led.md
@@ -55,6 +55,18 @@ input.onButtonPressed(Button.B, function () {
 
 ## Step 5
 
+Because a button press moves the rope by **0.1** in either direction, plot the ``||math:rounded||`` value of ``||variables:rope||`` instead.
+
+```spy
+let rope = 2
+basic.forever(function() {
+    basic.clearScreen()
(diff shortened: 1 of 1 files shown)


ganicke:
OK, still want to take this change as a demo of the use of `round`?

pelikhan:
Let's remove it when we ship...

pelikhan:
maybe...

PR #3186, opened 2020-05-28
Extensions: Add planetx kit

microbit-mark:
@lionyhw @elecfreaks1

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -248,6 +248,11 @@ Check out [microbit.org](http://microbit.org/resellers/) for more information on
 
 ```codecard
 [{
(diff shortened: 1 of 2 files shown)


abchatra:
@lionyhw @elecfreaks1

Quick comment on the package dependency:
Any reason to hard code to particular version?  github:Microsoft/pxt-ws2812b#v0.0.4
It is better if you just point to the latest version to get any future fixes.

lionyhw:
> @lionyhw @elecfreaks1
> 
> Quick comment on the package dependency:
> Any reason to hard code to particular version? github:Microsoft/pxt-ws2812b#v0.0.4
> It is better if you just point to the latest version to get any future fixes.

No restrictions, fixed

PR #3215, opened 2020-06-03
Update extensions gallery

microbit-mark:
- Recategorised extensions
- added TOC

What do you think @pelikhan @abchatra? I think this covers the categories of extension much better. We could have a button link from microbit.org to each category

Should the TOC be in the gallery template?
Should each category have a 'back to top' link at the bottom? If so, would this be in the gallery template too?

We'd need to either merge existing PRs or I'd edit them based on this updated template.


pelikhan:
We could add a SUMMARY.md in that folder to get a proper TOC

ganicke:
Let me add a summary and see how that looks...

ganicke:
I added a SUMMARY and changed the section links to cards.

ganicke:
Seems that `pxt checkdocs` fails the section link format in `SUMMARY.md`.

microbit-mark:
Thanks for the work on this @ganicke. Is there anything I can do to help progress it?

ganicke:
If I pull out the SUMMARY it will build. BTW, did you want the "Utilities" extensions in the list too?

microbit-mark:
Thanks for this. I think it would be good to add the link to "Utilities" too if possible.

Issue #3210, opened 2020-06-02
Get rid of Basic - More subcategory


Jaqster:
There are only 2 blocks in there anyway.  And Clear Screen is a commonly used block that shouldn't be hidden.

![image](https://user-images.githubusercontent.com/22108954/83546327-562c4880-a4b5-11ea-90a9-2e3ad3d52ecf.png)

pelikhan:
Yup, flatten this category.

PR #3216, opened 2020-06-03
Remove basic.more subcategory

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3210

Removes the advanced = true tags for the two blocks in the basic:more subcategory


pelikhan:
Where do those blocks end up located in the basic category? Can you upload a scrreenshot?

livcheerful:
> Where do those blocks end up located in the basic category? Can you upload a scrreenshot?

@pelikhan yup, so the 'clear screen' block appears near the top under "show string", and the 'show arrow' block is at the very bottom

image
image

PR #3219, opened 2020-06-03
don't load webusb support in winrt

pelikhan:
Don't plug webusb packetio in winrt mode.


editor/extension.tsx
```diff
@@ -48,9 +49,11 @@ pxt.editor.initExtensionsAsync = function (opts: pxt.editor.ExtensionOptions): P
         subclassCode: 0x03
     }])
 
-    res.mkPacketIOWrapper = flash.mkPacketIOWrapper;
     res.blocklyPatch = patch.patchBlocks;
     res.renderBrowserDownloadInstructions = dialogs.renderBrowserDownloadInstructions;
-    res.renderUsbPairDialog = dialogs.renderUsbPairDialog;
+    if (!pxt.winrt.isWinRT()) {
```
abchatra, line 54: Do you need to check for electron app too?

pelikhan:
This is handled in pxt itself.

PR #3229, opened 2020-06-04
Daplink version

pelikhan:
Query and log daplink interface version


pelikhan:
Not sure how valuable this info is, but it might help us understand the state of devices on the field.

pelikhan:
it's kind of useless since it always returns 1.10. closing.

PR #3240, opened 2020-06-08
Bump dal to v2.2.0-rc6 (fixes #3231)

### LINKED ISSUE #3231: Edge event for pin is not working
**Describe the bug**
Rising or falling edge event for pin is not fired.

**To Reproduce**
```
pins.setEvents(DigitalPin.P8, PinEventType.Edge)
control.onEvent(EventBusSource.MICROBIT_ID_IO_P8, EventBusValue.MICROBIT_PIN_EVT_RISE, function () {
    led.toggle(0, 0)
})
```
This code doesn't work and the LED is not toggle when sending high to pin 8.

**Expected behavior**
This works on the released version but not the beta version.

mmoskal:


mmoskal:
we should test the resulting build on hw

PR #3272, opened 2020-06-12
Releasing 3.0.13 to live 

abchatra:
Has following fixes https://github.com/microsoft/pxt/pull/7137
https://github.com/microsoft/pxt/pull/7171
https://github.com/microsoft/pxt/pull/7180
and https://github.com/microsoft/pxt/pull/7181


abchatra:
Try fixes at /beta

PR #3284, opened 2020-06-18
Releasing 3.0.16

abchatra:


pelikhan:
when are you releasing? i need to know when i can use arrays in python  :)

abchatra:
3:00pm today. Waiting for folks to test their fixes.

PR #3287, opened 2020-06-18
Extensions: add time and date

microbit-mark:
cc @bsiever

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -630,6 +630,10 @@ Check out [the accessories pages on microbit.org](https://microbit.org/buy/acces
 
 ```codecard
 [{
+  "name": "Time & Date",
+  "url":"/pkg/bsiever/microbit-pxt-timeanddate",
+  "cardType": "package"
+}, {
   "name": "Kitronik Realtime Clock",
   "url":"/pkg/KitronikLtd/pxt-kitronik-rtc",
   "cardType": "package"
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -178,7 +178,8 @@
             "elecfreaks/pxt-PlanetX",
             "elecfreaks/pxt-nezha",
             "philipphenkel/pxt-powerfunctions",
-            "1010Technologies/pxt-makerbit-ir-transmitter"
+            "1010Technologies/pxt-makerbit-ir-transmitter",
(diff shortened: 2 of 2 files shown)


abchatra:
@bsiever

bsiever:
It is unwieldy.  I'm open to alternatives, but I thought about this a lot and chose this block style for a reason.   Time is atomic, so non-atomic access leads to problems.  Using smaller value (reporter?) blocks could lead to inconsistent values.  For example, if code is reading the time at 1:59.59 an "hour reporter" could read the hour as 1.  A "minute reporter" may run after the second has ended and report "0".  The time would appear to be 1:00 rather than 2:00.  

Another option is to return a composite object that represents the date/time and then provide "getter" methods that extract individual components. Your suggestion of using the timestamp falls into this category.  I thought that was clunky and it requires a lot more blocks.  It's also still easily prone to the non-atomic access problem.  I think "get object" block would just be passed to every "extract component", which would lead to the same problems described above.  (That is, rather than `let t = timeStamp()` and `let h = hourOf(t); let m=minuteOf(t)`, there'd be stuff like:  `let h = hourOf(timeStamp()); let m = minuteOf(timeStamp())').  

Another options was using separate "with blocks", each having a subset of this block's variables...but that has the same potential problems with consistency.  It would be possible to nest them and get an inconsistent set of values.

One "middle ground" option:  I could remove the "day of year" and "weekday" and create blocks that compute them from the MM/DD/YYYY.  (This would actually be re-computing them.  They are natural by-products of the process, so I just passed them through).

abchatra:
Thanks for the explanation @bsiever. It is your decision. Typically such large blocks will consume so much space it is harder to navigate in the smaller devices like chromebook.

I can merge the change if you are good with this.

bsiever:
Thanks @abchatra :  Go ahead and merge please...I'll continue to try to think about a better approach.  Perhaps if people actually use it I can gather feedback on what's actually useful/not and update the API.

BTW, earlier today I updated my repo with an initial test.ts.   I hope to finish it up in the next few days.

  Bill

bsiever:
@abchatra I decided to take your advice and make the `current time as numbers` block a bit shorter (still long-ish though). 

Thanks!

Issue #3296, opened 2020-06-25
Add default 'volume' variable


microbit-giles:
**Is your feature request related to a problem? Please describe.**
I would like a default system 'volume' variable similar to the 'tempo (bpm)' block.

It would make a project like this simpler: https://makecode.microbit.org/_6X9Hz28Kk3VP

Although the workaround is not complex, the fact that we have a **tempo** and LED **brightness** default system variable block suggests a similar **volume** one should exist.

**Describe the solution you'd like**
I would like a default system 'volume' variable similar to the 'tempo (bpm) block to be added to music blocks.

**Describe alternatives you've considered**
Create own volume variable as per https://makecode.microbit.org/_6X9Hz28Kk3VP - but it's long-winded

abchatra:
This is good issue. We have set volume and we need a get block too. @pelikhan

PR #3298, opened 2020-06-25
Adding get volume block

abchatra:
Fixes https://github.com/microsoft/pxt-microbit/issues/3296


abchatra:
@ganicke would it be possible to add help for this API?

abchatra:
It is pair API with setVolume which we have documentation already.

ganicke:
Yes, I'll add it.

PR #3301, opened 2020-06-26
Extensions: add maqueen plus

microbit-mark:
cc @tangjie133

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -353,6 +353,10 @@ Check out [the accessories pages on microbit.org](https://microbit.org/buy/acces
 
 ```codecard
 [{
+  "name": "DF Robot Maqueen Plus",
+  "url":"/pkg/DFRobot/pxt-DFRobot-Maqueenplus",
+  "cardType": "package"
+}, {
   "name": "Kitronik :MOVE Motor",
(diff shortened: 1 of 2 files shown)


tangjie133:
@abchatra Hi,abchatra，How long can it be approved？

abchatra:
Sorry for the delay. Here are the comments:

- License says: "_Copyright (c) 2018 Sam El-Husseini"_.This should be changed to right copyright.
- This block seems typo:
image
- Category color looks incorrect
image


Optional feedback
Some of the blogs don't have student friendly names.
- what does CW mean in the following block:
image
- What does T & E mean below
image
You should look at the blocks and see how to make them student friendly.

Looks good otherwise.

tangjie133:
Thank you for your advice，The third recommendation cannot be modified because it is generated automatically by the compiler，The second suggestion is not very clear. Please correct me，Thank you very much..

abchatra:
>> The second suggestion is not very clear. Please correct me

What does _(stop:0 forward: 1 back:2)_ mean in that block? Dropdown doesn't have these values.

abchatra:
Is the return value?

tangjie133:
yes

PR #3340, opened 2020-08-10
Extensions: add magic wand

microbit-mark:
cc @lionyhw


lionyhw:
@abchatra Can you check this PR?

PR #3341, opened 2020-08-12
Gallery page for blocks

ganicke:
A page with API lists to show all the rendered blocks. 

**Note**: Image copy still requires screen caps or snippers. Also, built-ins don't render with this.

RE: https://github.com/microsoft/pxt-minecraft/issues/1902


ganicke:
Got garbage commits in my stale master.

PR #3389, opened 2020-09-23
Generalize 'MakeCode' instead of 'app' in UCP Science

### LINKED ISSUE #3388: Remove MakeCode App notice on the Science Course
Now that we have WebUSB, let's remove the notification that you have to have the MakeCode windows app to do this course.
https://makecode.microbit.org/courses/ucp-science
image

ganicke:
De-emphasizing the use of "Windows 10 MakeCode app" to just "MakeCode" in the UCP Science lessons. Since WebUSB is in place, the use of the app is no longer required. The app is left as an option in the setup pages where it is mentioned.

Closes #3388


Jaqster (approved):
Thanks Galen!

PR #3401, opened 2020-10-06
"coding for teachers" video series

pelikhan:
- [x] added stu lowe video series
image

- [x] check for youtube for playlists
- [x] randomize live video orders


Jaqster:
Looks great.  Where are you going to put this row on the Home page?  Feels like we could organize things a bit better - Tutorials, Projects, and Videos all grouped together.

pelikhan:
It goes at the bottom. We have the tools to generate rows based on playlists so Stu would have to organize it better and we can integrate.

Issue #3376, opened 2020-09-16
Android: Hide download complete/pairing modal when using app


microbit-mark:
**Is your feature request related to a problem? Please describe.**
As the Android app is using Chrome in the webview, the WebUSB pairing dialogues are visible. Whilst it's technically possible to use it, it confuses the experience of using an app designed for BLE flashing.

Selecting the more menu by Download opens the pairing options

![image](https://user-images.githubusercontent.com/31242877/93305614-a4602000-f7f6-11ea-9dc3-20732d11ad37.png)

If a BLE flash is successful we return to a 'Download Complete' modal that shows a 'Pair Device' option. Since we commonly talk about `pairing` in terms of Bluetooth, I think this could really confuse people.

![image](https://user-images.githubusercontent.com/31242877/93305628-ab872e00-f7f6-11ea-9587-30432715f307.png)

**Describe the solution you'd like**
Hide this modal dialogue when using the app. We pass a URL parameter to the editor of `?androidapp=27` could we hide the modal when the editor uses this address? I think you may have done something similar for iOS

**Describe alternatives you've considered**

**Additional context**

PR #3407, opened 2020-10-12
add 'androidapp' queryparameter to disable webUSB

jwunderl:
close https://github.com/microsoft/pxt-microbit/issues/3376


abchatra (approved):
Looks good to me. I am assuming androidapp is sent as part of the app.

PR #3445, opened 2020-10-15
Extensions: add micro:bit extension

microbit-mark:
cc @microbit-matt-hillsdon @jaustin 

I have not added this to the gallery as it doesn't fit into a particular category, but we can do if we feel it's helpful.

We might still want to update the name in pxt.json before merging

### DIFF
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -1,7 +1,7 @@
 {
     "packages": {
         "approvedRepos": [
-            "pxt-microbit-v2-extension",
+            "microbit-foundation/pxt-microbit-v2-extension",
             "microsoft/pxt-neopixel",
             "microsoft/pxt-microturtle",
             "microsoft/pxt-sonar",
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -1,6 +1,7 @@
 {
     "packages": {
         "approvedRepos": [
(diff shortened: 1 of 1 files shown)


targetconfig.json
```diff
@@ -1,6 +1,7 @@
 {
     "packages": {
         "approvedRepos": [
+            "pxt-microbit-v2-extension",
```
abchatra, line 4: This is incorrect. Needs the full path.

abchatra (changes requested):

targetconfig.json
microbit-mark: Whoops. Long week :). Done in https://github.com/microsoft/pxt-microbit/pull/3445/commits/df1f1f814f0a88de76f232cd1fe419a6e3b79191

abchatra:
Extension also requires png.

microbit-mark:
> Extension also requires png.

@abchatra Does https://github.com/microbit-foundation/pxt-microbit-v2-extension/blob/master/icon.png not cover this?

abchatra:
It does. Sorry somehow missed it. Can we have v2 marked in big in that microbit png?

abchatra (approved):

PR #3457, opened 2020-10-21
SImulator v2 feel

pelikhan:
Detect that v2 is being used

- [x] notches
- [x] v2 text
- [x] head in gold
- [x] Detects it from the parts.

image


jwunderl (approved):
Ahh, I was adding a separate parts entry for v2 to add it for the other sound blocks and making that parse / load correctly (e.g. `parts="headphones,v2"`), separating it out to speaker vs headphones is much simpler

PR #3459, opened 2020-10-21
Logo touch in v2 sim

pelikhan:
- [x] mounts the logo as a touch button;
- [x] sim support

TODO (future PR): don't switch to breadboard layout when button is used. Better touch target; accessilibility

image


jwunderl (approved):
Looks good, tested on my micro:bit as well. I can handle conflicts in my PR~

PR #3460, opened 2020-10-21
V2sim updates

jwunderl:
couple changes:

* `speaker` never gets to the sim, so adding a `v2` part (probably useful for future more processing power intensive stuff as well?)
* check whether it's v2 before initializing the svg, sometimes it was loading without updating for me
* v2 text: make it a gold circle as we mentioned, and make it reposition slightly when A+B is used
* round the inner edge of the notches slightly (closer to the one on the microbit, but just a hand edit still)

image

image


sim/dalboard.ts
```diff
@@ -147,22 +147,23 @@ namespace pxsim {
...
-                visual: boardDef.visual,
-                boardDef: boardDef,
-                highContrast: msg.highContrast
-            }), opts);
 
             if (opts.partsList) {
                 const v2 = opts.partsList.indexOf("microphone") > -1
                     || opts.partsList.indexOf("logotouch") > -1
-                    || opts.partsList.indexOf("speaker") > -1;
+                    || opts.partsList.indexOf("v2") > -1;
```
pelikhan, line 154: let's rename this to "onboardspeaker"

pelikhan (approved):

PR #3462, opened 2020-10-22
Microhpone v2

pelikhan:
- [x] align with microphone blocks
- [x] register as v2 feature
- [ ] needs HW testing


pelikhan:
@ganicke more docs needed here

libs/bluetoothprj/pxt.json
```diff
@@ -2,7 +2,8 @@
     "name": "{0} block",
     "dependencies": {
         "core": "file:../core",
-        "bluetooth": "file:../bluetooth"
+        "bluetooth": "file:../bluetooth",
```
jwunderl, line 5: There's a missingpackage upgrade rule we probably need to load this into old projects, right?

jwunderl (commented):
👍 Tested on hardware, seems to work fine! A loud-ish clap right next to it reads at about ~195

PR #3465, opened 2020-10-22
setTouchMode

pelikhan:
Added v2 block to set touch mode on pins


pelikhan:
@ganicke some more docs needed here

mmoskal (approved):

PR #3480, opened 2020-10-23
Round plotted value in blocks Tug-of-LED

### LINKED ISSUE #3025: The moving step is not accurate in tutorial 'Tug-Of_LED'

ganicke:
Round the plotted value of rope in the Blocks version of "Tug of LED".

RE: #3138

Fixes #3025 again

### DIFF
--- a/docs/projects/tug-of-led.md
+++ b/docs/projects/tug-of-led.md
@@ -32,7 +32,6 @@ basic.forever(function() {
 ## Step 3
 
 Add an event on ``||input:button A pressed||`` to change the ``||variables:rope||`` value by **-0.1**.
-The @boardname@ will automatically round the ``||variables:rope||`` value to the nearest LED coordinate.
 
 ```blocks
 let rope = 2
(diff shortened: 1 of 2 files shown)


docs/projects/tug-of-led.md
```diff
@@ -51,34 +50,45 @@ input.onButtonPressed(Button.B, function () {
...
-
 ## Step 5
 
+Because a button press pulls the rope by **0.1** in either direction, plot the ``||math:round||`` value of ``||variables:rope||`` to the nearest LED.
+
+```spy
+let rope = 2
+basic.forever(function() {
+    basic.clearScreen()
+    led.plot(Math.round(rope), 2)
```
pelikhan, line 61: this seems wrong. we should  be rounding the index in the API itself. is it not working on hardware?

pelikhan (approved):

PR #3484, opened 2020-10-26
use new offline-app page

jwunderl:
Need to wait for release to pick up the new docfiles

mbit-offline-app-page


jwunderl:
For post release, requires the docfiles in stable6.0~

abchatra:
Not sure I understand that comment Joey. We can merge this post release right?

jwunderl:
Ah, yes, that's what i meant - we have to wait for release to merge this, as the release will include files this change is dependent on

Issue #3453, opened 2020-10-21
Asynchronous binary UART in background (NovaFitness SDS011)


irukard:
Hi!

I'm not sure this report belongs to pxt-microbit. Maybe problem lay even deeper in core. Please advise.

I'm working on extension for makecode. I want to read binary serial data from NovaFitness SDS011 air quality sensor. Protocol is very simple, 10 bytes of data incoming every second. Datagram looks like this:

```
Header: AA C0
PM 2.5 value: XX XX
PM 10 value: XX XX
ID: XX XX
Checksum XX
Tail: AB
```

I managed to read the data using serial.readBuffer(10).

```
        sdsbuffer = serial.readBuffer(10)
        // check if frame starts with 0xAA 0xC0 and ends with 0xAB
        if (sdsbuffer.getNumber(NumberFormat.UInt8LE, 0) == 170
            && sdsbuffer.getNumber(NumberFormat.UInt8LE, 1) == 192
            && sdsbuffer.getNumber(NumberFormat.UInt8LE, 9) == 171) {

            pm25 = sdsbuffer.getNumber(NumberFormat.UInt16LE, 2) / 10
            pm10 = sdsbuffer.getNumber(NumberFormat.UInt16LE, 4) / 10
        }
```

But I encountered strange behavior. When I place block in main loop, everything works ok. Downside is that I have to wait for readBuffer to grab 10 bytes (which is done once per second, due to SDS011 data sending intervals). But when I place  readBuffer function in the background communication with micro:bit dies. Often shows 983 error code. 

Extension code: https://github.com/irukard/makecode-sds011/tree/d2607ea9d09d22ad1fc3d66a09c04825a346a156

It would be better to wait for any data on serial, like in micropythons uart.any(), and than read it byte by byte. That would be non blocking way, but I don't see that option in javascript.

I also tried to use serial.onDataReceived(serial.delimiters(0xAA), function () {}) but it doesn't work very well, strange things happen when 0xAA value is part of datagram. 

In current state (https://github.com/irukard/makecode-sds011/blob/8ca4a8eef51bad06233bf56a7732f290b51b40e6/sds011.ts) I made a "service" to read data in background. It uses infamous onDataReceived(serial.delimiters(0xAA). Code is very unstable. micro:bit hangs after 20-250 seconds. Very often without any visible error. Usually it reboots itself, however sometimes even reset button doesn't work. 

Code: https://makecode.microbit.org/_FCbgYge0wcM5
[...]

PR #3488, opened 2020-10-27
Serial readbuffer non blocking

pelikhan:
When a buffer read of length 0 is requested, read the buffered data in a non-blocking way.
Fix for https://github.com/microsoft/pxt-microbit/issues/3453


docs/alpha-ref.json
```diff
@@ -0,0 +1,3 @@
+{
+    "appref": "v"
```
abchatra, line 2: I don't want to send folks to different alpha link yet. I want everyone to be on /beta. I have ported the fixes necessary for /beta. Lets keep it there for next 2 weeks.

abchatra (changes requested):

docs/alpha-ref.json
pelikhan: removed

libs/core/serial.cpp
```diff
@@ -109,17 +109,21 @@ namespace serial {
...
+        length = uBit.serial.getRxBufferSize();
+        mode = ASYNC;
+      }
 
       auto buf = mkBuffer(NULL, length);
-      int read = uBit.serial.read(buf->data, buf->length);
+      int read = uBit.serial.read(buf->data, buf->length, mode);
       if (read != length) {
         auto prev = buf;
         buf = mkBuffer(buf->data, read);
```
mmoskal, line 129: this is unsafe; if there is a GC before the second buffer allocation, it will collect the first buffer and you will copy some junk

you can do registerGCPtr()/unregisterGCPtr() or you can use app_alloc() to allocate initial memory, then allocate mkBuffer() and then do app_free()

mmoskal (changes requested):

mmoskal (approved):

PR #3489, opened 2020-10-27
more logging in webusb

pelikhan:
A tinybit more of logging.

### DIFF
--- a/editor/flash.ts
+++ b/editor/flash.ts
@@ -4,7 +4,7 @@ const membase = 0x20000000;
 const loadAddr = membase;
 const dataAddr = 0x20002000;
 const stackAddr = 0x20001000;
-const FULL_FLASH_TIMEOUT = 60000;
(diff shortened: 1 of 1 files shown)


editor/flash.ts
```diff
@@ -212,13 +216,15 @@ class DAPWrapper implements pxt.packetio.PacketIOWrapper {
             })
             .then((res) => {
                 const binFile = resp.outfiles[this.binName];
+                log(`bin file ${this.binName}, ${binFile?.length || -1}b`)
```
mmoskal, line 219: maybe better use hexUint8 length? the binFile length is b64-encoded (so ~25% bigger)

mmoskal (approved):

editor/flash.ts
pelikhan: i'm really logging the binName which something seems to be out of wack. I'll add that on the next line

PR #3493, opened 2020-10-27
Don't use debug cycle counter for ws2812b

### LINKED ISSUE #3415: NeoPixel extension does not work with V2 board revision
**Describe the bug**
Using neopixel does not work on the latest micro:bit revision

image

**To Reproduce**
Flash a program using the neopixel extension and neopixels attached.
Works on V1.* not on V2

**Expected behavior**
Works on all board variants

**Screenshots**
Add screenshots to help explain your problem. You can copy paste the screenshot in the github report. The .gif screen recording is very useful as well. 

**Additional context**

mmoskal:
should be back-ported to codal later

The daplink was interfering with the neopixel bit bang impl that was using debug cycle counters

fixes #3415

### DIFF
--- a/libs/core/light.cpp
+++ b/libs/core/light.cpp
@@ -2,9 +2,15 @@
 
 #if MICROBIT_CODAL
 
-// WS2812B timings, +-0.15uS
-// 0 - 0.40uS hi 0.85uS low
-// 1 - 0.80uS hi 0.45uS low
+// WS2812B timings, datasheet v1
+// 0 - 0.25-0.55us hi 0.70-1.00us low
+// 1 - 0.65-0.95us hi 0.30-0.60us low
+// datasheet v5
+// 0 - 0.22-0.38us hi 0.58-1.00us low
+// 1 - 0.58-1.00us hi 0.58-1.00us low
+
+// 0 - 0.31us / 0.65us
+// 1 - 0.71us / 0.62us
 
 __attribute__((noinline, long_call, section(".data"))) static void
 neopixel_send_buffer(Pin &pin, const uint8_t *ptr, int numBytes) {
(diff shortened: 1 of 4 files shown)


pelikhan (approved):

AlasdairAtKitronik:
@mmoskal @pelikhan I've just been having a look through the code changes and noticed a couple of things.

The comment at the start about the WS2812B timings seems a bit concerning if that's the actual timings which are being used, as these don't meet the specs in any of the WS2812 datasheets (all sizes - 5050, 3535, 2020 - now have the same timings).
According to the WorldSemi datasheets, the timings should be as follows:

0: 220ns-380ns HI, 580ns-1us LO
1: 580ns-1us HI, 580ns-1us LO
RES: >280us

Even with the +/- 150ns in the comment, this still puts the 0 HI and 1 LO potentially out of spec.
The reset at 100us is also way out of spec.

Also, the function respecting brightness has 'brightness' spelled incorrectly (neopixel_send_buffer_brigthness), not sure if that's going to cause some issues somewhere down the line.

mmoskal:
Thanks!

Do you know if they changed silicon wrt timings or just the datasheets?

mmoskal:
(my timings came from here https://cdn-shop.adafruit.com/datasheets/WS2812B.pdf )

AlasdairAtKitronik:
> Thanks!
> 
> Do you know if they changed silicon wrt timings or just the datasheets?

I'm assuming they changed the silicon, don't know for certain. Thought I'd go direct to the source to see what they were saying (as in WorldSemi). We've had a lot of grief in the past with timings, which is why we're quite careful about checking!

mmoskal:
@AlasdairAtKitronik  I'm somewhat suspicious of these datasheet v5 timings - I think there's lots of code using the old timings that would break. Nevertheless, I tried to have the code fit in both old and new; needed to rewrite in assembly though.

Thanks for notice about the typo!

pelikhan (approved):

mmoskal:
@abchatra  this need merge to stable branch

PR #3517, opened 2020-10-29
support for getting data off microphone

mmoskal:
This adds:
* input.onSoundData(buf => ...)
* input.soundSamplingPeriod()
* input.setSoundSamplingPeriod()

The first one has to be implemented in pxt-microbit, not in extension, since it plugs into microphone pipeline before the level detector.

TODO:
* [ ] sim support


jaustin:
@finneyj has a model for an input pipeline to make sure that these things don't cause breakages with the existing blocks/features and also doesn't break the oversampling switch as we go between mic only to mic and ADC. Worth discussing I think @mmoskal

PR #3536, opened 2020-10-30
Always checksum

pelikhan:
- [x] If UICR check passes, validate checksums. If changed > 50%, go for full flash.
- [x] properly reset flashAborted state
- [x] if cortex halted, resume
- [x] avoid race where serial fails while flashing
 - [x] stop/start serial reader around quick flash


mmoskal (approved):

abchatra:
Does this change impact Windows App code path?

pelikhan:
It should not, untested though.

PR #3540, opened 2020-10-30
microphone led

pelikhan:
image


pelikhan:
@jaustin

jaustin:
Like it! Thanks Peli. Does it look large/clear enough in the context of the whole editor? I think from the screenshot probably yes.

Can the text and mic move so the relative location to the LEDS matches, or is the location governed by the text label? If the mic sat over the 'level bar' and the 128 to the right of that I think it'd be in about the right place.
@microbit-mark wondered about also showing the mic port so people know where the mic hole actually is?

image

microbit-mark:
I think it would be useful to show the input. For kids learning about microphones and sound and for teachers demoing it. ie. it's not a magical LED microphone, but a physical input on the board. :)

PR #3543, opened 2020-11-01
Stop all sounds

pelikhan:
Support for stopping sound expressions; e.g. button smashing a button and restarting the sound ASAP.

Untested on HW.


abchatra:
Does this also stop tones like 'play tone' & 'ring tone'?

pelikhan:
Yes ``rest(0)`` should stop those.

abchatra (approved):
Looks good to me. Can you share the screenshot of the blocks? Does 'sounds' category come before 'melody advanced'? Should this be just part of 'Volume' category?

pelikhan:
Well we need to decide where to put it. I have it right under "play melody" right now.

jwunderl (approved):
I think I'd vote for putting it as the last block under the volume group? That's where I'd start looking for it.

livcheerful (approved):
I like Joey's suggestion of in the Volume category -- under the Melody Advanced category with the "stop melody all" would also make sense to me. Then you're forced to ask yourself what the difference is

pelikhan:
Moved to volume.

PR #3573, opened 2020-11-03
Add microphone to older project

pelikhan:
Adding microphone to older projects.

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -113,11 +113,11 @@
                     }
                 }
             ],
-            "0.0.0 - 3.0.0": [
+            "0.0.0 - 3.0.18": [
                 {
                     "type": "missingPackage",
                     "map": {
-                        ".": "microphone"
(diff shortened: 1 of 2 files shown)


pxtarget.json
```diff
@@ -112,6 +112,14 @@
                         "pxt-microbit-v2-extension": "microphone"
                     }
                 }
+            ],
+            "0.0.0 - 3.0.0": [
```
abchatra, line 116: This should be 3.0.17 Right?

abchatra (changes requested):

jwunderl (approved):

pxtarget.json
jwunderl: [max is exclusive](https://github.com/microsoft/pxt/blob/42096c86dce1a19c013049a9a7dee5dcfba07874/pxtlib/semver.ts#L115-L116) so 3.0.18~

PR #3574, opened 2020-11-03
Use logo.isPressed() not io.logo.isTouched()

### LINKED ISSUE #3535: Strange behaviour with touch modes in V2
**Describe the bug**

1. If you use `logoIsPressed` in your program with the logo, it seems to be the case that on first run the logo touch mode is set to resistive rather than capacitive by default. It's hard to trigger a logo touch without grounding it. Using a button to set touch to capacitive resolves the issue.. but
2. Showstring doesn't seem to work properly. The string is not scrolled, it appears to be offset to the right. If you interupt this by touching the logo it seems to fix the issue and the string is scrolling.

image

[touch-test.hex.zip](https://github.com/microsoft/pxt-microbit/files/5464948/touch-test.hex.zip)

**Expected behavior**
Touch works as expected and strings are scolled on the display.

**Screenshots**
Add screenshots to help explain your problem. You can copy paste the screenshot in the github report. The .gif screen recording is very useful as well. 

**micro:bit version (please complete the following information):**

You can find this information in the lower right hand corner of the back of micro:bit (the side that says BBC micro:bit).
 - Which version of the micro:bit is this relevant to [ EG V1.3, V1.5, V2.0, or specify it's not hardware related ]

**Additional context**

mmoskal:
fixes #3535

### DIFF
--- a/libs/core/logo.cpp
+++ b/libs/core/logo.cpp
@@ -43,7 +43,7 @@ namespace input {
     //% help="input/logo-is-pressed"
     bool logoIsPressed() {
 #if MICROBIT_CODAL
-        return uBit.io.logo.isTouched();
+        return uBit.logo.isPressed();
 #else
         target_panic(PANIC_VARIANT_NOT_SUPPORTED);
         return false;


mmoskal:
@jaustin @abchatra this doesn't seem to change anything; I think it's in codal

finneyj:
Hmm... not so sure. I just tests this against v0.2.16, and the logo definitely seems to be in captouch mode:

```
void onTouchFace(MicroBitEvent e)
{
    DMESG("TOUCH: FACE");
}

void 
cap_touch_test()
{
    uBit.messageBus.listen(MICROBIT_ID_FACE, MICROBIT_BUTTON_EVT_CLICK, onTouchFace);

    while(1)
    {
        uBit.display.image.setPixelValue(2,4,uBit.logo.isPressed() ? 255 : 0);
        uBit.sleep(100);
    }
}
```

jaustin:
This is not for the event though, it's for the "logo is touched" block
which reads the value.

It was calling uBit.io.logo but @mmoskal also tried uBit.logo ( I presume
the getValue() method?)

finneyj:
@jaustin

No, sorry, that test uses both APIs - event based and polled. The event based API logs to DMESG, but the polled API controls an LED on the display. See the first line in the while(1) loop. This would cover the necessary test case wouldn't it?

I can see why uBit.io.logo.isTouched() would flip sensing of the logo into resistive mode (as that's the default we decided upon), but uBit.logo is configured for capacitative when it is initialised... that's why I brought it out as a first class button alongside buttonA and buttonB.

I'm wondering if there might be another rogue call to uBit.io.logo.isTouched() somewhere on the PXT side?

jaustin:
@finneyj my reading of the original bug, having also had a conversation with Mark about this previously, is that IF one ever calls explicit set of touch mode then things work. I think the first sentence of the issue isn't right but I've asked @microbit-mark to check the wording. My understanding is that if the first thing you call is the "logo is pressed" it's resistive (that's in line with the code before this PR that uses uBit.io.logo()

Does registering that event actually do anything? Or is it all initialised at uBit contruction?

I also wonder if this is slightly the result of flaky touch? This patch looks correct to me though. @finneyj what would you expect behaviour to be before this patch?

finneyj:
Agree this patch looks good. Before this patch, I would expect resistive behaviour, yes.

There's no need to use that event handler no, and it has no effect. I also tested without the event handler and got the same result.

Not sure what you mean by "flaky touch" though @jaustin?

finneyj:
Just to confirm - this also behaves capcitatively:

```
void 
cap_touch_test()
{
    while(1)
    {
        uBit.display.image.setPixelValue(2,4,uBit.logo.isPressed() ? 255 : 0);
        uBit.sleep(100);
    }
}
```

pelikhan (approved):

Issue #3616, opened 2020-11-06
Shake overlaps with microphone in v2


riknoll:
<img width="388" alt="Screen Shot 2020-11-06 at 11 20 07 AM" src="https://user-images.githubusercontent.com/13754588/98406113-40055400-2022-11eb-9f47-a38ca667578f.png">

jaustin:
In the past it was also possibe to wobble the mouse a lot on the sim to shake it. Was that an old TouchDevelop feature? IIRC also shake a tablet or phone would work?

pelikhan:
It's still possible to shake the mouse and trigger the shake event or in fact, any of the events. We map the mouse position to a quadratic force field.

We decided to not use the phone/ipad accelerometer. Mostly to keep this consistent but also because shaking the device triggers "undo" events in iOS which are very annoying.

jaustin:
The undo thing makes sense, thanks for lettting me know :). I didn't realise wobble-to-shake still worked - macOS makes my pointer large before it shakes but can confirm it's great and does work for me, especially easy if I go diagonally.

PR #3634, opened 2020-11-10
Move On Shake up

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3616

lemme know what you guys think! I can move it to other places if you think that'd be better This is what it looks like [updated]:
image

### DIFF
--- a/sim/visuals/microbit.ts
+++ b/sim/visuals/microbit.ts
@@ -482,9 +482,9 @@ path.sim-board {
                 accessibility.enableKeyboardInteraction(this.shakeButton, undefined, () => {
                     this.board.accelerometerState.shake();
                 });
-                accessibility.setAria(this.shakeButton, "button", pxsim.localization.lf("Shake the board"));
+                accessibility.setAria(this.shakeButton, "button", "Shake the board");
                 this.shakeText = svg.child(this.g, "text", { x: 420, y: 122, class: "sim-text-small" }) as SVGTextElement;
-                this.shakeText.textContent = pxsim.localization.lf("SHAKE");
+                this.shakeText.textContent = "SHAKE";
             }
         }
 
--- a/sim/visuals/microbit.ts
+++ b/sim/visuals/microbit.ts
@@ -482,7 +482,7 @@ path.sim-board {
                 accessibility.enableKeyboardInteraction(this.shakeButton, undefined, () => {
                     this.board.accelerometerState.shake();
                 });
-                accessibility.setAria(this.shakeButton, "button", "Shake the board");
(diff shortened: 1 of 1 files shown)


pelikhan:
Try using radio and compass in the same program.

pelikhan:
I think you want it under the 128, above the "b" label. SHAKE text would be smaller.

sim/visuals/microbit.ts
```diff
@@ -475,7 +475,7 @@ path.sim-board {
                     this.board.accelerometerState.shake();
                 });
                 accessibility.setAria(this.shakeButton, "button", "Shake the board");
-                this.shakeText = svg.child(this.g, "text", { x: 400, y: 110, class: "sim-text" }) as SVGTextElement;
+                this.shakeText = svg.child(this.g, "text", { x: 400, y: 50, class: "sim-text" }) as SVGTextElement;
                 this.shakeText.textContent = "SHAKE"
```
pelikhan, line 479: now that we support translations in the sim, we shoud consider translating this or using an icon instead.

pelikhan (changes requested):
Overlaps with radio antenna + RSSI text.

sim/visuals/microbit.ts
livcheerful: I added pxsim.localization.lf() in two places. Is that enough for translation? (or is there a good way for me to test it?)

pelikhan:
Try using the editor in a different locale.

pelikhan:
Best to do this in another PR.

sim/visuals/microbit.ts
```diff
@@ -72,6 +72,13 @@ namespace pxsim.visuals {
         pointer-events: none;
         }
 
+        .sim-text-small {
+            font-family:"Lucida Console", Monaco, monospace;
+            font-size:20px;
+            fill:#fff;
+            pointer-events: none;
+        }
+
```
jwunderl, line 81:
```suggestion
        .sim-text-small,
```

nit: seems like you can combine these two?

jwunderl (approved):

pelikhan (approved):
LGTM. Make sure we don't do anything with sim-text-pin in high contrast mode (where we have different sim colors).

sim/visuals/microbit.ts
livcheerful: Joey that is genius

livcheerful:
> LGTM. Make sure we don't do anything with sim-text-pin in high contrast mode (where we have different sim colors).

oh good point! I'll try it out

livcheerful:
High contrast mode doesn't change the simulator at all for me right now. I'm going to complete this PR, see what's wrong with HC, and if it affects the SHAKE text I'll open up a new PR to fix

Issue #3645, opened 2020-11-10
The new built in sounds for the micro:bit v2 are way too loud


riknoll:
The sounds from these blocks:

<img width="329" alt="Screen Shot 2020-11-10 at 8 44 05 AM" src="https://user-images.githubusercontent.com/13754588/98703913-e73eff80-2330-11eb-835b-206cdc500191.png">

They are too loud even if you call set volume with 1

abchatra:
discussed in the meeting. @jaustin to send new .wav files to @pelikhan to update.

pelikhan:
Someone from the MakeCode team could pick this up.

abchatra:
I can update the files.

jaustin:
Can we just check that the sound volume we care about is simulator not on the device?
Attached are a bunch of different volume reductions.
[quieter-sounds.zip](https://github.com/microsoft/pxt-microbit/files/5519774/quieter-sounds.zip)

```
for f in *.wav; do
    for vol in "0.01" "0.05" "0.1" "0.5"
    do
        sox -D -v $vol $f ${f%.*}-$vol.wav
    done
done
```

abchatra:
This is simulator specific issue as per @riknoll and not hardware related. Is that your question @jaustin ?

jaustin:
That was the question. Those wav files should still help based on what Peli was saying though

pelikhan:
Misunderstanding :) The .wav files are exclusively used in the simulator and have no influence on the hardware sounds.

abchatra:
-0.5 files are still high. I guess I will go with 0.1. @riknoll can you give it a try? 
@jaustin can we get 0.2?

riknoll:
I can generate 0.2! I'll try it out in the browser

PR #3647, opened 2020-11-10
Making the simulator less deafening

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3645

Swapping out the wav files for quieter versions and also dividing the volume of the melodies/tones by 10 (they were also Quite Loud).

Now everything sounds much more reasonable. The melodies are still pretty loud but much better than before


abchatra (approved):

jwunderl (approved):
Sounds good to me!

We might want to look at matching the volume in the play melody field editor field editor so it has about this volume, there's a very distinct difference between clicking a melody vs playing in a simulator after this (and simulator volume is much better)

Issue #3600, opened 2020-11-05
Change on-board speaker to built-in speaker for translatabiliy/clarity


jaustin:
It seems 'on-board' is less clear to translate from English to other langauges than 'built-in', and the effect in English is pretty similar so we prefer 'built-in' for the speaker block and API

<img width="243" alt="image" src="https://user-images.githubusercontent.com/147113/98272040-873f0680-1f88-11eb-91db-17342dddbfb3.png">
--> set built-in spekaer <on/off>

```setOnBoardSpeakerEnabled``` --> ```setBuiltInSpeakerEnabled```

pelikhan:
What about not putting this adjective?

________________________________

jaustin:
<img width="268" alt="image" src="https://user-images.githubusercontent.com/147113/98273309-054fdd00-1f8a-11eb-8ecb-c3c46ef55deb.png">
We've got some differing precedents there. I think I prefer keeping 'set'

pelikhan:
I meat remove « builtin  » or « on board »

________________________________

jaustin:
We don't wan't people to think this will apply to a connected speaker via an accessory. IE this will still make sound via P0 so we think 'built-in' is clearer. Lots of accessories/robots etc have speakers.

pelikhan:
I think onboard or builtin is the same in terms of translation since it is not going to be translated literally.

________________________________

jaustin:
@abchatra, @mmoskal, @carlosperate all felt that translating to their native language built-in was easier. That's not a huge sample size but here we've asked people and built-in is preferred anyway over on-board.

PR #3657, opened 2020-11-11
Change "on-board" speaker to "built-in"

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3600

The docs also weren't working for this block, so I fixed the path for that. Question - should I be able to test the docs locally?

### DIFF
--- a/docs/device/v2.md
+++ b/docs/device/v2.md
@@ -28,7 +28,7 @@ input.logoIsPressed()
 input.soundLevel()
 soundExpression.giggle.playUntilDone()
 soundExpression.giggle.play()
-music.setOnBoardSpeakerEnabled(false)
+music.setBuiltInSpeakerEnabled(false)
 pins.touchSetMode(TouchTarget.P0, TouchTargetMode.Capacitive)
 ```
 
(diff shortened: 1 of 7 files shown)


pelikhan (approved):

pelikhan:
The original extension used the "on board" name so our api was backward compatible; we should add a update rule to convert the old name to the current one. Otherwise looks great.

pelikhan (approved):

livcheerful:
> The original extension used the "on board" name so our api was backward compatible; we should add a update rule to convert the old name to the current one. Otherwise looks great.

oooooooh okay cool good to know I'll do that

jwunderl (approved):
You should be able to check the docs locally; http://localhost:3232/reference/music/set-built-in-speaker-enabled and the 'help' for the block should both show up~

livcheerful:
> The original extension used the "on board" name so our api was backward compatible; we should add a update rule to convert the old name to the current one. Otherwise looks great.

Question -- was the original extension only supported with beta?

pelikhan:
Yes 3.0 and above.

Issue #3673, opened 2020-11-12
microphone svg not rendered properly


pelikhan:
In Windows 10 app,

![image](https://user-images.githubusercontent.com/4175913/98927180-0d79b200-248d-11eb-9bb9-048ef6a23cf8.png)

abchatra:
Edge specific. Not a blocking bug.

PR #3680, opened 2020-11-13
Use an absolute path for microphone SVG

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3673

Old Edge does transformations differently than the other browsers, so I just made the microphone an absolute path with all the transformations baked in

it looks the same as before!
image


pelikhan (approved):
Looks great!

Issue #3644, opened 2020-11-10
Universal Hex format updates


microbit-carlos:
The current 512-byte block format results in very slow drag&drop flashes in DAPLink 0234 (over 30s for a minimal programme).
To fix this we've updated the format to use "sections" instead of blocks (main difference is only a single block start record is used for v1 and one for v2, but padding is still needed to align the two "sections"):  https://github.com/microbit-foundation/universal-hex/pull/10

jaustin:
Following up from the discusion on the call, a simple 'show LEDs' program on an 0234 micro:bit takes ~35s to flash.

As to why, I believe it's that 0234 processes certain records very slowly, and we have a bunch of those records in the 'block-like' format: 
```
:020000040000FA
:0400000A9901C0DEBA
:10000000C0070000D1060000D1000000B1060000CA
:1000100000000000000000000000000000000000E0
:100020000000000000000000000000005107000078
:100030000000000000000000DB000000E500000000
:10004000EF000000F9000000030100000D010000B6
:1000500017010000210100002B0100003501000004
:100060003F01000049010000530100005D01000054
:1000700067010000710100007B01000085010000A4
:100080008F01000099010000A3010000AD010000F4
:10009000B7010000C1010000CB010000D501000044
:0C00000C424242424242424242424242D0
:020000040000FA
:0400000A9901C0DEBA
:1000A000DF010000E9010000F3010000FD01000094
:1000B00007020000110200001B02000025020000E0
:1000C0001FB5C046C04600F0EFFA04B00FB41FBD24
:1000D00008205A49096809580847382057490968CB
:1000E000095808473C2055490968095808474020E5
:1000F0005249096809580847442050490968095875
:10010000084748204D490968095808474C204B4981
:10011000096809580847502048490968095808479C
:100120005420464909680958084758204349096836
:10013000095808475C204149096809580847602068
:0C00000C424242424242424242424242D0
:020000040000FA
:0400000A9901C0DEBA
:100140003E4909680958084764203C49096809582C
:100150000847682039490968095808476C20374919
:100160000968095808477020344909680958084740
:100170007420324909680958084778202F490968CE
:10018000095808477C202D490968095808478020EC
:100190002A490968095808478420284909680958E4
:1001A0000847882025490968095808478C202349B1
:1001B00009680958084790202049096809580847E4
:1001C00094201E4909680958084798201B49096866
:1001D000095808479C201949096809580847A02070
:0C00000C424242424242424242424242D0
[...]

PR #3718, opened 2020-11-16
Use more compact hex representation for faster flashing

mmoskal:
fixes https://github.com/microsoft/pxt-microbit/issues/3644

@abchatra I think you wanted this merged

### DIFF
--- a/compiler/combiner.ts
+++ b/compiler/combiner.ts
@@ -11,7 +11,7 @@ namespace ts.pxtc.extension {
 
         let outp = ""
 
-        wrapHex(mbdal, 0x00, [0x99, 0x01, 0xc0, 0xde])
+        wrapHex(mbdal, 0x00, [0x99, 0x00, 0xc0, 0xde])
         wrapHex(mbcodal, 0x0D, [0x99, 0x03, 0xc0, 0xde], true)
 
         outp += ":00000001FF\n"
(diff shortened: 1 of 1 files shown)


mmoskal:
tested as working in https://github.com/microsoft/pxt-microbit/issues/3644#issuecomment-728180515

pelikhan (approved):

compiler/combiner.ts
```diff
@@ -11,7 +11,7 @@ namespace ts.pxtc.extension {
 
         let outp = ""
 
-        wrapHex(mbdal, 0x00, [0x99, 0x01, 0xc0, 0xde])
+        wrapHex(mbdal, 0x00, [0x99, 0x00, 0xc0, 0xde])
```
jaustin, line 14: @mmoskal sadly this change, while correct for the spec, seems to trip up the Android App's parser. It's an easy fix in the app, but as there's a bit of a propogation time for app releases, I think it might make sense to revert to 0x01 now. @martinwork can you confirm the iOS app is okay with 9900 and 9901?  @microbit-sam and thoughts? Is it JUST the first 9900->9901 that you need  in order for the app to work?

mmoskal: should this stay with the spec-correct value on master and only be changed in stable?
cc @abchatra

abchatra: Yes lets keep it same in the master. Just patch fix stable3.0.

jaustin: Works for me.

microbit-sam: Yep, that'll work for me

Thanks - apologies for the change!

pelikhan: I don't understand the fix: are we taking this fix in the next release?

microbit-sam: The Android app's parser is written using the 0.3.0. spec for universal hexes, and looks for 9901 blocks
https://github.com/microbit-foundation/spec-universal-hex/blob/master/docs/format.md#block-type

I've made the change to use 9900, but the exisiting app on peoples devices won't handle the 9900 blocks, and it will take a week or so to: build a release, submit to app store, installed on X% of devices

martinwork:
I think 9900 would trip up the iOS. I have been looking at https://github.com/microbit-foundation/spec-universal-hex/blob/master/docs/format.md, which only mentions 9901. Where is the current spec?

compiler/combiner.ts
martinwork: I think 9900 would trip up the iOS. I have been looking at https://github.com/microbit-foundation/spec-universal-hex/blob/master/docs/format.md, which only mentions 9901. Where is the current spec?

mmoskal: https://github.com/microbit-foundation/spec-universal-hex/blob/spec-v0.4/docs/format.md#block-type

I think it would be wise then to move the new spec back to 9901, and revert this on both stable and master.

martinwork: I should have mentioned that the released iOS will fall over the change in softdevice and bootloader. The current beta needs work, so I will add 9900 as a possibility anyway.

microbit-sam: Is it also possible to switch back to 16 byte data records for the time being?
Partial Flashing in the Android app expects 16 byte records and misses half of the data, I'm changing this to a more flexible approach but it won't be live until the 9900/9901 fix also goes out

mmoskal: So in short, neither of the apps work now, but are being updated. I think we should leave this as is with 9900 and 32 bytes then and wait for app updates.

microbit-carlos: Right so if the apps currently need 9901 and 16 byte records we have two options for the imminent release:
- Go back to the old block format
    - This is bad for 0234 where it takes too long to flash
- Keep the new format, but temporarily change to 9901 and the 16 byte block record
    - Then move back to the 32 byte format after a few weeks when the apps have been updated and deployed
    - The board ID doesn't matter too much, but it is worth moving back to 32 bytes to improve flashing time
        - So if we are going to have a temporarily change anyway for the data length we could also do the same for the 9900 id

jaustin: @mmoskal I prefer the second of Carlos's two suggestions (9901 and 16 byte). Does that work for you?

mmoskal: done, see https://github.com/microsoft/pxt-microbit/pull/3728

martinwork: @mmoskal @jaustin My new build of the iOS app (and the current beta) is failing to do a partial flash with V2. The magic in the 9903 part of a new file downloaded from MakeCode beta is at 0x44400 rather than being on a page boundary. I think this happened before and we decided to stick to page boundaries. Am I remembering wrong?

mmoskal: added fix in https://github.com/microsoft/pxt-microbit/pull/3728

PR #3720, opened 2020-11-16
set sound pin

pelikhan:
Add a new block, ``music.setSoundPin`` to control which pin the sound should generated to.

- [x] on v1, it has the same backward compat effect of ``analogSetPitchPin``
- [x] on v2, it enables generating sound to a pin via the mixer
- [x] on v2, analogSetPitchPin is deprecated and has the effect of bypassing the mixer (backward compat square waves)
image

This way: existing packages using analogSetPitchPin will get square waves; packages who want to leverage the mixer on a pin, can use this api.


libs/core/pins.cpp
```diff
@@ -332,6 +332,7 @@ namespace pins {
     //% help=pins/analog-set-pitch-pin weight=3 advanced=true
     //% name.fieldEditor="gridpicker" name.fieldOptions.columns=4
     //% name.fieldOptions.tooltips="false" name.fieldOptions.width="250"
+    //% deprecated
```
jwunderl, line 335: Do we need to worry about the blocks being removed mid year? I'm guessing it's probably fine / unlikely that teachers were using these

jwunderl (approved):

libs/core/pins.cpp
pelikhan: it's an advanced API that most beginners will never find (advanced section in pins under advanced) so hiding to really discourage using it. We'll see,

pelikhan: on the other hadn, does not hurt to keep it around eighter.

abchatra:
Sorry late here. I think this block belongs in Pins & not Music. Student wouldn't understand what Pin means in a Music category.

image

Issue #3506, opened 2020-10-28
State of the Extension world


pelikhan:
See the documentation on how to fix in https://support.microbit.org/en/support/solutions/articles/19000121371-makecode-extension-compatibility-v1-and-v2

- [ ] error:  [tinkertanker/pxt-iot-environment-kit](https://github.com/tinkertanker/pxt-iot-environment-kit) build error / [new issue](https://github.com/tinkertanker/pxt-iot-environment-kit/issues/new?title=build%20error&body=Exit%20code%3A%201%20from%20cd%20temp%2Fghpkgs%2Ftinkertanker%2Fpxt-iot-environment-kit%3B%20pxt%20build%20--ignoreTests)
- [ ] error:  [microsoft/pxt-bluetooth-midi](https://github.com/microsoft/pxt-bluetooth-midi) build error / [new issue](https://github.com/microsoft/pxt-bluetooth-midi/issues/new?title=build%20error&body=Exit%20code%3A%201%20from%20cd%20temp%2Fghpkgs%2Fmicrosoft%2Fpxt-bluetooth-midi%3B%20pxt%20build%20--ignoreTests)
- [ ] error:  [laboratoryforplayfulcomputation/pxt-blockytalkyble](https://github.com/laboratoryforplayfulcomputation/pxt-blockytalkyble) build error / [new issue](https://github.com/laboratoryforplayfulcomputation/pxt-blockytalkyble/issues/new?title=build%20error&body=Exit%20code%3A%201%20from%20cd%20temp%2Fghpkgs%2Flaboratoryforplayfulcomputation%2Fpxt-blockytalkyble%3B%20pxt%20build%20--ignoreTests)
- [ ] error:  [microsoft/pxt-bluetooth-temperature-sensor](https://github.com/microsoft/pxt-bluetooth-temperature-sensor) build error / [new issue](https://github.com/microsoft/pxt-bluetooth-temperature-sensor/issues/new?title=build%20error&body=Exit%20code%3A%201%20from%20cd%20temp%2Fghpkgs%2Fmicrosoft%2Fpxt-bluetooth-temperature-sensor%3B%20pxt%20build%20--ignoreTests)
[...]

PR #3726, opened 2020-11-18
Disabling extensions in the v2 which are not building

abchatra:
List from #3506


mmoskal (approved):

pelikhan (approved):

abchatra:
I tested using pxt-bluetooth-temperature-sensor extension. Compilation succeeds and I can flash v1 hardware. If I flash v2 hardware with the same hex file it fails with 506 error. I believe this is expected.

abchatra:
I will merge the change tomorrow. @jaustin for visibility.

mmoskal:
I think it should be panic 927

jaustin:
Yes, 506 is from the DAPLink range, so suggests something else.

jaustin:
@abchatra can you share the hex file please? Or is there an editor instance that does this?

abchatra:
It gave me E927 on second try. I am not sure what is wrong with my v2. Probably before I haven't updated bootloader to 0255 yet. @jaustin can you send me instruction?

jaustin:
Yea, can send instructions, but for now can you share a hex so we can do a full test across other platforms in case there's something related to the hex. 

506 can be caused if there's still webUSB activity going on when drag-and-drop starts

4tronix:
Panic 927 is described here as "You are trying to use a feature designed for micro:bit V2 on a V1 board." This is clearly going to confuse users. https://support.microbit.org/support/solutions/articles/19000016969-micro-bit-error-codes
I released the fix for BitCommander but then started getting this 927 error without any explanation :(

mmoskal:
@abchatra The 4tronix line needs to be updated to:

```
            "4tronix/bitcommander": "min:1.1.1",
```

note that it's `1.1.1` not `v1.1.1`, because that's how the tag in repo is called. Should the tag be `v2.3.4` then it would be `min:v2.3.4`.

@4tronix you can submit a PR if you want.

@jaustin the error description should be extended with something like 'or the extension you're using is only compatible with v1'; the actual error is `variant not supported`.

microbit-mark:
Thanks @4tronix @mmoskal I have updated the error description accordingly.

PR #3728, opened 2020-11-18
Use 9901 for v1 and use 16 byte record

mmoskal:
(Android app compat)

See https://github.com/microsoft/pxt-microbit/pull/3718#discussion_r526098133

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -153,7 +153,8 @@
         "mbcodal": {
             "compile": {
                 "flashUsableEnd": null,
-                "flashEnd": null
+                "flashEnd": null,
+                "flashCodeAlign": 4096
             },
             "compileService": {
                 "buildEngine": "codal",
--- a/compiler/combiner.ts
+++ b/compiler/combiner.ts
@@ -11,7 +11,7 @@ namespace ts.pxtc.extension {
 
         let outp = ""
 
-        wrapHex(mbdal, 0x00, [0x99, 0x00, 0xc0, 0xde])
+        wrapHex(mbdal, 0x00, [0x99, 0x01, 0xc0, 0xde])
         wrapHex(mbcodal, 0x0D, [0x99, 0x03, 0xc0, 0xde], true)
 
         outp += ":00000001FF\n"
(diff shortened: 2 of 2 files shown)


pxtarget.json
```diff
@@ -153,7 +153,8 @@
         "mbcodal": {
             "compile": {
                 "flashUsableEnd": null,
-                "flashEnd": null
+                "flashEnd": null,
+                "flashCodeAlign": 4096
```
mmoskal, line 157: This should fix https://github.com/microsoft/pxt-microbit/pull/3718#discussion_r526199117

Also note that both flash ends now set to `null` should be replaced with 491520 so we get error message when the binary is too big, but I have no bandwidth to test that

martinwork: 491520 = 0x78000. Should that be 0x73000 = 471040?

jaustin:
@mmoskal is there an instance we can test with these hex files, or just a hex file please?

mmoskal:
There's always a link at the end of travis build log, here https://makecode.microbit.org/app/5627355602874826dbee4ab6aed7057ace1542e9-43e8e9c2f4

(this only applies for PRs on targets not pxt or pxt-c-p, but here we're good)

pxtarget.json
mmoskal: Can this be updated then? https://github.com/lancaster-university/codal-microbit-v2/blob/master/lib/README.md#creating-o-from-hex (the linker file a screen below)

mmoskal: and this https://github.com/lancaster-university/codal-microbit-v2/blob/master/ld/nrf52833.ld is `0x77000`...

jaustin:
Thanks, that's really helpful. We'll test this widely now (focusing just on the hex part, not the wider MakeCode features, as I presume this is against master so not representative of the release).

mmoskal:
No, this is based on stable3.0

pxtarget.json
martinwork: That first link is out of date. The LD script is right I think. While not knowing what flash end is, I was assuming, it would be the end of the application region. 78000 is one page into the bootloader at 77000.
@microbit-sam Where is the memory map info?

martinwork:
> https://makecode.microbit.org/app/5627355602874826dbee4ab6aed7057ace1542e9-43e8e9c2f4

Tested OK BLE flashing with the live iOS app to V1 and my development version of the app to V1 and V2 - full and partial flashing.

jaustin:
@abchatra we've now confirmed this works on Android too.

abchatra (approved):

PR #3779, opened 2020-11-28
display a neopixel strip as a matrix

tballmsft:
- the neopixel extension has setMatrixWidth already, which allows the mapping of an (x,y) coordinate to a linear neopixel strip
- this change (coupled with https://github.com/microsoft/pxt-common-packages/pull/1200) provides a helper method, light.setMatrixWidth(pin,width) that instructs the simulator to display the neopixel strip as a matrix. 

Example code:
```
let strip = neopixel.create(DigitalPin.P1, 256, NeoPixelMode.RGB)
strip.setMatrixWidth(16)
light.setMatrixWidth(DigitalPin.P1, 16)
for(let i=0; i<16; i++) {
    strip.setMatrixColor(i, i, neopixel.colors(NeoPixelColors.Red))
    strip.setMatrixColor(15-i, i, neopixel.colors(NeoPixelColors.Blue))
}
strip.show();
```

image


pelikhan:
It works for rectangular shapes? Can you specify row vs column order?

pelikhan:
There should be an argument specify the topology of the matrix, see https://learn.adafruit.com/adafruit-neopixel-uberguide/neomatrix-library.

tballmsft:
I gave it some thought This would require lots more changes. Not worth it at the moment.

PR #3807, opened 2020-12-10
Extensions: add DumbDisplay

microbit-mark:
cc @trevorwslee


microbit-mark:
@trevorwslee We're unable to approve this as it is incompatible with micro:bit V2 https://github.com/trevorwslee/MCExtension-DumbDisplay/blob/4e0b89a115e9496edbf4f739eab64469601780d4/pxt.json#L25

Please can you update it to use DAL/CODAL APIs as per this article https://support.microbit.org/en/support/solutions/articles/19000121371-makecode-extension-compatibility-v1-and-v2

PR #3827, opened 2021-01-06
document width setter for neopixel

tballmsft:
move setWidth function from lights to pins and document


jwunderl (approved):
Changes look good, my understanding is this one's not released yet / only available in beta so no need for upgrade rules?

tballmsft:
correct. just beta

PR #3845, opened 2021-01-21
Extensions: add BEP

microbit-mark:
cc @Bouw-je-BEP

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -377,6 +377,10 @@ Check out [the accessories pages on microbit.org](https://microbit.org/buy/acces
 
 ```codecard
 [{
+  "name": "Bouw je BEP",
+  "url":"/pkg/Bouw-je-BEP/Bouw-je-BEP",
+  "cardType": "package"
+}, {
   "name": "DF Robot Maqueen Plus",
   "url":"/pkg/DFRobot/pxt-DFRobot-Maqueenplus",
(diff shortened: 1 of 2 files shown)


abchatra:
@Bouw-je-BEP. Looks great overall.

One comment:

Language text is not rightly translated for the blocks. You need to have English text in the block definition and _locales/**/ file for any additional language. Otherwise it will be confusing for the users.  You don't need to have all languages under locales/** just the languages you care. 

Read more about how to rightly translate:
https://makecode.com/extensions/localization

Bouw-je-BEP:
@abchatra thank you for the feedback and information!

Our accessory is only available in the Netherlands at the moment that's why the block definitions are in dutch. I'll take a look into translating via _locales// 

Is it possible for the extension to be added so our dutch users can already benefit from it?

abchatra:
Thanks @Bouw-je-BEP 
Our filtering service might remove the blocks if they are of not the right format. So better to adhere to the semantics.

Bouw-je-BEP:
Hey @abchatra 

I updated te repo and changed all the dutch to english. I will make a translation in the coming weeks to be added.

Thanks!

abchatra (approved):

abchatra:
Thanks I will merge this. 

It is straight forward to add dutch. You need _locales/nl/ json file for strings.

PR #3852, opened 2021-01-26
Switch to GitHub actions

riknoll:
This uses the same setup as pxt-maker, but runs the push job on every push and not just master (so that it works on stable branches). Also deletes the travis.yml

I'm marking this as draft so that it doesn't get merged until I have filled in all the necessary environment variables


abchatra (approved):

jwunderl (approved):

livcheerful (approved):

riknoll:
Okay, slight update! There are now three build configurations:

1. main - includes all secrets and runs on each push on main
2. push - runs on every non-main push and does not include the crowdin key
3. pr - runs on all prs and doesn't include any secrets

.travis.yml
```diff
@@ -1,27 +0,0 @@
...
-# - "(cd libs/lang-test1; node ../../node_modules/pxt-core/built/pxt.js run)"
-# - "(cd libs/lang-test0; node ../../node_modules/pxt-core/built/pxt.js test)"
-# - "(cd libs/lang-test1; node ../../node_modules/pxt-core/built/pxt.js test)"
-# - "node node_modules/pxt-core/built/pxt.js testdir tests"
-#  - "(cd libs/hello; node ../../node_modules/pxt-core/built/pxt.js testconv http://az851932.vo.msecnd.net/files/td-converter-tests-v1.json)"
-sudo: false
-notifications:
-  email:
-    - touchdevelop-build@microsoft.com
-cache:
```
shakao, line 22: do we need this caching stuff in github actions also?

riknoll: Hmm, it's supported but i might do it in a separate pr: https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows

I think this is just to speed things up.

jwunderl (approved):

PR #3867, opened 2021-02-04
add pxt-jacdac/button repo to approved list

pelikhan:
Mostly testing that cloud supports nested repoes.


abchatra:
If I search for the repo:
https://github.com/microsoft/pxt-jacdac/button

It doesn't work. I am not why we are allow listing this one.

pelikhan:
Needed for testing jacdac; but i can do without.

PR #3904, opened 2021-03-16
Es2017

pelikhan:
@jwunderl did you start upgrading microbit to es2017? There's quite a bit of code in this target that needs patching.


sim/state/soundexpression.ts
```diff
@@ -46,10 +46,10 @@ namespace pxsim.music {
                     return pp;
             }
             // don't wait
+            cb();
             return Promise.resolve();
         }).catch((e) => {
             console.log(e)
-        }).finally(() => {
```
pelikhan, line 52: this blows up when we try to extract shims from the simulator.

jwunderl (approved):

pelikhan:
magic

sim/state/soundexpression.ts
jwunderl: Ah, this one was working before but I dropped it down to es2017 to keep support for old edge; it's a bit hard to get finally working because it looks like this code gets run through both the ts2.6 compiler for simshims and the ts3.7 compiler for normal use, and the "lib" for finally is different in those two versions (`esnext.promise` vs `es2018.promise`), and you can't include both as it crashes on invalid entries.

If we want finally in sim code I guess can just extend the interface in pxsim? Already polyfilling it there anyways, but this is the only place I've seen it used in the sim so probably just replacing here and moving on is fine too

pelikhan:
I can't see to build locally....

jwunderl:
Hm, `pxt clean` / `rm -rf node_modules` / `rm package-lock.json` / `npm i` / `pxt serve`? It served / ran ci on my machine fine, but these changes seem to get hit by all of the caching issues across our build / npm / ts / etc

pelikhan:
Waiting for new minor bump in pxt.

PR #3932, opened 2021-03-25
Extensions: add DS-2wire

microbit-mark:
cc @bsbsiever

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -160,6 +160,10 @@ Check out [the accessories pages on microbit.org](https://microbit.org/buy/acces
   "url":"/pkg/bsiever/microbit-dstemp",
   "cardType": "package"
 }, {
+ "name": "DS18B20 two wire",
+  "url":"/pkg/bsiever/microbit-dstemp-2wire",
+  "cardType": "package"
+}, {
  "name": "DS3231",
   "url":"/pkg/keble6/pxt-DS3231",
   "cardType": "package"
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -202,7 +202,8 @@
             "Wappsto/pxt-wappsto",
             "KS-Bulme/pxt-mikRobot",
             "CytronTechnologies/pxt-rekabit",
-            "bsiever/microbit-dstemp"
+            "bsiever/microbit-dstemp",
+            "bsiever/microbit-dstemp-2wire"
         ],
         "preferredRepos": [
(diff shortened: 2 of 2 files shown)


abchatra:
What is the difference between  bsiever/microbit-dstemp &  bsiever/microbit-dstemp-2wire extensions?

abchatra:
@bsbsiever - extension code looks good. Looks like both the extensions create same category as well.

bsiever:
@abchatra They are both for the same sensor, the DS18B20 (https://www.maximintegrated.com/en/products/sensors/DS18B20.html), but it can operate in two modes:  

* "Normal" mode requires three wires: power, ground, and signal.  The signal needs a 4.7k Ohm pull-up resistor too.   The bsiever/microbit-dstemp version is for this mode.   
* It can also operate in parasitic power mode, where two of the three wires on the sensor are connected together (the power and ground) and then only two wires connect to the micro:bit: the combined power/ground from the sensor is connected to the micro:bit's ground and the sensor's signal is connected to an I/O pin. This version (bsiever/microbit-dstemp-2wire) supports the two wire mode.  
  * This extension uses the micro:bit's internal pull-up rather than an external pull-up.  This is operating the sensor a little out of specifications (~13k pull-up vs. the 4.7k normally used.  I haven't had any problems in my testing). 
  *  This approach it easier for people to build their own sensor setup (strip and twist together a few wires --- it takes about 5-10 min and about $7 in parts).  It also make connecting a sensor easier (two wires rather than three).  The readme has instructions and pictures that describe building the sensor: https://github.com/bsiever/microbit-dstemp-2wire/blob/master/README.md
  * The downside:  There are a lot of sensors that are listed as DS18B20 but aren't genuine. These knockoffs may not work in this mode, but the other version of the extension *may* work.   The readme also links to a supplier that is an authorized distributor of the sensors in North America.

I hope that helps.  Let me know if you have any questions or thoughts

abchatra:
@bsiever. Thanks for the details. 

I would have distinct categories. It is easier to manage for the user and it is clear which one they have. Is it possible to update your extension?

bsiever:
> I would have distinct categories. It is easier to manage for the user and it is clear which one they have. Is it possible to update your extension?

@abchatra DONE!  It's now "DS Temp 2-Wire" in the blocks palette and a `dstemp2wire` namespace.

abchatra (approved):

Issue #3949, opened 2021-04-05
immersive reader icon colors


pelikhan:
Orange and teal is a bit painful.
![image](https://user-images.githubusercontent.com/4175913/113571576-ef1c7a80-9616-11eb-9411-c715f9025e60.png)

abchatra:
@soniakandah can you please help here.

Issue #2290, opened 2019-07-18
WCAG 1.4.3: Elements must have sufficient color contrast (.landscape.text.only)


jwunderl:
#### Issue
`Elements must have sufficient color contrast` ([color-contrast](https://dequeuniversity.com/rules/axe/3.2/color-contrast?application=msftAI))
#### Target application
[Microsoft MakeCode for micro:bit](https://makecode.microbit.org/#editor)
#### Element path
.tutorialname > .landscape.text.only
#### Snippet
`<span class="ui text landscape only">Flashing Heart</span>`
#### How to fix
    Fix any of the following:
      Element has insufficient color contrast of 2.15 (foreground color: #fff3e9, background color: #ff8b27, font size: 12.0pt, font weight: normal). Expected contrast ratio of 4.5:1
#### Environment
Chrome version 75.0.3770.142
This accessibility issue was found using Accessibility Insights for Web 2.7.1 (axe-core 3.2.2), a tool that helps find and fix accessibility issues. Get more information & download this tool at [http://aka.ms/AccessibilityInsights](http://aka.ms/AccessibilityInsights).

tutorial title header
![Screen Shot 2019-07-18 at 2 35 18 PM](https://user-images.githubusercontent.com/5615930/61493789-4e30ec00-a969-11e9-920a-8e7e7f91c309.png)

PR #3972, opened 2021-04-09
Purple tutorial theme

jwunderl:
As discussed in design office hours, the orange is jarring & is very inaccessible / conflicts with other editor colors. This switches the tutorial colors to be the microbit purple colors as we had talked about there

* fix https://github.com/microsoft/pxt-microbit/issues/3949
* fix https://github.com/microsoft/pxt-microbit/issues/2290
* fix color bunch of colors: tutorial preview in my projects gallery, detail view, tutorial buttons, etc (no issue filed but is currently inaccessible)

build: https://makecode.microbit.org/app/8674fc02997b6643600aa3d7182a320ea5887d60-f82ef95661

image
image
image
image

### DIFF
--- a/theme/style.less
+++ b/theme/style.less
@@ -62,19 +62,20 @@
         background-position: 50% 25%;
     }
 
-    .detailview .actions .card-action .button.approve {
-        background-color: @mainMenuTutorialBackground;
+    .detailview .actions .card-action {
+        .button.approve {
+            background-color: @mainMenuTutorialBackground;
+        }
+
+        &:first-child .button.approve {
+            background-color: @green;
+        }
+
     }
     .tutorial-progress.orange {
         background-color: @mainMenuTutorialBackground !important;
         border-color: @mainMenuTutorialBackground !important;
     }
-}
-
-.hc #homescreen .ui.home .tutorial-progress.orange {
-    border-color: @HCtextColor !important;
-    background-color: @HCbackground !important;
-    color: @HCtextColor !important;
 }
 
 #tutorialcard {
--- a/theme/style.less
+++ b/theme/style.less
@@ -66,11 +66,6 @@
         .button.approve {
             background-color: @mainMenuTutorialBackground;
         }
-
-        &:first-child .button.approve {
-            background-color: @green;
-        }
-
     }
     .tutorial-progress.orange {
         background-color: @mainMenuTutorialBackground !important;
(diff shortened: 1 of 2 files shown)


shakao:
hmm i think the js/python "start tutorial" cards should be a different color still... or maybe "blocks" should be green? i think it's useful to differentiate

jwunderl:
Ah yeah I had it as a darker purple at first but dropped it cause it looked bad, green could work. For what it's worth arcade is all orange currently / could be fixed to match as well

theme/style.less
```diff
@@ -57,13 +57,31 @@
...
+    .detailview .actions .card-action .button.approve {
+        background-color: @mainMenuTutorialBackground;
+    }
+    .tutorial-progress.orange {
+        background-color: @mainMenuTutorialBackground !important;
+        border-color: @mainMenuTutorialBackground !important;
+    }
+}
+
+.hc #homescreen .ui.home .tutorial-progress.orange {
```
shakao, line 74: small: slight preference for keeping hc styles in the highcontrast.less file

shakao:
oh yeah! didn't realize arcade was all the same color--maybe it's fine then!

theme/style.less
jwunderl: Yeah, I had it in here as it's microbit specific override but I'll bump it over there as it seems possible to change in other targets later on too. I'll add it on to https://github.com/microsoft/pxt/pull/8085 

also I checked green first button and I think I like it, eye catching & nice parallel with completed / started tutorials. Adding that in 
image

shakao: lsdkfjf ok now looking at this, i'm vaguely remembering that it used to be green and we recolored it (https://github.com/microsoft/pxt/pull/6908) BUT that was pre-having-a-real-designer so maybe this is worth a quick pass in design office hrs..... ?

Jaqster:
I do think it's strange that Blocks tutorials are a different color...  I would keep all the tutorials the same color.  From a user perspective, there are 3 options - why is one a different color?

theme/style.less
jwunderl: Ah sounds good, I dropped the style from this pr for now / will probably want to walk through all these changes with Sonia real quick before merging anyway just to confirm~

jwunderl:
Feedback from design meeting was just changing the color for the step indicator for contrast / prettiness:

image

added in last commit ~

abchatra:
@jwunderl ready to merge?

abchatra (approved):

abchatra:
Let me merge it for now. We can test and decide tomorrow.

Issue #3253, opened 2020-06-11
The first simulator can't show in 'Hack Your Headphones'


Sheila111:
**Describe the bug**
The first simulator can't show in 'Hack Your Headphones'

**Steps to reproduce the behavior**
1.Open Microbit app
2.Music----->Hack Your Headphones---->Show Instructions
3.Click run

**Expect behavior**
Simulator shows after clicking **run**

**Actual behavior**
The first simulator can't show online & offline
![music](https://user-images.githubusercontent.com/59192069/84374317-660fff00-ac10-11ea-9595-de6dc192b354.gif)

**Additional context**
1.OS: Windows(rs6)
2.makecode.microbit.org version: 3.0.10
3.Microsoft MakeCode version: 6.0.11

Sheila111:
Also repro on the website

PR #3989, opened 2021-04-19
Swap order of first steps in Hack Your Headphones

shakao:
fixes https://github.com/microsoft/pxt-microbit/issues/3253

we have a namespace called "light" so this top level variable can't be named "light". i just swapped the order of the first two steps so we never get that floating variable block.


ganicke (approved):
Could use a variable named `lite` or `luminate`. This is OK though.

shakao:
oh haha yeah i was going strictly by which was the lowest churn but i'm happy to do it the other way if you think that's better! i did wonder if changing the order would have an effect on classrooms/teachers who have taken screenshots, perhaps?

PR #4048, opened 2021-05-11
Adding sound emoji support to the simulator

riknoll:
Requires https://github.com/microsoft/pxt/pull/8140

Adds simulator support for the micro:bit "sound emoji". I translated this code into TypeScript basically line for line from the C++ in codal so as to make it easier to integrate changes in the future. As part of this change, I also removed the old .wav files we were using for the built-in sound emoji since we can parse them accurately now.

The codal source files are linked in the code.

Here's a test build:

https://makecode.microbit.org/app/66ff61a2520f83910eac03e207ad1026591700d3-c0e42ddcdb

...and here's a test project you can import that cycles through the builtin effects when you press A:

https://makecode.microbit.org/_LsTHejd1XAEs

Tested in Chrome, Safari, and Firefox on macOS


sim/state/soundEmojiSynthesizer.ts
```diff
@@ -0,0 +1,220 @@
...
+                            this.status &= ~EMOJI_SYNTHESIZER_STATUS_STOPPING;
+
+                            // Event(id, DEVICE_SOUND_EMOJI_SYNTHESIZER_EVT_DONE);
+                            // lock.notify();
+                        }
+                    }
+                }
+
+                // If we have something to do, ensure our buffers are created.
+                // We defer creation to avoid unecessary heap allocation when genertaing silence.
```
jwunderl, line 131:
```suggestion
                // We defer creation to avoid unnecessary heap allocation when generating silence.
```

jwunderl (approved):

sim/state/soundSynthesizerEffects.ts
```diff
@@ -0,0 +1,135 @@
...
+    export function exponentialRisingInterpolation(synth: SoundEmojiSynthesizer, context: ToneEffect) {
+        synth.frequency = synth.effect.frequency + Math.sin(0.01745329 * context.step) * context.parameter[0];
+    }
+
+    // Exponential falling function
+    export function exponentialFallingInterpolation(synth: SoundEmojiSynthesizer, context: ToneEffect) {
+        synth.frequency = synth.effect.frequency + Math.cos(0.01745329 * context.step) * context.parameter[0];
+    }
+
+    // Argeppio functions
```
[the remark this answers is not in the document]
riknoll: yeah in the code it's "appregrio" which is also misspelled, but I decided to leave it to match codal

sim/state/musicalProgressions.ts
```diff
@@ -0,0 +1,58 @@
+namespace pxsim.music {
+    export interface Progression {
+        interval: number[];
+        length: number;
+    }
+}
+
+namespace pxsim.music.MusicalIntervals {
```
pelikhan, line 8: nit: camerlize namespace?

PR #4112, opened 2021-05-26
Extensions: Add PTKidsBIT

microbit-mark:
cc @iBuilds


abchatra:
This looks good to me. 

One suggestion. If you can change the color of the extension category. It is bit hard to see for folks with contrast issues.

abchatra (approved):

Issue #3933, opened 2021-03-25
bluetooth package should not be the very first package for micro:bit


pelikhan:
We should weight down the bluetooth package so that it's not the very first extension in the dialog.

PR #4121, opened 2021-05-27
lower bluetooth to last of built in exts

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/3933

This drops it down to be the last of the built in extensions (for reference, [default weight=50](https://github.com/microsoft/pxt/blob/e46205c8a3bbc245687d7daccfd79daa6214f247/webapp/src/scriptsearch.tsx#L356)). Also added a flag so that it only shows up after a search has been started, which requires https://github.com/microsoft/pxt/pull/8207 to do anything but is fine to merge before that gets bumped in.

### DIFF
--- a/libs/bluetooth/pxt.json
+++ b/libs/bluetooth/pxt.json
@@ -10,6 +10,7 @@
         "BLEHF2Service.h",
         "BLEHF2Service.cpp"
     ],
+    "weight": 10,
     "icon": "./static/packages/bluetooth/icon.png",
     "public": true,
     "dependencies": {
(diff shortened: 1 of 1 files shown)


abchatra:
Hide this and make it searchable.

jwunderl:
Updated with searchOnly flag from https://github.com/microsoft/pxt/pull/8207, so that it won't show up by default:

image

but will if you search for something and it's a valid response for that:

image

(I left the lowered weight so that it's still the end of those results anyways~)

abchatra (approved):

PR #4124, opened 2021-05-28
Fixed Button Color When Code is Validated

Jxwoon:
See title - made changes from 5/25 Design Office Hours
Note: These colors are for Microbit

Screenshot (32)


jwunderl (approved):

jwunderl (commented):
Ah I reviewed this one before I saw the pxt one; can we actually just set tutorialCodeValidationCorrectBackground and 
tutorialCodeValidationCorrectTextColor instead of having the second 'Microbit' variables? just setting:


```
@tutorialCodeValidationCorrectBackground: #cd0365;
@tutorialCodeValidationCorrectTextColor: @white;
```

and dropping in theme/style.less should work?

theme/style.less
```diff
@@ -80,6 +80,16 @@
             background-color: darken(@mainMenuTutorialBackground, 20%);
         }
     }
+
+    .isValidated {
```
shakao, line 84: is this necessary? it looks duplicated from pxt

shakao (changes requested):
joey's comment re: removing the changes from style.less

theme/site/globals/site.variables
```diff
@@ -84,4 +84,10 @@
...
 @serialTextColor: black;
-@serialGraphBackground: #d9d9d9;
\ No newline at end of file
+@serialGraphBackground: #d9d9d9;
+
+/*---------------------------
+   Tutorial Code Validation
+----------------------------*/
+@tutorialCodeValidationCorrectBackground: #cd0365;
+@tutorialCodeValidationCorrectTextColor: @white;
```
shakao, line 93: this one is also duplicated, i think

shakao (approved):

Issue #3817, opened 2020-12-16
WaitMicros not working in the simulator


microbit-carlos:
**Describe the bug**
When the wait microseconds blocks is used in the simulator (in this case with a value of 1_000_000 for a 1 second delay) it doesn't seem to delay the code.

**To Reproduce**
1. Run code as shown in the GIF
2. The LEDs toggle much quicker than expected

**Expected behavior**
An LED to toggle per second

**Screenshots**
![makecode-wait](https://user-images.githubusercontent.com/29712657/102366143-36232900-3fb0-11eb-9c0e-7e689945d31b.gif)

**micro:bit version (please complete the following information):**
N/A

**Desktop (please complete the following information):**
 - OS: macOS Mojave
 - Browser Chrome
 - Version 87

**Smartphone (please complete the following information):**
N/A

**Additional context**
N/A

pelikhan:
waitMicros should not be used for more than a few hundred micros (or at least less then 6ms)

microbit-carlos:
Is there a technical limitation for that?
It wouldn't be obvious to the user that this would be the case. If we cannot fix it, should the input range of the block be limited?

pelikhan:
waitMicros is busy wait and prevents any other code to run.

________________________________

livcheerful:
I can add a max valid value for the input for the block!

PR #4139, opened 2021-06-04
Add async label for waitMicros

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3817

Wasn't sure what the difference between changing control.cpp vs shims.d.ts, so I added async to both! Let me know what the difference is and if I should remove it lol

### DIFF
--- a/libs/core/control.cpp
+++ b/libs/core/control.cpp
@@ -273,7 +273,7 @@ namespace control {
     * Blocks the current fiber for the given microseconds
     * @param micros number of micro-seconds to wait. eg: 4
     */
-    //% help=control/wait-micros weight=29
+    //% help=control/wait-micros weight=29 async
     //% blockId="control_wait_us" block="wait (µs)%micros"
     void waitMicros(int micros) {
         sleep_us(micros);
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

libs/core/control.cpp
```diff
@@ -273,7 +273,7 @@ namespace control {
     * Blocks the current fiber for the given microseconds
     * @param micros number of micro-seconds to wait. eg: 4
     */
-    //% help=control/wait-micros weight=29
+    //% help=control/wait-micros weight=29 async
```
pelikhan, line 276: wait Micros is blocking.

pelikhan (changes requested):
waitMicros is blocking by design.

pelikhan:
The async label was used in TouchDevelop times; i'm not sure it's relevant anymore. Needs to be checked.

It is misleading to mark waitMicros as async as it is blocking on hardware. common-packages version should be fixed i guess. It does create a discreptancy in the way we simulate the code in the browser but there's not much we can do there.

Why the change?

pelikhan:
To be clear: waitMicros is mean to be used with a **small** amount of micro to do real-time computation (like measuring time between 2 pin pulses). Anything larger than a few hundred microseconds, should use pause which has a minimum delay of 6ms.

waitMicros hogs the MCU, and thus will start fighting the BLE stack and so on.

pelikhan (approved):

Issue #3743, opened 2020-11-20
function to determine which version of micro:bit is present


tballmsft:
I would like to write a program that uses micro:bit V2 play API (Giggle, for example), but still works on V1. So, I need a function to determine which micro:bit is running my program (V1 or V2).

mmoskal:
`control.ramSize()` may be good

abchatra:
Request is for typescript API for microbit version?

abchatra:
@finneyj we are planning to build an API called control.hardwareVersion() for extension authors to have different behavior in v1 vs v2.  Is there way to get the exact hardware version?

Current implementation plan looks like this:

```C++
#CODAL 
  return 2
#else                       
  return 1
```
Exact hardware version would be great to be future proof .

abchatra:
@tballmsft your job to test the API :-)

finneyj:
apologies @abchatra  - I meant to get back to you on this on Monday. It looks like you have this in hand MakeCode side now, but for future reference:

For v2 we have a hardware version baked into the PCB, and interface firmware version which can be extracted from here at runtime: https://github.com/lancaster-university/codal-microbit-v2/blob/master/inc/MicroBitPowerManager.h#L187

We could wrap this in a static function also exposed on v1 if that would help (although we could only get major version number for v1 accurately).

If you need compile time version checking, I'd also be happy to add a #define for you on both v1 and v2.

abchatra:
Thanks @finneyj. I think it is sufficient if there is an API in codal, I don't think specific v1 version is necessary.

@livcheerful can we use this function in #codal case?

finneyj:
no worries @abchatra

@livcheerful FYI - there's already an instance of this class in the MicroBit singleton, so this should get you what you need:

`uBit.power.getVersion();`

I seem to recall the values returned from the hardware are a little cryptic. @carlosperate may be able to comment on whether we have a look up table defined to map them to the version number printed on the silkscreen on the PCB (like v2.00 etc).

martinwork:
Related: https://github.com/lancaster-university/codal-microbit-v2/issues/77#issuecomment-855238467

PR #4144, opened 2021-06-09
API to fetch microbit version 

livcheerful:
fixes https://github.com/microsoft/pxt-microbit/issues/3743

wasn't sure if i should be checkin in shims.d.ts, but it looked like it had recent changes so here it is.

Returning as a number bc that makes the most sense with just v1 and v2. It will be difficult to deal with if we ever do add minor versions too, but right now we don't have a good way to find that out! But lemme know if you think it should be...a string instead?

### DIFF
--- a/libs/core/control.ts
+++ b/libs/core/control.ts
@@ -132,6 +132,11 @@ namespace control {
             t += 0x3fffffff
         return t
     }
+
+    //% shim=control::hardwareVersion
+    export function hardwareVersion(): number {
+        return 2;
+    }
 }
 
 /**
--- a/libs/core/control.ts
+++ b/libs/core/control.ts
@@ -133,7 +133,7 @@ namespace control {
         return t
     }
 
-    //% shim=control::hardwareVersion
+    //% shim=control::_hardwareVersion
     export function hardwareVersion(): number {
         return 2;
     }
--- a/docs/reference/control.md
+++ b/docs/reference/control.md
@@ -4,7 +4,7 @@ Runtime and event utilities.
 
 ```cards
 control.inBackground(() => {
-    
+
 });
 control.reset();
 control.waitMicros(4);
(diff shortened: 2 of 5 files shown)


libs/core/control.ts
```diff
@@ -132,6 +132,11 @@ namespace control {
             t += 0x3fffffff
         return t
     }
+
+    //% shim=control::_hardwareVersion
+    export function hardwareVersion(): number {
+        return 2;
```
pelikhan, line 138: the simulator actually knows if the device is v1 or v2.

pelikhan (approved):

pelikhan:
Yes you want to checkin the regenerated shims.

libs/core/control.ts
livcheerful: oooohhh okay i'll add it

livcheerful: wait like the physical device?

pelikhan:
Yes it infers it from the APIs being used.

libs/core/control.ts
jwunderl: this field: https://github.com/microsoft/pxt-microbit/blob/datalogger/sim/dalboard.ts#L34, since we change the look depending on whether v2 blocks or used or not; currently this will return 2 but the simulator will look like v1 if the user does not use any v2-only features

livcheerful: ? is there a way to access that sim code from /libs?

pelikhan:
Right you would have to move this function into the sim part of the house and implement it under ``pxsim.control.hardwareVersion()``.


Sent from Outlook<http://aka.ms/weboutlook>

livcheerful:
ah okay yeah I see. I tried it out, and it looks like the board only changes to look like "v2"  when it hits the line of code that uses a v2 API. So if you put the hardwareVersion fetch before, it will return 1, and if you put it after, it will return 2.

image
image


I could try to take a look to see if there's a way to check all the code for v2 APIs before running, or I think it might be okay to just return 2 all the time. I think since this API doesn't have a block, it won't be super visible and you might want to test your v2 code anyway with it.

pelikhan:
I guess 2 is the future, so let's always return that.


Sent from Outlook<http://aka.ms/weboutlook>

Issue #4068, opened 2021-05-19
feature request: "on timer" block


eduvik:
**Is your feature request related to a problem? Please describe.**
It would be really useful to have a top-level block that is triggered when the specified amount of time has elapsed. 

**Describe the solution you'd like**
A top level block called "on time elapsed" or "on time expired" with a slot to enter a number of seconds.

**Describe alternatives you've considered**
* use pause to wait a certain time within a forever block
* use running time variable
* use a timer variable

**Additional context**
All of the above alternatives would work, but a "on timer" block would fit the event-driven paradigm that the environment uses, and make it easier for students to get stuff done.

pelikhan:
``every``

eduvik:
> `every`

Sounds good, just as long as a search for "time" or "timer" discovers this as well.

microbit-mark:
Also the default values have been re-used from another dropdown. It might be useful, eg for the datalogging feature in https://github.com/microsoft/pxt-microbit/pull/4146 to change these defaults to longer periods ![87F94D50-1D4F-45D3-BAA6-B72225B74262](https://user-images.githubusercontent.com/31242877/121657332-39e3c400-ca98-11eb-89d8-958f2462398e.jpg)

Issue #4141, opened 2021-06-07
Data Logger Extension Thumbnail


soniakandah:
### **The Problem**

The thumbnail for the data logger extension is not very clear and may not attract our desired users. 

![image](https://user-images.githubusercontent.com/22336391/121089060-70b99180-c79b-11eb-97ab-aa9ccb48cb68.png)

We should create a new asset for this extension. 

@jwunderl  for visibility

microbit-giles:
It's a tricky concept to get across visually, but I agree this image should be updated (it would be good for a thermal printer driver!). Perhaps something with a simple line graph and table of numbers would work?

soniakandah:
Here's my take on the data logging extension thumbnail, let me know what y'all think!
 
![Data_Logger_Ext](https://user-images.githubusercontent.com/22336391/121942873-d232b080-cd05-11eb-989f-696fac15dff7.png)

jwunderl:
This is great, I'll update at same time as https://github.com/microsoft/pxt-microbit/issues/4142 ~

microbit-giles:
Much clearer! Do the curved lines suggest it's a radio feature? I wonder if it would be just as good without them?

soniakandah:
@microbit-giles Ah I meant it to refer to like either using the speaker, shaking the device, etc. But yeah I guess it's not communicated well, should I change them?

soniakandah:
The team talked about this during our design sync and I will be replacing the radio waves with "V2 Only"

soniakandah:
Here's the updated image!

![Data_Logger_Ext 2](https://user-images.githubusercontent.com/22336391/122131460-0e414080-cdee-11eb-8bfd-7236d3eb9041.png)

abchatra:
@jaustin

jaustin:
Thanks for this! I see I was too slow today to respond but we have the 'works with' badges that we'd really like to overlay on the image (either directly, or in MakeCode when showing the tile?)

https://microbit.org/brand-guidelines/

![image](https://user-images.githubusercontent.com/147113/122275156-4b373100-cedb-11eb-8899-13bc48527931.png)

The icon itself looks great!

soniakandah:
@jaustin How is this? 
![Data_Logger_Ext 3](https://user-images.githubusercontent.com/22336391/122284066-15c22180-cea2-11eb-9a18-137daf3f7163.png)

@jwunderl reopening

jaustin:
[...]

PR #4146, opened 2021-06-09
Datalogger extension

jwunderl:
Extension for the new datalogger apis:

image

fix https://github.com/microsoft/pxt-microbit/issues/4068
Along with the requested 'every {blank} ms' forever equivalent requested in association with the datalogger blocks:

image

various notes:
* There's currently a bug I'll look at in pxt where the + button on the array parameter doesn't fill in shadow blocks of the key/value pair object
* updating codal typing dropped a bunch of defines from jacdac; I'm assuming that's expected?
* Right now I set the mirror to serial to output as key: value pairs, which should show up nicely in the serial view as graphs. If we want it to send the exact same lines to serial as the ones that are logged, that should probably go to codal in some form, as otherwise we'll have to double the storage of the headers / logic associated with that and we'd still get timeouts wrong.;
    * 'read last line' function?
    * 'on line end' that runs in endRow and passes the string?
    * Or just a flag we set to enable serial in the first place
* I left the underlying 'flashlog' as it's own separate extension that only shows up under search, so that api authors can use that without requiring them to pull in our extension (ideally avoiding them adding their own cpp to bypass that)
* I asked sonia for a proper icon, tracking that in https://github.com/microsoft/pxt-microbit/issues/4141
* Side note, there was a cpp error in flashlog since we bumped codal versions which renamed the delete log function (format -> clear) -- I was a little surprised by this as it didn't show up as an error in the build, I thought we built the hex for all the bundled extensions which would have thrown on that.
* worth considering circling back to the original api design which had timestamp as an optional parameter on the set headers block and thinking about it again -- right now, if you set the timestamp after setting the headers we'll need to double them still (as it adds a new column for the timestamp)

Build: https://makecode.microbit.org/app/b54cc751bfd6d9c1be98ef61a497e03e6754f4cc-aeef45bcc3
If you have one of the prerelease v2 mbits make sure to update firmware or it will silently fail and not save anything


libs/core/shims.d.ts
```diff
@@ -487,12 +487,6 @@ declare namespace control {
 }
 declare namespace control {
 
-    /**
```
livcheerful, line 490: do we wanna delete this? I kept on getting this deleted too when I was building, wasn't sure

abchatra:
Couple of uber comments: 
- Column value should be an enum with prepopulated values for now and ability to add new values (like PropertyKind in Arcade)
-  Default measurement in set timestamp should be in seconds and not millseconds

abchatra:
Timestamp should be on by default and mirror should be on by default as well.

libs/core/shims.d.ts
jwunderl: It being removed on build means the function doesn't exist / isn't exposed anymore on the c++ side right? so any attempt at using this would crash. If we don't want to delete this then we probably want to re-expose the function but I think checking in is right for now

jwunderl:
Fixed all but the dropdown for columns ^^ if we want that for the initial release it's a new field editor / subclass of the spritekind one -- that only keeps numbers so there's no immediate way to get from 'column ID' -> 'column name' without adding it to the emit -- e.g. instead of 

```
namespace SpriteKind {
    export const my_kind= SpriteKind.create()
}
```

Will need a new version that emits something like

```
namespace SpriteKind {
    export const my_kind= SpriteKind.create("my kind")
}
```

and handling the compilation / decompilation for that.

libs/core/shims.d.ts
livcheerful: hmmm but it's still in common packages here: https://github.com/microsoft/pxt-common-packages/blob/98df68ac19c130c33f0912b035025b890ccbc524/libs/base/control.cpp#L100

maybe it should be moved? adding it to shims.d.ts was a recent change too https://github.com/microsoft/pxt-microbit/pull/4040

jwunderl: I think that's overwritten here without this which is why it's getting dropped? https://github.com/microsoft/pxt-microbit/blob/datalogger/libs/core/control.cpp#L390

livcheerful (approved):

jaustin:
This is looking awesome, we'll give some more feedback next week.
One quick thought now...

>worth considering circling back to the original api design which had timestamp as an optional parameter on the set headers block and thinking about it again -- right now, if you set the timestamp after setting the headers we'll need to double them still (as it adds a new column for the timestamp)

@jwunderl @finneyj was going to modify CODAL so that if you enabled or disabled timestamps after set headers, but before actually logging any data, CODAL would still put timestamp column first. I guess as far as the sim/serial abstraction goes you could only send the headers at the point data is first logged to have similar behaviour?

jwunderl:
That makes sense, yes; since it sounds like we want the serial output to match the csv format, I think that on the hardware side that'd have to be in codal -- if it were on our side, we'd have to duplicate the column matching logic in our c++ completely (as well as more precisely match the timestamp formatting). Right now, the serial logging I have is just `column: data` pairs so that it shows up as the graphs in the console view.

(Matching the behavior you mentioned on the sim side should be easy enough though in the future~)

sim/state/flashlog.ts
```diff
@@ -23,12 +24,19 @@ namespace pxsim.flashlog {
...
-        Runtime.postMessage(<SimulatorSerialMessage>{
-            type: 'serial',
-            data,
-            id: runtime.id,
-            sim: true
-        })
+        logSize += data.length;
+
+        // TODO: maybe grab exact size that will fail
+        if (logSize >= 30000) {
```
pelikhan, line 30: this shoudl be a constant somewhere, use exact size

jwunderl: From https://github.com/lancaster-university/codal-microbit-v2/issues/108#issuecomment-845965279 it looks like it's currently ~120kb but might go down to ~60kb effective size (log end is set here: https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/MicroBitLog.cpp#L210 but it will take a bit of digging to get exact effective size I think) -- for now I'll set it at 60k and see if I can get a more accurate size next week, will be merging soon so we can get some testing in tomorrow / will likely need more updates next week anyways.

finneyj: Hey @jwunderl @pelikhan. The magic number you're looking for here is `121852`

AFAIK, no plans for this to drop to ~60KB. The issue you mention above was really just exploring options.

Issue #4142, opened 2021-06-07
Homepage Banner Images


soniakandah:
### **The Problem**

Our current image carousel on the homepage needs some banner images to attract users. We have five items we need carousel images for: 

- [Microsoft MakeCode for the micro:bit](https://makecode.microbit.org/#tutorial:/projects/flashing-heart) 
- [Intro to the micro:bit](https://microbit.org/get-started/first-steps/introduction/)
- [Behind the MakeCode hardware](https://www.youtube.com/playlist?list=PLMMBk9hE-SeqDYtw9pGNPsQ10V_EGMyGe)
- [Send messages with your micro:bit](https://makecode.microbit.org/#tutorial:/projects/micro-chat)
- [Track the soil moisture of your plants](https://makecode.microbit.org/projects/soil-moisture)

@jwunderl @Jaqster for visibility

soniakandah:
[Microsoft MakeCode for the micro:bit](https://makecode.microbit.org/#tutorial:/projects/flashing-heart)

Inspiration: 
![image](https://user-images.githubusercontent.com/22336391/121093107-71552680-c7a1-11eb-9dc8-74dd0b75ce61.png)

Banner Image: 
![mb-banner-01](https://user-images.githubusercontent.com/22336391/121093189-8467f680-c7a1-11eb-8dc0-2f39a8e17661.png)

soniakandah:
[Intro to the micro:bit](https://microbit.org/get-started/first-steps/introduction/)

Inspiration: 
![image](https://user-images.githubusercontent.com/22336391/121093258-9fd30180-c7a1-11eb-9798-a78899007719.png)

Banner Image: 

![mb-banner-02](https://user-images.githubusercontent.com/22336391/121239734-77551100-c84e-11eb-8552-48c3c37830b7.png)

soniakandah:
[Behind the MakeCode Hardware](https://www.youtube.com/playlist?list=PLMMBk9hE-SeqDYtw9pGNPsQ10V_EGMyGe)

Inspiration: 

![image](https://user-images.githubusercontent.com/22336391/121239968-b7b48f00-c84e-11eb-874c-b50174496987.png)

Banner Image: 

![mb-banner-03](https://user-images.githubusercontent.com/22336391/121240002-bc794300-c84e-11eb-9ab7-4315575557e2.png)

soniakandah:
[...]

PR #4149, opened 2021-06-09
Pull in hero banners

jwunderl:
re: https://github.com/microsoft/pxt-microbit/issues/4142, for now I just duplicated the 'intro to microbit' one for soil moisture project while we're waiting on the final image for that. (I did run the images through a quick compression to cut size from ~1mb each to ~100-200kb each)

hero-banner-imgs

I believe docs/static/hero.png will update the live site as well when it is merged (unless there's special casing for it & it gets pulled), but I don't think that's an issue / banner looks better anyway so more of a sneak peek?


abchatra (approved):

abchatra:
>> I believe docs/static/hero.png will update the live site as well when it is merged (

That might be an issue for foundation. Can you separate out that one into a different check in. We can merge close to the release.

jwunderl:
Okay, swapped the file back and pushed that over to https://github.com/microsoft/pxt-microbit/pull/4151 which can wait till release. Will probably include the soil moisture png in there as well~

PR #4164, opened 2021-06-11
various minor datalogger updates

jwunderl:
* More accurate log full, re: https://github.com/microsoft/pxt-microbit/pull/4146#discussion_r649761702
(the column headers are still not being counted, but I'll leave that for when the mirroring is finished on codal side & we finish matching that in the sim)
* drop the on/off toggle from set timestamp and just make it the dropdown defaulting to 'none' -- it's weird to have the two buttons doing the same thing (setting to 'off' and setting to 'none') and this reads better to me
image
* swap delete log optional parameter from toggle to enum: 
image 
image
* Set error code for log full so it's searchable (I chose 044 as that's the next unlisted one, but I guess we'll have to ask micro:bit if they have a preferred one?)
* updated description, add a group:
image
image


libs/datalogger/datalogger.ts
```diff
@@ -16,18 +23,27 @@ namespace datalogger {
...
+                basic.showLeds(`
+                    # . . . #
+                    # # . # #
+                    . . . . .
+                    . # # # .
+                    # . . . #
+                `);
+                basic.pause(1000);
+                basic.clearScreen();
+                basic.showString("044");
```
abchatra, line 42: I would use makecode error code such as 910 or 928
https://support.microbit.org/support/solutions/articles/19000016969-micro-bit-error-codes

abchatra (approved):

libs/datalogger/datalogger.ts
jwunderl: Sounds good, 928 it is!

Issue #4185, opened 2021-06-16
Edit default values for every block


microbit-mark:
The default values have been re-used from another dropdown. It might be useful, eg for the data logging feature in https://github.com/microsoft/pxt-microbit/pull/4146 to change these defaults to longer periods eg 1 minute/ 1 hour? Adding @microbit-giles for comment.

It would also be useful as stated in the original request to have this block show up in a search for 'time' or 'timer'.

 ![87F94D50-1D4F-45D3-BAA6-B72225B74262](https://user-images.githubusercontent.com/31242877/121657332-39e3c400-ca98-11eb-89d8-958f2462398e.jpg)

_Originally posted by @microbit-mark in https://github.com/microsoft/pxt-microbit/issues/4068#issuecomment-859389624_

microbit-giles:
To facilitate longer data logging experiments a list like:
100ms
200ms
500ms
1 second
1 minute
1 hour
would be useful. At the moment it would be quite hard to explain to a young student how to set it up to log every hour.

PR #4198, opened 2021-06-17
Add longer time intervals for the every interval block

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4185


jwunderl (approved):
hah was just about to put up the same fix, lgtm

abchatra (approved):

PR #4219, opened 2021-06-18
fix datalogger on winrt

jwunderl:
My check on whether textencoder was available was bad, fix that.

fix https://github.com/microsoft/pxt-microbit/issues/4220


aznhassan (approved):
LGTM

abchatra (approved):

Issue #4212, opened 2021-06-18
'Play sound [  ]until done' blocks the program from running indefinitely on iPad


kiki-lee:
**Describe the bug**
When using iPad, adding 'play sound [ ] until done' doesn't play, then stops the program at that point.
![image](https://user-images.githubusercontent.com/1633958/122601138-b42ac000-d025-11eb-8e78-66428db1e542.png)

**Expected behavior**
Sound should play and program should continue when it's finished.

**Tablet (please complete the following information):**
 - Device: [iPad Pro]
 - OS: [ iOS14.6]
 - Browser [Chrome]
 - Version [91.0.4472.80]

jwunderl:
Also repros on mac safari14.0.1

PR #4224, opened 2021-06-21
Fix issue in safari where audio blocks forever

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4212

Requires https://github.com/microsoft/pxt/pull/8256

Build will fail until the PXT PR is merged/bumped


jwunderl (approved):
delaying it just seems like such a weird behavior on the browser side... ¯\\_(ツ)_/¯

livcheerful (approved):

PR #4252, opened 2021-06-25
[stable 4.0] Revert codal to 0.2.25 & drop datalogger/flashlogger

jwunderl:
return dal.d.ts to the state before https://github.com/microsoft/pxt-microbit/pull/4146 until flashing issues get figured out


abchatra:
Keep this alive for next couple of weeks. Just in case we hit further issues. This is stable only change. @riknoll @eanders-MS let me know if you need context here.

jwunderl:
no longer necessary

PR #4268, opened 2021-07-08
Update v2-blow-away.md

kiki-lee:
Added Kiki's edits

### DIFF
--- a/docs/projects/v2-blow-away.md
+++ b/docs/projects/v2-blow-away.md
@@ -18,7 +18,7 @@ A wild Haven has appeared!
 
 ► From the ``||basic:Basic||`` category, find ``||basic:show icon [ ]||`` and add it to your ``||basic:on start||`` container.
 
-► Set it to show a ghost.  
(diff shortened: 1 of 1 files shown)


docs/projects/v2-blow-away.md
```diff
@@ -133,20 +133,20 @@ basic.forever(function () {
...
-        }	
+        }   
     }
 })
 ```
 
 ## 8. Displacing LEDs part 2
 
-► Go back into ``||variables:Variables||`` and drag out another ``||variables:set [row] to [0]||``. Place this one at the **bottom** of your ``||logic:if then||`` statement.
+► Go back into ``||variables:Variables||`` and drag out another ``||variables:set [row] to [0]||``. Place this one below the last one (at **the end**) of your `if then` statement.
```
anzhou5, line 143: @kiki-lee I still like this one as "at **the bottom**" rather than "**the end**" because bottom still indicates being inside the if/then

anzhou5 (approved):
Looks good :) Thanks, Kiki!

Issue #3421, opened 2020-10-13
Radio does not work between board variants without first setting the group


abchatra:
<img width="414" alt="image" src="https://user-images.githubusercontent.com/31242877/95192082-92f0af00-07c9-11eb-87ce-65cbcb0ae6dd.png">

Does not work v2-v1
Works v1-v1
Works v2-v2

Setting the group in on start fixes this, but it should work without having to do this

pelikhan:
The issue here is that the hash of the program is different between nrf51 and nrf52; which leads to a different random radio group. SOlutions are:
- always use group 1
- always use group 2
- always use group ...254
- add a new stable hash regardless of the architecture
@mmoskal

mmoskal:
We could embed the project guid in the generated code, and use that as the group. Then the group wouldn't change as the user modifies program, but is different from other users.

abchatra:
My vote is "always use group 254" and document it. If someone doesn't use a group, it is always 254. Easy to remember. No kid understands the hash logic. 

Most folks set their group anyway. 

@Jaqster @jaustin

abchatra:
>> _Why not always default to group 1? easier to explain..._

There might be bunch of scripts out there with group 1 which might conflict if default group is also 1. Unlikely there are many scripts with group 254. Only for backward compat reasons. Otherwise group  1 is better.

Jaqster:
How would the scripts conflict?  Wouldn't it just work anyways - whether group 1 is set explicitly by the user or default-under-the-covers?  To me, having the default group setting be something obvious and easy to explain is better...
![image](https://user-images.githubusercontent.com/22108954/96016723-f009e780-0dfd-11eb-81ca-892ae0ccf7e7.png)

pelikhan:
The point of this feature was to avoid unexpected packet collisions between all the random projects running on micro:bit in a classroom. At this point, if we are to pick a number, we could just default to ``1`` and **always** start the tutorial with setting up the group number.

Jaqster:
That's okay IMO.  Most times the group setting is optional - kids/teachers expect messages to flow to all micro:bits in the classroom.  I've seen Teachers say "if you want to send a secret message" then they have the students set the group number to something secret that only the sender/receiver know.

abchatra:
Sounds good to me. Lets set to default group to "1"

jaustin:
[...]

PR #4289, opened 2021-07-24
Use 'setGroup' in radio examples

### LINKED ISSUE #4261: Radio examples need a set radio group block for V1/V2 compatibility
As a workaround for #3421, MakeCode tutorials need a set group block 

For example https://makecode.microbit.org/projects/multi-dice would not work between board variants.

An alternative, until the issue is resolved is to note that the group must be set if you are working with both board variants, as described in https://support.microbit.org/support/solutions/articles/19000121980-using-radio-with-micro-bit-v1-and-micro-bit-v2

ganicke:
A default radio group isn't set on v2+ boards. Including `radio.setGroup()` in examples.

Closes #4261

### DIFF
--- a/docs/reference/radio/set-group.md
+++ b/docs/reference/radio/set-group.md
@@ -24,8 +24,7 @@ to talk to each other because they will have the same group ID.
 
 #### Default radio group
 
-The radio group for @boardname@ v2 and later is NOT set by default. To make your program work on
-all versions of the @boardname@, it's best to set a radio group number when your program begins.
+If you haven't set a radio group for the @boardname@, it will use one selected randomly. If you are transmiting data to a @boardname@ that has a different hardware version from the sending @boardname@, it will select a random default group that is not the same as the other @boardname@. To be certain that your program will send or receive data using the same radio group, you will need to first choose and set a radio group for your program if you want it to work between different versions of the @boardname@.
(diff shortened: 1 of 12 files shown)


docs/reference/radio/set-group.md
```diff
@@ -20,6 +20,15 @@ to talk to each other because they will have the same group ID.
 
 * **id**: a [number](/types/number) from ``0`` to ``255``.
 
+### ~ reminder
```
ganicke, line 23: Is this the correct thing to say here? RE: #3421

docs/reference/radio/set-group.md
```diff
@@ -20,6 +20,15 @@ to talk to each other because they will have the same group ID.
...
+
+#### Default radio group
+
+The radio group for @boardname@ v2 and later is NOT set by default. To make your program work on
```
ganicke, line 27: Is this the correct thing to say here? RE: #3421

abchatra: It is best to say this _"Default radio group is random if you don't set it explicitly"_

abchatra: This is true for all versions of microbits.

abchatra (approved):

PR #4293, opened 2021-07-26
Use radio set group in "micro chat" activity

### LINKED ISSUE #4290: micro-chat needs to setup the group on the first step
The micro-chat tutorials takes a string input; and it is highly probably that kids enter different strings in their code. This leads to a different group number for the radio and eventually radio does not work.

My suggestion is to move the "set group" step to the start of the tutorial.

ganicke:
Move the radio "Set Group" to the beginning of the tutorial to avoid message collisions when multiple micro:bits are present during the same activity session.

Fixes #4290

### DIFF
--- a/docs/projects/micro-chat.md
+++ b/docs/projects/micro-chat.md
@@ -8,7 +8,7 @@ Use the **radio** to send and receive messages with other @boardname@.
 
 ## Set a radio group
 
-First thing you want to do is to put in a ``||radio:set group||``. This will make certain that you and another @boardname@ will receive just the messages assigned to that group instead messages from some other @boardname@ that's not part of your conversation. This like tuning to a particular channel to talk on.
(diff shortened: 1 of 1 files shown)


docs/projects/micro-chat.md
```diff
@@ -6,6 +6,14 @@
 
 Use the **radio** to send and receive messages with other @boardname@.
 
+## Set a radio group
+
+The first thing you want to do is to put in a ``||radio:set group||``. This will make certain that you and another @boardname@ will receive just the messages assigned to that group instead messages from some other @boardname@ that's not part of your conversation. This like tuning to a particular channel to talk on.
+
+```blocks
+radio.setGroup(123)
```
pelikhan, line 14: Leaving 1 is probably good enough (default value in the bloc)

ganicke: OK, "1" it is.

pelikhan (approved):

Issue #4300, opened 2021-07-29
Some homepage carousels broken in non-english locales


jwunderl:
**Describe the bug**

it looks like the translated versions of homepage galleries broke after microsoft/pxt-microbit#4288 (this is an issue on either our upload or crowdin side, not in the pr)

![image](https://user-images.githubusercontent.com/5615930/127445605-b3eed39c-15e8-426d-8fac-753f310523b3.png)

In some of the galleries, it's dropping the `---` separators and instead marking the last lines as headers; e.g. https://crowdin.com/translate/kindscript/5124/en-fr?filter=basic&value=3

![image](https://user-images.githubusercontent.com/5615930/127418402-874e4ab5-e8e2-4bb1-9789-fffcf44c7d38.png)

where others are fine https://crowdin.com/translate/kindscript/10292/en-fr?filter=basic&value=3

![image](https://user-images.githubusercontent.com/5615930/127445832-a2868a11-2562-4935-9f66-3d76c04abdc5.png)

I'm gonna put up some a few whitespace only changes in the files currently broken on crowdin to see if it will force through a reupload and fix it, a few of these will be broken for 24hrs at least though.

**To Reproduce**
1. Go to microbit homepage in french
2. see error

abchatra:
This is fixed.

PR #4288, opened 2021-07-24
Markdown codecard conversions

ganicke:
Convert all ```` ```codecard```` sections from the old JSON format to the newer text, name-value form.

@pelikhan - the auto-create for `./docs/projects.md` needs an update to use the new codecard format.

Closes #4036

PR #4304, opened 2021-07-29
Temp revert code card conversion

jwunderl:
re: https://github.com/microsoft/pxt-microbit/issues/4300, reverts

* https://github.com/microsoft/pxt-microbit/pull/4301
* https://github.com/microsoft/pxt-microbit/pull/4298
* https://github.com/microsoft/pxt-microbit/pull/4288

After this is in we can try to do it piecemeal and identify why it's currently breaking

### DIFF
--- a/docs/behind-the-makecode-hardware.md
+++ b/docs/behind-the-makecode-hardware.md
@@ -2,7 +2,6 @@
 
 * [Watch Playlist on Youtube](https://www.youtube.com/watch?v=qqBmvHD5bCw&list=PLMMBk9hE-SeqDYtw9pGNPsQ10V_EGMyGe)
 
-
 ## Videos
 
 ### ~ codecard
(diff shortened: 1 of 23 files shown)


abchatra (approved):

ganicke:
Do we want to try this conversion again but with just a few pages at a time?

jwunderl:
If you want to try 1 gallery at a time we can try it, it's still not clear / I'm checking with their side but it sounds like it's possible it will fix itself if we re upload. It's easy to verify once the build goes through whether it will work or not (see screenshots from https://github.com/microsoft/pxt-microbit/issues/4300) so I'm happy to review if you want to go through it, otherwise I'll find some time to get them updated eventually

ganicke:
This might require a manual review and properly applying the translations in the right places after the upload. Initially, we could do the gallery related pages and perform a manual pass to apply the proper translations?

pelikhan:
I'd try with one gallery. Possible a copy of an existing one to test things out; then pick a gallery way down the bottom of the home screen.

ganicke:
Turtle will be our home screen test victim.

PR #4313, opened 2021-08-03
Codecard conversion for 'Turtle' gallery

ganicke:
Convert the 'Turtle' gallery from JSON to the new codecard format. This is an initial test to see how the gallery item strings are uploaded to Crowdin. 

I can massage the translations in Crowdin if they need help. This will have low impact on home screen appearance if it doesn't work out well initially.

RE: https://github.com/microsoft/pxt-microbit/pull/4304#issuecomment-891768416

### DIFF
--- a/docs/projects/turtle.md
+++ b/docs/projects/turtle.md
@@ -4,24 +4,25 @@ Turtle graphics on your @boardname@ screen!
 
 ## Turtle graphics
 
-```codecard
-[{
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
Ah, it's worth noting that it doesn't appear to be the translations themselves that are wrong; when it gets messed up it's the parsing of the source file doing wonky things

PR #4338, opened 2021-08-19
Allow ml4f (machine learning) editor extension

mmoskal:
Extension https://github.com/microsoft/pxt-ml allows for running Tensor Flow machine learning models on cortex-m4f+ (including m:b v2) using my compiler from https://github.com/microsoft/ml4f

The UI for importing models is right now quite basic https://microsoft.github.io/ml4f/

This is mostly an experiment for now - the user has to search for "ml4f" extension, and then click "Import Model" button in "ML" namespace.

### DIFF
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -13,7 +13,7 @@
             "microsoft/pxt-ws2812b",
             "microsoft/pxt-apa102",
             "microsoft/pxt-radio-firefly",
-            "microsoft/pxt-ml4f",
+            "microsoft/pxt-ml/ml4f",
             "KitronikLtd/pxt-kitronik-servo-lite",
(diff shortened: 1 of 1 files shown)


abchatra:
Looks good to me. Minor nit: needs icon.png. 

Also I see only API in the toolbox. Is this what you expect? 
image

abchatra (approved):

mmoskal:
You get other APIs after you import a model. It should in fact be initially empty, but then the "Import Model" button doesn't show... 

Attached is sample model (unzip before dropping on the extension window)

[model.zip](https://github.com/microsoft/pxt-microbit/files/7018190/model.zip)

mmoskal:
Sorry, didn't look at screenshot. The "Import Model" button will appear after targetconfig.json syncs.

PR #4184, opened 2021-06-16
Add feature flag to show Windows App deprecation notice

livcheerful:
Also add the doc page to microbit.

partially fixes https://github.com/microsoft/pxt-microbit/issues/4165

PR #4345, opened 2021-08-25
Extend the 'windows-app' page from common-docs

### LINKED ISSUE #4159: Add information to windows app deprecation help page
We have to deprecate the windows app because Monaco no longer supports legacy Edge (and the windows app can only use legacy edge :( ). I made a help page for users to go to from the app in the deprecation banner if they want more info.

I created a preliminary page here https://github.com/microsoft/pxt/blob/master/docs/windows-app.md but wasn't sure what level of information would be good enough.

I think we need to explain how they can export their projects, maybe point them to the Electron app? And maybe talk about the differences between the Windows app and the website? I think the only one is serial logging/file system access?

ganicke:
Change the flat version of `windows-app.md` to extend the base version from `pxt/common-docs`.

**NOTE**: not sure if this requires a bump of pxt to include the updated page from `pxt/common-docs`. I had to `npm link ..\pxt` to get the extend to work.

Closes #4159
RE: #4184, https://github.com/microsoft/pxt/pull/8434


abchatra:
@ganicke any update here?

ganicke:
This depends on pxt 7.2.6 or better.

PR #4365, opened 2021-09-09
Extension: add micro:bit power saving

microbit-mark:
Adding this for visibility. 

Could you review code in  https://github.com/microbit-foundation/pxt-microbit-v2-power/ to see if there are any changes to make before we approve @abchatra

Also for consideration of @soniakandah. We'd love a better icon than https://github.com/microbit-foundation/pxt-microbit-v2-power/blob/master/icon.png that shows that the extension can be used to save battery power / sleep and wake the micro:bit, if you're able to take a look.


microbit-mark:
> What the reason to call the blocks as _"full power"_ or _"low power"_ ? I am out of loop here.

After feedback, we changed this as technically it's not power on/off as we are just going into a low power/deep sleep state. The micro:bit is still consuming power. Both `power on` and `power up` are problematic. eg `power on on button A` is confusing, `power up` is more familiar to kids in a gaming context and `power up on` won't translate well.

`full power` and `low power` conveys the state more explicitly.

microbit-mark:
@abchatra can we merge this one this week? We can then start communicating to beta testers

microbit-mark:
@abchatra ANother nudge on this one. I don't think it's dependent on graphic design work, but would be a nice to have. I can create a separate ticket for it.

abchatra:
This looks good and I am merging. 

I was looking at the typescript API and would suggest following names for the API which seems more appropriate on what they are doing. No strong opinion though.

lowPowerEnable -> lowPowerMode
lowPowerPause -> lowPowerDuration

abchatra (approved):

microbit-mark:
@martinwork for visibility. Does this change make sense for you? If so, I'll raise a ticket to implement on the extension repo.

> I was looking at the typescript API and would suggest following names for the API which seems more appropriate on what they are doing. No strong opinion though.
> 
> lowPowerEnable -> lowPowerMode 
>lowPowerPause -> lowPowerDuration

martinwork:
> lowPowerEnable -> lowPowerMode

I used "Mode" at one point, but changed it back because "Mode" felt too generic, and I didn't see any chance the parameter would be more than essentially a boolean.

If not thinking about the block popup menu, my favourite might be lowPowerPrevent( bool) or lowPowerVeto( bool), but lowPowerPrevent(Prevent/Allow) doesn't seem right.

"Enable" is used in the CODAL API, and is a very familiar term that gets close to the right idea.

I could not think of words to convey what's really going on: the prevents and allows increase and decrease a counter, so low power is prevented when there have been more prevents than allows.

> lowPowerPause -> lowPowerDuration

lowPowerPause was intended to hint that it's related to basic.pause
lowPowerDuration sounds to me like it's setting an attribute for other APIs to use.
"Duration" isn't a verb but "Pause" can be.

Issue #3962, opened 2021-04-07
Feature Request: Creation of drop down icons/images for experimental sound features


microbit-pauline:
**Is your feature request related to a problem? Please describe.**
We would like to expand our sound features by giving the user the ability to create sounds. This is experimental and we would like to refine the sound creation work with additional wave and curve type features.

**Describe the solution you'd like**
As per https://github.com/microsoft/pxt-microbit/pull/803, we propose having a drop down menu for a possible new sound creation block. 

We propose adding a **Wave Type** and **Curve Type** feature to sound

**Wave Type**
We have created a mock prototype block for Wave Type (which we can provide to you) as a drop down in words. It looks like:

![image](https://user-images.githubusercontent.com/29146842/113869919-2f136700-97a9-11eb-8169-0f0b8a7fa388.png)

We would like to propose changing the words to icons or images such as:

![image](https://user-images.githubusercontent.com/29146842/113870088-5f5b0580-97a9-11eb-8d70-309b935b2832.png)

**Curve Type**
We have created a mock prototype block for Curve Type (which we can provide to you) as a drop down in words. It looks like:

![image](https://user-images.githubusercontent.com/29146842/113881643-8539d780-97b4-11eb-9c7f-fb007c899421.png)

We would like to propose changing the words to icons or images such as:

Standard Curve
![image](https://user-images.githubusercontent.com/29146842/113881759-a26ea600-97b4-11eb-9d49-3113fd9f7538.png)

Linear Curve
![image](https://user-images.githubusercontent.com/29146842/113881727-98e53e00-97b4-11eb-98cb-0a5a5e02b615.png)

Exponential Rising Curve
![image](https://user-images.githubusercontent.com/29146842/113881821-b0bcc200-97b4-11eb-827c-951fd1203f15.png)

**Describe alternatives you've considered**
A drop down with words has been created and a request made by the Education team to replace words with icons

**Additional context**
This is part of the project on creating new sounds

pelikhan:
@abchatra @soniakandah  could we get icons for those and use a similar widget as for the accelerometer events?

microbit-pauline:
This ties in with the discussions on https://github.com/microsoft/pxt-microbit/issues/4049

riknoll:
In beta now

Issue #4049, opened 2021-05-11
Feature request: a way to select preset values for fields in a block all at once (from a dropdown)


jaustin:
We're working on some experimental blocks, and they can have lots of parameters.

It's really helpful for students if we provide examples of how things behave with different 'presets' for all the parameters. For example, showing them some known-good values that demonstrate what the block can do. A bit like the 'gallery' of the sequencer in `play melody`...
<img width="435" alt="image" src="https://user-images.githubusercontent.com/147113/117857479-bbbbb400-b284-11eb-921c-3c9a4338f807.png">

For example, with the SoundEmojiSynth you can get very different sounds by tweaking the parameters, and it can be overwhelming to start from nothing. However, we know it's still important for the parameters for these blocks to be modified as part of the block, and for it to be possible to put variables or expressions in the fields (eg in a physics lesson where you are deriving frequencies). We don't want to ONLY allow a small number of sounds, we want creativity to be available and encouraged. The idea of the presets is as a starting point for this creativity and demonstartion of the range of capabilities, without adding lots of very similar blocks.

Therefore, we would like a way to add a dropdown to a block that populates all fields of that block with some predefined values.

Here's a mockup of what we're proposing:
<img width="874" alt="image" src="https://user-images.githubusercontent.com/147113/117856635-c4f85100-b283-11eb-9aeb-eb226505255b.png">
In this example, clicking on one of the items in the drop down (eg 'squeak') would change the start, end, duration fields to the preset value for a squeak. It would also change the fields NOT shown yet on that block....
(for example, when fully expanded, this block has many parameters)
<img width="745" alt="image" src="https://user-images.githubusercontent.com/147113/117856813-ff61ee00-b283-11eb-9755-a0c549629ac3.png">

This is a step towards a full field editor for these kinds of block, however even without the full field editor, the gallery is really valuable.

<img width="1119" alt="image" src="https://user-images.githubusercontent.com/147113/117857753-089f8a80-b285-11eb-80ef-a83f8952eb9d.png">

We also looked at an alternative treatment where the presets are more 'generic' by being at the beginning of the block - IE perhaps a feature available to other blocks in the future.
<img width="568" alt="image" src="https://user-images.githubusercontent.com/147113/117856912-1bfe2600-b284-11eb-9c0a-3cbb8da7a513.png">

Some discussion we've had while thinking about this
* we have chosen not to show the gallery as part of editing any individual field as per the sketch because
[...]

PR #4376, opened 2021-09-21
Sound editor

microbit-josephine:
These changes have been made in conjunction with https://github.com/microsoft/pxt/pull/8504.  The introduction of the new `play sound` block provides the access point of the sound drop down field in pxt. This was work done for the https://github.com/microbit-foundation/v2/issues/864 issue. 

Most of these changes in pxt-microbit were done in music.ts to introduce the new blocks in the `sound (V2)` category (as requested in the issue https://github.com/microbit-foundation/v2/issues/829). The block also includes drop down fields for `wave types` and `interpolation` which contain png images, which fulfil the requirements of the issue https://github.com/microsoft/pxt-microbit/issues/3962, which requests drop down images. These images were simply drawn by me and converted into png images on-line. They are rather large images, and so may need to be resized.  

The rest of the changes were made in blockly.less for styling. This overrides the styling in semantic.css in pxt, as I found I couldn't render changes directly from there (probably just needed to be re built each time, but I found for the sake of testing etc it was quicker and easier to put changes in pxt-microbit). This was therefore a temporary fix I made, and so these styling changes should ultimately be put  in pxt (I think..).


abchatra:
Please sign the CLA

microbit-mark:
@microbit-josephine, can you provide descriptions for this and https://github.com/microsoft/pxt/pull/8504 so that someone who doesn't know your code and the context of the PR can easily pick it up and work with it? 

I expect this PR would reference 
https://github.com/microsoft/pxt-microbit/issues/4049 and
https://github.com/microsoft/pxt-microbit/issues/3962

microbit-mark:
@abchatra @riknoll what are the next steps here so that we can get a an editor URL for the work here on pxt/pxt-microbit?

abchatra:
@microbit-mark there are dependency issues with the PR in the pxt repo. Those needs to be addressed before we can produce the build. Following needs to be addressed in the pxt.

_The dependencies this change adds are pretty large -- this one alone is 81.9mb unzipped https://www.npmjs.com/package/plotly.js -- this is notably larger than all the other dependencies we have in pxt combined. I'm not certain the backend would handle things properly / it could cause issues if we tried to create a build with this in it_

abchatra:
Who is going to address these issues?

microbit-mark:
> Who is going to address these issues?

@abchatra I wonder if these chart libraries might be related to experiments and not actually used. Josephine has finished her contract with us, but I'll follow up on email and let you know.

PR #4419, opened 2021-11-02
Extension: add Yahboom tinybit

microbit-mark:
cc @YahboomTechnology

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -461,6 +461,10 @@ Check out [the accessories pages on microbit.org](https://microbit.org/buy/acces
 
 ```codecard
 [{
+  "name": "Yahboom Tiny:bit",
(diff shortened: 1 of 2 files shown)


microbit-mark:
This extension has a duplicate function for each board version 
Screenshot2021_11_02_125223

Is this the best approach @abchatra ? I wonder if they should put this in a ‘micro:bit V2’ category at least.

Or could they even detect the board version as part of the function using `control.hardwareVersion()` and only have one block?

abchatra:
This is not the right way. Please make one block and use "control.hardwareVersion()" api inside the implementation to change the behavior if it is less than 2 or great greater than equal to 2. This returns a string "1.X" for anything less than 2

abchatra:
Closing old pull requests.

Issue #4292, opened 2021-07-26
Simulator crashes with native extension


philipphenkel:
**Describe the bug**
The latest public editor fails to integrate the simple native library 
 https://github.com/philipphenkel/pxt-buttons with `Simulator crashed, no error handler TypeError: Cannot read property 'disable' of undefined.` Compilation works as soon as you do whitespace changes to the code. See animated gif.

**To Reproduce**
1. Go to https://makecode.microbit.org/_JwdhzeCu27Pw
2. Open the code editor
3. Compiles with error "Cannot read property 'disable' of undefined"

**Workaround**
4. add a blank line to get rid of the compilation error

**Expected behavior**
Compiles without error form the beginning.

**Screenshots**
![Jul-26-2021 19-59-50](https://user-images.githubusercontent.com/828416/127036487-57695def-2199-42a2-855f-e136d0894b2c.gif)

**micro:bit version (please complete the following information):**
code runs both on v1 and v2. buttons.disable() disable the microbit buttons on v2 devices.

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser: chrome
 - Version 91.0.4472.114

**Additional context**

abchatra:
This is a runtime exception in the simulator.

We don't have support for C++ extensions in the simulator (not possible to run C++ code in the browser as we have javascript VM).  Exception goes away when you add new line because we clear on every edit of the file. If you run the simulator issue exception will reappear.

We can't address this issue.

philipphenkel:
@abchatra I have other extensions with C++ code that work well, e.g. https://github.com/philipphenkel/pxt-serial-read. Did MakeCode drop the support recently? I hope not.

To more specific about the problem here: I don't expect the browser to run C++ code (the C++ part works well on the microbit). Instead, I added a shim which does nothing (empty function). Saying that, I could try adding some useless code to the shim. Maybe the problem is, that the browser removes empty code blocks.

abchatra:
Oh yes, that should work. I am reopening this. However if you find the reason let me know. pxt-serial-read seems to work fine.

philipphenkel:
Once again, another sleep fixed the problem ;-) 

Seriously, the workaround is to add some code to the shim. The shim below is working. If basic.pause is removed and the function turned into a Nulloperation, you will get the simulator crash.

```
    /**
    * Enable the micro:bit buttons. 
    */
    //% shim=buttons::enable
    export function enable(): void { basic.pause(100) }
```

[...]

PR #4430, opened 2021-11-19
Update muvision to v1.2.25

tianli:
Need to fix some typescript bugs. Previous version does not compile and produce errors on makecode.


tianli:
Related to this issue: https://github.com/microsoft/pxt-microbit/issues/4292

abchatra (approved):

PR #4441, opened 2021-12-06
Improve `everyInterval()`

Amerlander:
The function `everyInterval()` was used to run all user's code in parallel. Since this doesn't work, when the execution time is bigger than the interval, it was changed to run the user's code and then pauses as long as the set interval. 

If the interval is 10 seconds and the execution of the code takes 5 seconds, the real interval between each code execution increases to 15s. If the user's code is 20 seconds, it's increasing to 30s. 

I changed this to measure the execution time of the user's code and only apply a pause if there is time left. For a chosen interval of 10 seconds and a script execution time of 5 this will end up to pause 5 additional seconds and the interval will be 10s as expected. If the execution time is 20 seconds it will not pause at all and the interval will be 20s since it just can't be shorter without causing issues.


riknoll (approved):
@abchatra FYI this is on stable4.0

Amerlander:
Oh yes, that was my mistake, because I was reverence on stable4.0 while working on the calliope target.

Shall I just change the target for this PR to the main/master branch?

abchatra:
@Amerlander yes please. Thanks so much for the fixes. Really appreciate it.

Amerlander:
Its changed now. But https://github.com/microsoft/pxt-microbit/pull/4442 was also into stable4.0... do you cherry pick this or shall we revert and create a new PR?

abchatra:
@Amerlander I cherry-picked #4442 to master. So, no need to anything there. 

However, this PR has lot of changes which is being ported to master. Is it possible to just port your change to master and update this PR?

Amerlander:
So, after having some trouble with the new line at eof it is now targeting the master branch and only contains my changes.
I also removed the `let now`, since it's not needed there.

The everyInterval-function in master is different to the one in stable4.0:

**Master**: Code runs every `interval` ms, but the MicroBit will run out of memory when `a()` takes more time than `interval` is set. E.g. scrolling "Hi!" on the display every second.
```
while (true) {
    control.runInParallel(a);
    pause(interval);
}
 ```
 
**Stable4.0**: After `a()` is finished it pauses for `interval` ms before `a()` is executed again. Will lead to longer distance between each execution of `a()` than the set `interval`.
 ```
while (true) {
    a();
    pause(interval);
}
 ```
 
**This PR**: Executes `a()` every `interval` ms, but will not reexucte `a()` while last execution is running if `a()` takes more time than `interval`. This prevents the MicroBit from running out of memory.
 ```
let start = 0;
while (true) {
    start = control.millis();
    a();
    pause(Math.max(0, interval - (control.millis() - start)));
}
```

jwunderl (approved):

pelikhan:
@riknoll @jwunderl this change looks good but maybe it belongs to pxt-common

PR #4442, opened 2021-12-07
Rotate compas counterclockwise

Amerlander:
The Compass in the simulator should increase its value counterclockwise instead of clockwise. When North is heading to the right side, it shows a value of 90 degrees. But when North is right it means the MicroBit is rotated by -90 degrees to the left and therefore the heading measures 270 degrees.


abchatra:
Thanks @Amerlander for the fix.

jwunderl (approved):

abchatra:
@Amerlander I cherry-picked #4442 to master. So, no need to anything there. 

However, this PR has lot of changes which is being ported to master. Is it possible to just port your change to master and update this PR?

PR #4461, opened 2022-01-28
Typo: directiion > direction

microbit-mark:


abchatra (approved):
Looks good. Changes in the pins.cpp is sufficient as the other one is generated file. However I don't think it harms.

Issue #3868, opened 2021-02-04
simple program out of memory error (possible music issue?)


DaveAtKitronik:
A customer reported an error using the Kitronik smart greenhouse. His system gives compilation error.
If he removed the sensing blocks (the board has a BME280 on it) then the error went but the code was not functional. 
we could not reproduce so asked him to send the actual code. 
The Java representation is:

input.onButtonPressed(Button.A, function () {
    music.startMelody(music.builtInMelody(Melodies.PowerUp), MelodyOptions.Once)
    kitronik_smart_greenhouse.controlHighPowerPin(kitronik_smart_greenhouse.HighPowerPins.pin13, kitronik_smart_greenhouse.onOff(true))
    basic.pause(100)
    kitronik_smart_greenhouse.controlHighPowerPin(kitronik_smart_greenhouse.HighPowerPins.pin13, kitronik_smart_greenhouse.onOff(false))
})
let zipLEDs = kitronik_smart_greenhouse.createGreenhouseZIPDisplay(3)
kitronik_smart_greenhouse.setBuzzerPin()
kitronik_smart_greenhouse.setTime(0, 0, 0)
zipLEDs = kitronik_smart_greenhouse.createGreenhouseZIPDisplay(8)
let zipStick = zipLEDs.zipStickRange()
basic.forever(function () {
    zipLEDs.setZipLedColor(0, kitronik_smart_greenhouse.colors(ZipLedColors.Red))
    zipLEDs.setZipLedColor(1, kitronik_smart_greenhouse.colors(ZipLedColors.Green))
    zipLEDs.setZipLedColor(2, kitronik_smart_greenhouse.colors(ZipLedColors.Blue))
    zipLEDs.show()
    basic.showString(kitronik_smart_greenhouse.readTime())
    if (kitronik_smart_greenhouse.readIOPin(kitronik_smart_greenhouse.PinType.analog, kitronik_smart_greenhouse.IOPins.p0) > 500) {
        basic.showIcon(IconNames.Happy)
    } else {
        basic.showIcon(IconNames.Sad)
    }
    basic.pause(1000)
    basic.clearScreen()
    basic.showString("P:")
    basic.showNumber(kitronik_smart_greenhouse.pressure(PressureUnitList.mBar))
    basic.pause(500)
    basic.clearScreen()
    basic.showString("T:")
    basic.showNumber(kitronik_smart_greenhouse.temperature(TemperatureUnitList.C))
    basic.pause(500)
[...]

PR #4481, opened 2022-03-02
Convert builtin melodies to hex buffers and simplify melody code a bit

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3868

This PR attempts to improve code size a bit by converting the builtin melodies from arrays of strings to hex buffers. This reduces the size of binary.js by about 10kb. Unfortunately, that comes at the cost of a bit of memory because the strings are now built at runtime instead of being compiled into the binary.

While I was looking at this code, I also simplified things a little bit in music.ts

The output of getMelody() is very slightly different than before because it will now always output sharps instead of flats. Some of the prebuilt melodies were inconsistent in which ones they used so I standardized them. This is purely a notation difference; the notes themselves are unchanged.

Overall I think this is very much a bandaid over the real problem which is how we defined this API back in the day. Arrays of strings are always going to be pretty inefficient code-gen wise. Instead of merging this PR, we could consider redefining this API.


libs/core/music.ts
```diff
@@ -566,4 +565,48 @@ namespace music {
...
+        let current = "";
+
+        // The buffer format is 2 bytes per note. First note byte is midi
+        // note number, second byte is duration in quarter beats. The note
+        // number 0 is reserved for rests
+        for (let i = 0; i < melody.length; i += 2) {
+            let octave = 4;
+            const note = melody[i] % 12;
+            if (melody[i] === 0) {
+                current="r"
```
jwunderl, line 586:
```suggestion
                current = "r"
```

nit

jwunderl (approved):

Issue #3322, opened 2020-07-18
Access to music.ts/freqs table for user programs


whaleygeek:
**Is your feature request related to a problem? Please describe.**
I wish to translate a note index into a frequency. The note index is an integer. I don't want to have to map those integers to the NoteEnum in order to get a frequency. I need this because I want to turn a mapped accelerometer reading into a tuned playable musical note (not a range of untuned pitches).

**Describe the solution you'd like**
There is a freqs hex table inside music.ts and a lookup feature, which are not exported, they are used directly by the playNextNote(Melody) function. Please can you refactor this code to expose a javascript function which looks up from a keyNumber and returns an integer frequency. User programs can then also perform this mapping without having to duplicate the hidden table.

**Describe alternatives you've considered**
I have created my own function based on your code and table, but it would be more memory efficient and more elegant to just call a function that you expose.

```
function getfreq (n: number, octave: number) {
    const freqs = hex`
        1f00210023002500270029002c002e003100340037003a003e004100450049004e00520057005c00620068006e00
        75007b0083008b0093009c00a500af00b900c400d000dc00e900f70006011501260137014a015d01720188019f01
        b801d201ee010b022a024b026e029302ba02e40210033f037003a403dc03170455049704dd0427057505c8052006
        7d06e0064907b8072d08a9082d09b9094d0aea0a900b400cfa0cc00d910e6f0f5a1053115b1272139a14d4152017
        8018f519801b231dde1e`;
return freqs.getNumber(NumberFormat.UInt16LE, (n * 2) + (octave*12)) || 0
}
```

**Additional context**

PR #4486, opened 2022-03-09
Provide a TS API for accessing the frequency lookup table

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3322


riknoll:
Note this is 1-indexed, but that's because our lookup table doesn't include the 0th octave. No one really wants that octave anyhow

jwunderl (approved):

Issue #4485, opened 2022-03-08
Add Usabilla tag


Jaqster:
Request from the Micro:bit Foundation to do a popup survey question.
In order to support we need to add the usabilla script tag.
I created a Micro:bit button in the Usabilla dashboard - just need to copy the script snippet from there.

microbit-emma-mumford:
Copying @microbit-matt-hillsdon

PR #4488, opened 2022-03-10
Add usabilla feedback button

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4485

Screen Shot 2022-03-10 at 2 21 44 PM


abchatra:
Does this appear in the editor or just the home screen? Home screen is sufficient in my opinion.  
Does this hide the scroll next icon in the projects row? 

If you have the upload target share it.

riknoll:
https://makecode.microbit.org/app/b2e87cb3a794f108858073628556c699443898d6-80527f5252

abchatra:
Probably change the default one to something which doesn't ask for email: 
image

riknoll:
I'll leave that to @Jaqster

microbit-emma-mumford:
Thank you v much! Would it be possible to shuffle the order of the options to have none of the above at the bottom?

microbit-roz:
Hi, Just to add to Emma's comment above. This would be the ideal order for the options ...
- Teacher
- Other type of educator
- Student
- Parent
- None of the above

Jaqster:
Usually the 'randomize order' option is better to ensure people aren't just clicking the first option by default...

microbit-roz:
Also if the user goes directly to the editor and bypasses the home page will they be shown the pop up survey there? That was what we were hoping for.

We have the text translated into different languages if that is of help at all.

... thanks for your work on this btw!

microbit-roz:
> Usually the 'randomize order' option is better to ensure people aren't just clicking the first option by default...

I see what you mean. But the 'Other type of educator' and 'None of the above' don't really make sense if they're presented in a random order. When we've run this survey elsewhere we've not found that the first option is the one that is actually most selected.

microbit-roz:
> Let's just keep on the Home page for now, if we need additional responses we can look to expand to other pages, but we like to avoid distractions in the editor.

Ok. Thanks for the update. When do you anticipate the survey going live? Our suggestion was that it just runs for a week.

riknoll:
Okay, long time coming but I have finally updated this PR so that the usabilla survey is only visible on the home screen. Here is the new test build: https://makecode.microbit.org/app/078e4c4290a081c58467c77dfcb1c84105f74384-317ebf12fa

@abchatra @Jaqster

riknoll:
@abchatra @Jaqster could you try that build and let me know if this is okay to merge?

abchatra:
Survey didn't popup for me in that build

riknoll:
try incognito or a different browser

abchatra:
Looks ok to me, popped in the sidebar

Jaqster:
I don't see it...  tried both Edge and Chrome incognito...

Jaqster:
Right now we have it set to show to only 25% of users:  
image

riknoll:
alright, i'm going to merge then. any objections?

Jaqster:
Go for it.  We can continue to test/tweak...

Issue #3937, opened 2021-03-26
Feature request: `edge connector sound on/off` block


microbit-mark:
It would be good to have an `edge connector sound on/off` block in a similar way to speaker on/off. The equivalent of `set audio pin none`. 

We could also update the simulator when this block is used to hide/strike-through the headphone connection image.

See comment in https://github.com/microsoft/pxt-microbit/issues/3740#issuecomment-753995990 for the user story of needing to access all edge connector pins.

_Originally posted by @microbit-mark in https://github.com/microsoft/pxt-microbit/issues/3740#issuecomment-778241513_

PR #4489, opened 2022-03-11
Add a block for disabling the audio output pin

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/3937

Tested on V1 and V2 by @livcheerful (thanks again!)

Also has simulator support that looks like this when pin is disabled:

Screen Shot 2022-03-11 at 11 50 04 AM

### DIFF
--- a/sim/state/edgeconnector.ts
+++ b/sim/state/edgeconnector.ts
@@ -29,6 +29,8 @@ namespace pxsim {
 }
 
 namespace pxsim.pins {
+    export let edgeConnectorSoundDisabled = false;
+
     export function digitalReadPin(pinId: number): number {
         let pin = getPin(pinId);
         if (!pin) return -1;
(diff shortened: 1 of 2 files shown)


sim/state/edgeconnector.ts
```diff
@@ -183,4 +185,41 @@ namespace pxsim.pins {
...
+                if (enabled) {
+                    existing.remove();
+                }
+                else {
+                    return;
+                }
+            }
+
+            if (!enabled) {
+                const img = document.createElementNS("http://www.w3.org/2000/svg","image") as SVGImageElement;
```
jwunderl, line 217:
```suggestion
                const img = document.createElementNS("http://www.w3.org/2000/svg", "image") as SVGImageElement;
```

nit

jwunderl (approved):

PR #4494, opened 2022-03-24
allow disabling plotbargraph console output

pelikhan:
Add a typescript only field to disable automatic output of led.plotBarGraph to the console. The default behavior is unchanged.

### DIFF
--- a/libs/core/led.ts
+++ b/libs/core/led.ts
@@ -25,7 +25,7 @@ namespace led {
     /**
      * Controls where plotbargraph prints to the console
      **/ 
-    let barGraphToConsole = true
+    export let barGraphToConsole = true
 
     /**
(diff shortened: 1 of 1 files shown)


libs/core/led.ts
```diff
@@ -21,6 +21,11 @@ namespace led {
     let barGraphHigh = 0;
     // when was the current high value recorded
     let barGraphHighLast = 0;
+    
+    /**
+     * Controls where plotbargraph prints to the console
+     **/ 
+    let barGraphToConsole = true
```
riknoll, line 28: should this be exported?

pelikhan: duh

riknoll (approved):

abchatra:
Looks good

Issue #4523, opened 2022-04-15
Rename new Play Sound block


Jaqster:
Rename this block to "Play Sound Effect" or something, so we don't have 2 Play Sound blocks...

![image](https://user-images.githubusercontent.com/22108954/163603860-99035391-ac00-46cb-bb82-241caf669c58.png)

abchatra:
Another idea is to have the giggle and other effects as a drop down in an get block which can be dragged into the the play sound block.

Issue #4454, opened 2022-01-20
Clean up Music API's


Jaqster:
Request to standardize, and simplify Music/Sounds API's

Blocks representation - 
![MusicBlocks](https://user-images.githubusercontent.com/22108954/150388381-aa488d31-cc70-47a2-9b17-5a24b10dbae6.png)

TS representation - 
![MusicTS](https://user-images.githubusercontent.com/22108954/150388401-1832d792-19fc-4770-ac13-f422348173fe.png)

Play Sound should be in the Music namespace.  
There are 3 different ways to represent a note/melody and 2 different ways to represent tempo...
Is there a way to standardize/simplify?

jaustin:
I'd like to add the following to consideration:
<img width="457" alt="image" src="https://user-images.githubusercontent.com/147113/164488526-0f10cc25-37df-4939-9581-d48701adb1a1.png">
```typescript
music.playMelody("- - - - - - - - ", 120)
music.startMelody(music.builtInMelody(Melodies.Dadadadum), MelodyOptions.Once)
music.playTone(262, music.beat(BeatFraction.Whole))
music.ringTone(262)
soundExpression.giggle.playUntilDone()
soundExpression.giggle.play()
soundExpression.playSoundEffect(soundExpression.createSoundEffect(WaveShape.Sine, 2000, 0, 1023, 0, 500, SoundExpressionEffect.None, InterpolationCurve.Linear))
pins.analogPitch(200, 100)
```

abchatra:
This is implemented.

PR #4545, opened 2022-04-19
Combine sound blocks

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4523

Also part of https://github.com/microsoft/pxt-microbit/issues/4454

Deprecates the old play sound blocks and combines them with the new block. Also move the new block under the music namespace and changes the return type to string.

Here are the new blocks (note that builtin sounds like "giggle" have their own reporter block):

Screen Shot 2022-04-19 at 4 14 10 PM

And here's what the old blocks were:

Screen Shot 2022-04-19 at 4 16 54 PM

@Jaqster FYI!

### DIFF
--- a/libs/core/music.cpp
+++ b/libs/core/music.cpp
@@ -47,11 +47,11 @@ int volume() {
 * @param enabled whether the built-in speaker is enabled in addition to the sound pin
 */
 //% blockId=music_set_built_in_speaker_enable block="set built-in speaker $enabled"
-//% blockGap=8
 //% group="micro:bit (V2)"
 //% parts=builtinspeaker
 //% help=music/set-built-in-speaker-enabled
 //% enabled.shadow=toggleOnOff
+//% weight=0
 void setBuiltInSpeakerEnabled(bool enabled) {
 #if MICROBIT_CODAL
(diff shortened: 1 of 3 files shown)


riknoll:
I *could* also combine the melody blocks with these if that is of interest... Might be worth discussing

jwunderl (approved):

Jaqster:
Hooray!!!  I love reducing the number of blocks we have!  What does the field editor look like now?  How are we surfacing the named sounds?

riknoll:
@Jaqster the named sounds are completely separate from the field editor. The field editor itself is unchanged. I'd love to put the named sounds in the the gallery, but unfortunately they are all multi-part sounds so it's not possible right now.

Jaqster:
Hmmm.  So they would have to grab that separate reporter block and drop it in?  Not ideal, and feels a bit like a regression...  Should we find some time to talk it out?  Would also like to get Micro:bit folks to weigh in...  @jaustin

jaustin:
We'd also like to discuss this with you all (and amongst ourselves, we haven't had a chance yet!). Could the Thursday call work?

Issue #4535, opened 2022-04-15
new tutorial theming is arcade based


jwunderl:
**Describe the bug**

![image](https://user-images.githubusercontent.com/5615930/163605700-6db95f1f-3286-464f-8b8b-603f608d74b4.png)

* shouldn't use a gameboy style icon for sim, should probably be micro:bit?
* teal is arcade color, not micro:bit
![image](https://user-images.githubusercontent.com/5615930/163605923-d2b6e9bd-934f-4ef8-bc74-f2adb943ae1c.png)

**To Reproduce**
1. Go to tutorial in beta

**Expected behavior**
should be themed for micro:bit

Issue #4534, opened 2022-04-15
Hard to read teal against the blue


ElizabethRiffle:
![image](https://user-images.githubusercontent.com/14082525/163605696-9db29e71-b7c9-4dda-86ff-4deefd27af17.png)

abchatra:
@unthinkmedia can we get the new colors for the above which blends with the micro:bit editor colors?

PR #4552, opened 2022-04-20
Tutorial theme

riknoll:
Requires https://github.com/microsoft/pxt/pull/8791

Fixes https://github.com/microsoft/pxt-microbit/issues/4535
Fixes https://github.com/microsoft/pxt-microbit/issues/4534

Theming the tutorial using micro:bit editor colors. Also adding some missing file associations to the checked in vscode settings.

Screen Shot 2022-04-20 at 2 27 27 PM

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,13 +8,19 @@
         "**/yotta_modules/**": true,
         "**/yotta_targets": true,
         "**/pxt_modules/**": true
-    },    
+    },
     "search.exclude": {
         "**/node_modules": true,
         "**/yotta_modules/**": true,
         "**/yotta_targets": true,
         "**/pxt_modules/**": true
-    },    
+    },
+    "files.associations": {
+        "*.blocks": "html",
+        "*.overrides": "less",
+        "*.variables": "less",
+        "*.jres": "json"
+    },
     "tslint.enable": true,
     "tslint.rulesDirectory": "node_modules/tslint-microsoft-contrib",
     "typescript.tsdk": "./node_modules/typescript/lib"
--- a/theme/site/globals/site.variables
+++ b/theme/site/globals/site.variables
@@ -87,6 +87,10 @@
 @serialGraphBackground: #d9d9d9;
 
 /*---------------------------
-   Tutorial Code Validation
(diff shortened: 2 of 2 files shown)


abchatra (approved):

dmonroym (approved):

abchatra:
Also fix the issue of arcade icon for simulator to microbit icon ?

riknoll:
oh.... nope! will fix that also

riknoll:
actually, I need a new icon for that. @unthinkmedia any suggestions?

jwunderl (approved):

unthinkmedia:
> 

Hmm...when do you need this by?  Using the microbit icon doesn't really make sense. Maybe a play button, although it doesn't really play?

abchatra:
This should indicate existence of simulator. May be small micobit?

Jaqster:
I have an unpopular suggestion - don't update micro:bit to the new tutorial format.  We know there are discoverability issues with the simulator in this layout.  The benefit of this layout was to accommodate Arcade tutorials where there are lots of instructions that didn't fit in the top bar.  Micro:bit tutorial instructions are much shorter.

abchatra:
Maintaining multiple code base for the same feature is very error prone and time consuming. I think old code is pretty much stale.
However I think we can bring back the old view in the new codebase as an option set by the target. @riknoll any idea how much that work is?

unthinkmedia:
> This should indicate existence of simulator. May be small micobit?

A microbit is to detailed to be discernible at a small scale.  Also, I don't think a chip communicates simulator.  I would suggest we go with what Jacqueline suggested if that is feasible, and if not we might try a "play" icon.  Ultimately I would be in favor of getting rid of the tabbed pattern as it is not very discoverable and this is going to be worse with microbit since there is not mini-sim on the bottom that some kids use to get to that view.  I have some thoughts on how we might go about that if we want a more comprehensive fix.

abchatra:
For this PR let's pick the play icon. Lets take the other things later.

Jaqster:
Let's cost out adding a 3rd view - https://github.com/microsoft/pxt-microbit/issues/4553

PR #4565, opened 2022-04-25
Extension: add iot cloud kit

microbit-mark:
cc @tangjie133

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -310,6 +310,10 @@ Check out [the accessories pages on microbit.org](https://microbit.org/buy/acces
 
 ```codecard
 [{
+  "name": "DFRobot IoT Cloud Kit",
+  "url":"/pkg/DFRobot/pxt-DFRobot_IoT_Cloud_Kit",
+  "cardType": "package"
+}, {
   "name": "iClass IoT",
   "url":"/pkg/KelieLeung/pxt-iClassIoT",
   "cardType": "package"
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -317,7 +317,8 @@
             "microsoft/pxt-jacdac/weight-scale",
             "microsoft/pxt-jacdac/wifi",
             "microsoft/pxt-jacdac/wind-direction",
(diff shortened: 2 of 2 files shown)


abchatra:
@microbit-mark  can you update this PR with the changes at 2 places as in  this PR https://github.com/microsoft/pxt-microbit/pull/4603/files ?

We need 2 changes before of target.config changes which are live for both /beta & live.

jaustin:
@carlosperate now that you have appropriate actions, could you please do the rebase please?

microbit-carlos:
Rebased and entry added to `approvedRepoLib` as well.

jwunderl (commented):
repo is configured correctly, right license, good icon; these two blocks do concern me a little bit though:

image

It seems like we'll definitely get some people creating share links with their username / password and leaking them that way which feels like a concern? I guess we don't have a better way to handle that since we don't have local only 'secrets' files or keyboard entry on the micro:bit.

I think I'll have to defer to @abchatra for final say for that reason, it would be bad if the extension being approved caused people to think there was some safety layer on top of it and shared sensitive information for that reason.

jaustin:
@jwunderl I think that's a good point. Are there currently any tools in MakeCode to create things that are redacted during sharing? We'd previously discussed with @abchatra a kinda of "secrets" or "settings" part of the program that was stored separately, but as far as I saw from our side we didn't anywhere beyond the feeling that this was possibly necessary for this sort of program. 

As far as the use-case here goes, I think one way I can see to avoid secrets going into the program is to do something like create a serial-based password request block, that the user then has to type their password over serial. However, MakeCode serial looks one-way at the moment?

abchatra:
Good point @jwunderl  I have filed an issue to track this: https://github.com/microsoft/pxt-microbit/issues/4841

There is no alternate way for IOT extensions today. I am going to approve this based on precedence (we have approved this in the past). Even though a bit risky, shared URL are private and I hope it is not shared with everyone if it includes password. 

Once we have the feature in the next release, we can retroactively fix these extensions.

abchatra (approved):

PR #4612, opened 2022-05-03
Update ranges on sound effect api

riknoll:
The pxt-microbit side of https://github.com/microsoft/pxt/pull/8811


jwunderl (approved):

dmonroym (approved):

libs/core/soundexpressions.ts
```diff
@@ -330,8 +340,8 @@ namespace music {
         const sound = new soundExpression.Sound();
         sound.wave = waveShape;
         sound.frequency = startFrequency;
-        sound.volume = startVolume;
-        sound.endFrequency = endFrequency;
+        sound.volume = ((startVolume / 255) * 1023) | 0;
+        sound.endFrequency = ((endFrequency / 255) * 1023) | 0;
```
jaustin, line 344: I was looking at this commit to learn how to use the field editor and block markup language and noticed this line looks fishy. Should that scaling be applied to volume in both cases not volume and frequency?

riknoll: ah, yup! good catch!

PR #4616, opened 2022-05-04
print out dmesg() buffer to serial on panic

mmoskal:
also add serial::writeDmesg() for manual printout

This is disabled when serial is redirected.

This helps diagnose crashes, especially 999 which is unhandled exception - it will print out exception value, which is useful especially when it's a string.

fixes https://github.com/microsoft/jacdac/issues/994

### DIFF
--- a/libs/core/serial.cpp
+++ b/libs/core/serial.cpp
@@ -46,6 +46,10 @@ enum BaudRate {
 //% weight=2 color=#002050 icon="\uf287"
 //% advanced=true
 namespace serial {
+#if MICROBIT_CODAL
+    bool is_redirected;
+#endif
+
     // note that at least one // followed by % is needed per declaration!
 
     /**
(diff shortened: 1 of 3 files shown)


libs/core/serial.cpp
```diff
@@ -167,8 +171,10 @@ namespace serial {
     //% blockGap=8
     void redirect(SerialPin tx, SerialPin rx, BaudRate rate) {
 #if MICROBIT_CODAL
-      if (getPin(tx) && getPin(rx))
+      if (getPin(tx) && getPin(rx)) {
         uBit.serial.redirect(*getPin(tx), *getPin(rx));
+        is_redirected = 1;
```
pelikhan, line 176: you should clear the flag in redirectToUSB

pelikhan (approved):

riknoll (approved):

PR #4623, opened 2022-05-09
better handling of nan numbers in common APIs

pelikhan:
A bit more runtime defense in the case of null/undefined/NaN parameters passed to common apis.

- [x] basic.showNumber(NaN) -> ?
- [x] led.plotBarGraph(NaN) -> clear screen
- [x] pause(NaN) => pause(20)

### DIFF
--- a/libs/core/basic.ts
+++ b/libs/core/basic.ts
@@ -10,7 +10,10 @@ namespace basic {
     //% async
     //% parts="ledmatrix" interval.defl=150
     export function showNumber(value: number, interval?: number) {
-        showString(Math.roundWithPrecision(value, 2).toString(), interval);
+        if (isNaN(value))
+            showString("?")
+        else
+            showString(Math.roundWithPrecision(value, 2).toString(), interval);
     }
 }
 
@@ -19,6 +22,7 @@ namespace basic {
  * @param ms how long to pause for, eg: 100, 200, 500, 1000, 2000
  */
 function pause(ms: number): void {
+    if (isNaN(ms)) ms = 20
     basic.pause(ms);
 }
 
--- a/docs/reference/basic/pause.md
+++ b/docs/reference/basic/pause.md
@@ -9,7 +9,7 @@ basic.pause(400)
 
 ## Parameters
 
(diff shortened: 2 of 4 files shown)


libs/core/basic.ts
```diff
@@ -10,7 +10,10 @@ namespace basic {
     //% async
     //% parts="ledmatrix" interval.defl=150
     export function showNumber(value: number, interval?: number) {
-        showString(Math.roundWithPrecision(value, 2).toString(), interval);
+        if (isNaN(value))
```
abchatra, line 13: Shouldn't we just show NaN as a string?

libs/core/basic.ts
```diff
@@ -19,6 +22,7 @@ namespace basic {
  * @param ms how long to pause for, eg: 100, 200, 500, 1000, 2000
  */
 function pause(ms: number): void {
+    if (isNaN(ms)) ms = 20
```
abchatra, line 25: Good to update documentation of pause.

abchatra (changes requested):
Looks good other than the showNumber change.

libs/core/basic.ts
pelikhan: NaN is meaningless for kids and in general anyone that does not have a CS background. In general, `?` felt like a more intuitive way of saying something is wrong with the input.

Also will scroll which is not great.

abchatra: I am not sure question mark is intuitive either. Either way we have document what it means. NaN (not a number) can be used to explain? @Jaqster opinion here?

pelikhan: I updated the docs to explain that ? is NaN... but deeper docs about NaN would be required. They've existed v1 but mostly our APIs never returned those. It's not true anymore with Jacdac when you query a sensor value where the hardware is missing.

riknoll: I think ? is more intuitive than NaN for sure. I like this change

pelikhan: @abchatra can we have a decision here?

pelikhan: It's not just NaN, also "null" and "undefined" are all handled to "?".

abchatra: I am good with your change

Issue #4174, opened 2021-06-11
Set Columns block - potentially confusing to use


Jaqster:
We should think about a better way to do data logging using the Set Columns block.
Lots of opportunities for students to misuse this - not using the right column names, or the right number of columns when logging data.
Idea - infer column names once they're set in the Log Data name/value pairs.
![microbit-screenshot (14)](https://user-images.githubusercontent.com/22108954/121733274-204a7880-caa8-11eb-943d-055e3502036c.png)

abchatra:
Lets try to do the prepopulated string values atleast.

abchatra:
@jwunderl did you end up implementing prepopulated string values?

microbit-mark:
Had someone report that logging wasn't working, on a closer inspection of the problem, it was this issue. A simple case of using mixed cases in declaring and using the variable.

![image](https://user-images.githubusercontent.com/31242877/128978793-668d0768-5d50-415a-8992-191b93b3bb5a.png)

I could see this being a really common problem.

microbit-mark:
In our user study, another user tried to replace the column headers with inputs. 
![Image 2021-11-16 at 9 19 41 am](https://user-images.githubusercontent.com/31242877/141957592-4ad32db4-0b94-449f-95eb-1ba7ac4a9e19.jpg)

This errors and they identified that this would not work and so they started again.

aorczyk:
When I used it first time in JS and hadn't seen blocks before, I tried to use it exactly this way. Imho ColumnValue is an unnecessary complication, but maybe there are cases where it is a better solution (ex. many columns).

abchatra:
Idea of columns is useful if you have to just log one column sometimes and still want to get the entire row of data and the column value to be at the right place.

abchatra:
Here is what we are planning to improve the discoverability of the column names

- we will build intellisense into the column names. Something like following
![image](https://user-images.githubusercontent.com/6107272/148420997-48ebc1a0-dc32-45c1-834a-f9056a172dd4.png)
 

Even if the user doesn't use _set columns_  api, intellisense would work if the column value is used once before.

abchatra:
This has an added benefit that in text we can build similar feature for column names.

Other option we brainstormed was enum for column names, however this is hard transition for students from blocks to text.

aorczyk:
[...]

PR #4669, opened 2022-05-19
remove the array from datalogger blocks & just have them directly as shadows on log

jwunderl:
hide the old array based blocks and have the new ones just take their place. they have a set max of 10 params

image

current datalogger category:

image

it would be nice to have proper ...rest params support that behaved in this one to avoid this fairly hacky solution (except... uncapped length just like arrays), but that can be left as future work :)

This doesn't have the field editor that would go on the setColumnTitle / ColumnValue's column field yet, just the right block shape.

re: https://github.com/microsoft/pxt-microbit/issues/4174 (leaving that open as part of the issue refers to the autocomplete field editor)

### DIFF
--- a/libs/datalogger/datalogger.ts
+++ b/libs/datalogger/datalogger.ts
@@ -43,7 +43,7 @@ namespace datalogger {
         });
     }
 
-    //% blockNamespace=datalogger
+
     export class ColumnValue {
         public value: string;
         constructor(
--- a/libs/datalogger/datalogger.ts
+++ b/libs/datalogger/datalogger.ts
@@ -43,6 +43,7 @@ namespace datalogger {
         });
     }
 
+    //% blockNamespace=datalogger
     export class ColumnValue {
         public value: string;
         constructor(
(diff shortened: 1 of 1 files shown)


libs/datalogger/datalogger.ts
```diff
@@ -93,6 +95,62 @@ namespace datalogger {
...
+                data2,
+                data3,
+                data4,
+                data5,
+                data6,
+                data7,
+                data8,
+                data9,
+                data10,
+            ].filter(el => !!el)
```
abchatra, line 150: trim at the end may be better than filtering? What if there is code in the typescript with the intermediate values being undefined?

abchatra (approved):

libs/datalogger/datalogger.ts
jwunderl: if any of data1-10 are undefined, this will filter them out and leave the other ones afterwards -- if they did

```typescript
datalogger.log(datalogger.createCV("a", 0), null, undefined, datalogger.createCV("b", 1));
```

then removing the two nullish ones and only passing the two valid ones to logData feels like the right behavior to me?

abchatra: Oh yes, these are column, value pairs. Makes sense. I was just thinking from data value perspective.

PR #4670, opened 2022-05-19
Add jsDoc to sound effect/expression functions

### LINKED ISSUE #4656: Lack introduction of "play sound" block
**Describe the bug**
Lack introduction of "**play sound**" block

1.Nabigate to https://makecode.microbit.org/beta#editor
2.Click **Music**
3.Scroll down to **micro:bit(V2)** part

**Expect behavior**
Show brief introduction of block to tell user the function of it.
image

**Actual behavior**
Only **set built-in speaker enabled block** has brief introduction. Another three blocks under **micro:bit(V2)** doesn't.
image

**Additional context**
1.OS: Windows10
2.makecode.microbit.org version: 4.1.35
3.Microsoft MakeCode version: 7.5.33

ganicke:
The jsDoc was missing from the function headers so the descriptions were missing in the Toolbox view.

Fixes #4656


libs/core/soundexpressions.ts
```diff
@@ -312,6 +317,17 @@ namespace music {
...
     }
 
+    /**
+     * Create a sound expression from a set of sound effect parameters.
+     * @param waveShape for the sound effect 
+     * @param startFrequency for the sound effect waveform
+     * @param endFrequency for the sound effect waveform
+     * @param startVolume of the sound, or starting amplitude
+     * @param endVolume of the sound, or ending amplitude
+     * @param duration in milleseconds (ms) that sound will play for
```
riknoll, line 327:
```suggestion
     * @param duration in milliseconds (ms) that sound will play for
```

riknoll (approved):

libs/core/soundexpressions.ts
ganicke: Thanks, why does my spelling suck? I'll have to break down and reinstall a spellcheck extension. I've tried a few but I generally haven't like any of them.

PR #4671, opened 2022-05-19
disable serial log by default for now

jwunderl:
For now datalogging blocks in simulator won't look like they do anything until special case simulator support is added, but doing that separately. Have to disable this default behavior as it requires a bootloader update to work properly on hardware


abchatra:
@jaustin fyi

abchatra (approved):

PR #4677, opened 2022-05-20
New format for approvedRepoLib

jwunderl:
See https://github.com/microsoft/pxt/pull/8856

@pelikhan added you to check on the jacdac bits, as I pulled them over to the new approved extension field. I also put them in a jacdac label just for testing 
image
but I'll defer to you on if we leave that or I can just delete the tag.


pelikhan (approved):

pelikhan:
You could also hide those libraries as there is a specialized experience to add them in the editor.

abchatra (approved):

Issue #4684, opened 2022-05-23
V1 Yotta target needs to be updated


microbit-carlos:
Basically this line:

https://github.com/microsoft/pxt-microbit/blob/09c107825802d77f8659fc84113fc910bcfc127e/pxtarget.json#L140

Should be updated to `bbc-microbit-classic-gcc@https://github.com/lancaster-university/yotta-target-bbc-microbit-classic-gcc`, as the Yotta registry is no longer operational and it needs to download the target directly from a GitHub repo.

Without this change I wasn't able to build locally for V1.

abchatra:
@microbit-carlos can you produce a PR?

microbit-carlos:
https://github.com/microsoft/pxt-microbit/pull/4687

PR #4687, opened 2022-05-24
Update micro:bit V1 Yotta target to use GitHub.

microbit-carlos:
By adding the GitHub repo URL, otherwise it tries to fetch the target from the Yotta registry, which is no longer operational.

Fixes https://github.com/microsoft/pxt-microbit/issues/4684.


abchatra (approved):

abchatra:
Thanks @microbit-carlos

Issue #4676, opened 2022-05-20
Datalogging unreliable when MakeCode is connected via webUSB on V2.00 DAPLink 0255


jaustin:
**Describe the bug**

(this is likely a CODAL/DAPLink bug that's affected by MakeCode, but filing here as the top level user-visible location)

Using current /beta and 0255 on a V2.00 micro:bit (Git SHA: 1436bdcc67029fdfc0ff03b73e12045bb6a9f272 ) I don't see reliable logging with the following progamme

**To Reproduce**
Load the following into /beta
```typescript
datalogger.onLogFull(function () {
    basic.showIcon(IconNames.No)
    while (true) {
        basic.pause(100)
    }
})
input.onGesture(Gesture.Shake, function () {
    if (input.buttonIsPressed(Button.AB)) {
        basic.showIcon(IconNames.Skull)
        datalogger.deleteLog()
        basic.clearScreen()
    }
})
basic.forever(function () {
    datalogger.log(datalogger.createCV("x", input.acceleration(Dimension.X)), datalogger.createCV("y", input.acceleration(Dimension.Y)), datalogger.createCV("z", input.acceleration(Dimension.Z)))
    led.toggle(2, 3)
})
```

**Expected behavior**
I would get continuous logging until the log fills up. If I disconnect web USB and flash the program via mass storage, it works reliably.

If I flash via web USB and then check the log, sometimes it's empty. If I re-flash, the log might log some small number of values (typically up to 4 seconds) and then stop. 

4 seconds is about the time it takes for the device to re-enumerate on my PC. I think this looks like the log getting clobbered during re-enumeration/mounting. 

**Screenshots**
Here's an example when mass storage flashing
<img width="1507" alt="image" src="https://user-images.githubusercontent.com/147113/169579239-44cbcaf0-7933-4240-aa55-a8d6779a2d41.png">

Here's an example using web USB

(I got the following assert Assert.txt

.\source\rtos\RTX_Config.c
Line: 191
Source: Application
)
(notice the much shorter length of time)

<img width="1491" alt="image" src="https://user-images.githubusercontent.com/147113/169582524-1dd9e930-069d-44ab-b5ce-a3c409b96df3.png">

[MY_DATA_web_USB_HTM.zip](https://github.com/microsoft/pxt-microbit/files/8742307/MY_DATA_web_USB_HTM.zip)

**micro:bit version (please complete the following information):**
V2.00 running 0255
[...]

PR #4689, opened 2022-05-25
Bumping pxt-microbit CODAL version to v0.2.38.

microbit-carlos:
The only addition since v0.2.37 is an increase in a timeout that should improve the data logging problem in issue https://github.com/microsoft/pxt-microbit/issues/4676.


pelikhan (approved):

pelikhan:
It's in beta

PR #4696, opened 2022-05-25
Configure expandable arguments

riknoll:
Requires https://github.com/microsoft/pxt/pull/8868 but safe to merge now!


jwunderl (approved):

jwunderl:
Hm did we want to use this to set a lower threshold for swapping to vertical for the data logger log data / set columns blocks as well?

abchatra (approved):

jaustin:
Yes, @jwunderl we were just discussing this - could we go to vertical after 2 for data logging please? So 3rd one is vertical?

PR #4699, opened 2022-05-26
adding flags to test reset on connection

pelikhan:
Two flags to test removing resets on connecting web usb:

- `webusbbaud=0` disables setting baud rate to 115200
- `webusbreset=0` disables reseting cortex on connection

@jaustin


mmoskal (approved):

pelikhan:
passing breaks `webusbbaud=0` serial output

PR #4708, opened 2022-05-27
rock paper scissors with sounds

pelikhan:
This is a variation on the rock paper scissor + sound. Most of the gifs are still kind of ok to reuse, but it does add one step. 
It's a fun extension.


Jaqster:
I feel like we should start adding some sort of "V2 only" label on the Home Page cards...  Maybe @clkantner can help us think through a design for this...

pelikhan:
Copied from the data logger card

ganicke (approved):

pelikhan:
@Jaqster  good to go?

Jaqster:
Yes!  We did this with the teachers yesterday, and they loved it!!!

Issue #4723, opened 2022-06-03
Remove hover highlight and button for microbit logo on V1 board simulator


kimprice:
**Describe the bug**
V1 boards in the simulator show the microbit logo as clickable, but this is not accurate. 

**To Reproduce**
1. Open new workspace (do not include any V2 blocks)
2. Hover on the microbit logo in the simulator
3. See the highlight on hover and the cursor indicate it is clickable

**Expected behavior**
We should disable this attribute when the board is V1 since the logo is not clickable on V1 boards.

![image](https://user-images.githubusercontent.com/15070078/171922876-a9883854-04ed-4d9f-a244-c8c57cdeafa6.png)

kimprice:
Also, when the logo is clickable, we should support clicking anywhere within the logo, not just on the outline of the logo.

abchatra:
@jwunderl can you take over this issue for tomorrow?

PR #4725, opened 2022-06-04
logo only clickable for V2 board

kimprice:
fixes https://github.com/microsoft/pxt-microbit/issues/4723

Now the logo is only clickable and has hover highlighting when the simulator is V2. Additionally, the entire logo is clickable in V2, rather than just the outline.

### DIFF
--- a/sim/visuals/microbit.ts
+++ b/sim/visuals/microbit.ts
@@ -387,10 +387,11 @@ path.sim-board {
             svg.fill(this.buttonsOuter[2], theme.virtualButtonOuter);
             svg.fill(this.buttons[2], theme.virtualButtonUp);
             svg.fills(this.logos, theme.accent);
-            if (this.domHardwareVersion > 1)
+            if (this.domHardwareVersion > 1) {
                 svg.fills(this.heads.slice(1), "gold");
-            else
+            } else {
                 svg.fills(this.heads.slice(1), theme.accent);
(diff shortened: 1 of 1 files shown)


sim/visuals/microbit.ts
```diff
@@ -1324,7 +1328,7 @@ path.sim-board {
 
         private attachABEvents() {
             const bpState = this.board.buttonPairState;
-            const stateButtons: Button[] = [bpState.aBtn, bpState.bBtn, this.board.logoTouch];
+            const stateButtons: Button[] = (this.domHardwareVersion > 1) ? [bpState.aBtn, bpState.bBtn, this.board.logoTouch] : [bpState.aBtn, bpState.bBtn];
```
jwunderl, line 1331: Does this attachabevents get called after ensureV2 for sure? might be worth pulling out the rest of the function into a separate ` private attachButtonEvents(btn)` function and calling it on `this.board.logoTouch` as part of `ensureHardwareVersion`, so that we don't have a hidden situation where order matters between that and attachABEvents

jwunderl (approved):
I'd slightly prefer pulling out the this.board.logoTouch registration from attachABEvents into updateHardwareVersion for consistency / making it easier to follow as noted in the other comment, but otherwise lgtm~

abchatra (approved):

PR #4726, opened 2022-06-06
Digi Yoyo: Incorrect increasing count by 1

pgittings:
Code should be:
`count = count + 1`

not:
`count = count + count - 1`

This wouldn't increase count by 1


pelikhan:
Good spot! and thanks for creating a PR!

pelikhan (approved):

Issue #4716, opened 2022-06-03
Data Logger block too long


Jaqster:
For small screens, very hard to work with the block.
Suggest moving to multi-line split after 2 fields.
![image](https://user-images.githubusercontent.com/22108954/171918666-6c214a91-a928-4380-a9bf-38888c7e2c53.png)

asmeets:
On the iOS side, the set columns block drops to a vertical view after five column names. The log data block does as well, as referenced by @Jaqster, but given the size of the original block, this does present a challenge.

PR #4727, opened 2022-06-06
datalogger block configure line break for inline input mode

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/4716

configure data logger block w/ options added in https://github.com/microsoft/pxt/pull/8868 to make it break sooner:

image

### DIFF
--- a/libs/datalogger/datalogger.ts
+++ b/libs/datalogger/datalogger.ts
@@ -132,6 +132,7 @@ namespace datalogger {
     //% data9.shadow=dataloggercreatecolumnvalue
     //% data10.shadow=dataloggercreatecolumnvalue
     //% inlineInputMode="variable"
+    //% inlineInputModeLimit=2
     //% group="micro:bit (V2)"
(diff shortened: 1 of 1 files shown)


pelikhan:
You could break even sooner. If you put input blocks in there, it'll cover the screen.

Jaqster:
Have we tested on a Chromebook screen size?  I like option A if we can get away with 2 column blocks inline before we break to multi-line.  I definitely don't want to always do external... the plus + on the next line by itself looks weird.  But if 2 column blocks are still too long on a Chromebook screen size, then fallback could be your 3rd option.

jwunderl:
Alright, going with 3rd option, depends on the chromebook but any block taking in two parameters that are normal block size will definitely make it too long for lots of small screens (including e.g. the one you had in the issue to begin with being partially cut off: https://github.com/microsoft/pxt-microbit/issues/4716)

abchatra (approved):

PR #4728, opened 2022-06-07
Adding card for new v2 tutorial

kiki-lee:
Adding Claire's card for "Dance to the Beat".  We won't need it until we add the tutorial to the homepage.


jwunderl (commented):
probably worth minifying, quick one came out to 80kb from 240~

dance-card-minified

kiki-lee:
> probably worth minifying, quick one came out to 80kb from 240~
> 
> dance-card-minified

Good call. I have a smaller version.
dance-card

PR #4740, opened 2022-06-13
pxt-jacdac min version to v1.8.13

pelikhan:
- [x] bump minimum version of pxt-jacdac


jwunderl (approved):
LGTM on bumping min version, I'll defer to abhijith on setting preferred or not. (Also note that as of beta setting an extension as preferred is done as a field on the extension itself instead of that list e.g. https://github.com/microsoft/pxt-microbit/blob/3894d380b58fb9dde134e93c622d0d63ed402070/targetconfig.json#L342)

pelikhan:
I've removed preferred for this change. it needs to wait for the release anyway

abchatra (approved):

PR #4761, opened 2022-06-16
Extension categorization

abchatra:
First cut with renames. We are going to categorize extensions which have more than hundred users per month.


jwunderl (approved):
LGTM

PR #4763, opened 2022-06-17
Support docs for datalogger extension

### LINKED ISSUE #4752: Docs for Data Logger extension
Let's add a documentation page for the Data Logger extension, like we have for NeoPixels (https://makecode.microbit.org/pkg/microsoft/pxt-neopixel) and our other 1st party extensions.

ganicke:
Add a `datalogger` top level page to support the extension card's "Learn more" link. Also, add the `log` and `setColumnTitles` pages.

Fixes #4752

### DIFF
--- a/libs/datalogger/docs/reference/datalogger/log.md
+++ b/libs/datalogger/docs/reference/datalogger/log.md
@@ -12,7 +12,7 @@ A log entry is made up of one or more "column-value" objects inserted into an ar
 
 ## Parameters
 
-* **data1 - data10**: one or more (up tp 10) of "column-value" objects to write to the data log.
+* **data1 - data10**: one or more (up to 10) of "column-value" objects to write to the data log.
 
 ## Example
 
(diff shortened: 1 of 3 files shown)


libs/datalogger/docs/reference/datalogger/log.md
```diff
@@ -0,0 +1,33 @@
...
+datalogger.log(null)
+```
+
+Data log entries are written to the log as an array of "column" and "value" data items. Each data value you want to record in the log has an column which it belongs to. The data value is attached to it's column by creating a "column-value" object with the data value matched to a column name.
+
+A log entry is made up of one or more "column-value" objects inserted into an array. This array is sent to the data log to write as the next log entry.
+
+## Parameters
+
+* **data1 - data10**: one or more (up tp 10) of "column-value" objects to write to the data log.
```
jwunderl, line 15:
```suggestion
* **data1 - data10**: one or more (up to 10) of "column-value" objects to write to the data log.
```

something something making sure this is tp top shape

jwunderl (approved):

Jaqster (approved):
Thanks Galen!

PR #4799, opened 2022-07-07
remove youtubeid in jacdac gallery

pelikhan:
The gallery item has a url and a youtubeid, pxt gets confused and renders a "play video" button that points to the URL. This feels like a regression (can we have a video + a project link?) but it's probably not worth a hot patch.

You can test the current broken behavior in makecode.microbit.org

image


jwunderl (approved):
iirc the default button behavior got to be too much and someone swapped it so that it'd only be one button by default w/ anything more handled by setting the otherActions field for extra buttons

abchatra:
Closing old pull requests.

Issue #4812, opened 2022-07-15
Maqueen Avoiding Obstacles 


4CodeLife:
Hello GitHub I am trying to download code for the maqueen onto the Microbit and I continuously get the following message: The hex file is not available, please connect to the internet and try again. However, I am connected to the internet. I tried downloading a different project on the same Microbit. That code downloaded with no problem. Does anyone have any suggestions? Please
Here is the code link:
https://makecode.microbit.org/_a34Jt9EAU02i

jwunderl:
Log: https://makecode.com/compile/dd13b0493958372b7a1c8b36291a7c639b55cbff34ae481be2e9dbd503eff21f.log w/

```
/home/build/prj2/source/pointers.cpp:342:13: error: 'vector' does not name a type; did you mean 'error'?
     typedef vector<Action> vA;
             ^~~~~~
             error
```

but hasn't changed in a while so probably on us?
https://github.com/DFRobot/pxt-maqueen/blob/master/maqueenIR.cpp#L6

martinwork:
It builds OK on live after updating the Maqueen extension.
The Maqueen extension in the shared project is 1.2.3. After updating, it's 1.7.0 and the IR category has a different name.

riknoll:
Thanks @martinwork! This is indeed an issue of the extension being outdated. @4CodeLife you can download the latest extension by switching to typescript and clicking the refresh button next to the extension in the file explorer on the left

Or, just open this project where I updated it for you:

https://makecode.microbit.org/_e6jdcAetk9qC

4CodeLife:
@riknoll Thank you for your response. I am teacher and I will be using the Maqueen Robot this school year. I am hoping that my students will not ran into the same issue as I did this week when they try to download MaQueen code. What should I have my students do they ran into a downloading issue with the MaQueen?

jwunderl:
Since there is a new tagged version, the students shouldn't run into this issue -- whenever you add the extension from the extensions view it grabs the latest release (1.7.0 currently). This issue would happen when someone has an old project that was created when that version of maqueen (1.2.3) was the latest version.

Could you say where you got that share link from -- was it an example you got from somewhere? It looks like it was likely made back in march 2020 looking at the version history of maqueen

@riknoll do you think we should add an upgrade rule like https://github.com/microsoft/pxt-microbit/blob/master/targetconfig.json#L458 since the old version c++ doesn't work or should it be fine as is?

riknoll:
[...]

PR #4816, opened 2022-07-20
set min version for pxt-maqueen to fix compile

jwunderl:
re: https://github.com/microsoft/pxt-microbit/issues/4812#issuecomment-1190785183

I just picked to always update to the current latest as that's been around for a while at this point


srietkerk (approved):
LGTM!

PR #4823, opened 2022-07-29
Adding Catalan (ca) language code

jwunderl:
As you mentioned https://crowdin.com/project/makecode/ca seems pretty close, @anzhou5 when you are comfortable with the state it is in state we can merge this in then bump and it will show up in /beta


riknoll (approved):

anzhou5:
Thanks for setting this up @jwunderl! I'll let you know when it's ready

abchatra (approved):

PR #4824, opened 2022-08-02
Add dialog for 'program too large' errors

riknoll:
This is the micro:bit side of https://github.com/microsoft/pxt/pull/8991

The build will fail until that PR is merged/released. 

Opening this as a draft PR so that we can discuss what the contents of this dialog should be. This dialog will show whenever a user tries to download a program that is too large for the micro:bit v1's flash. In addition to text, we can also link a help/support page and provide images.

The purpose of this dialog is to provide the user the option to download their program just for the micro:bit v2 which has a much larger flash size than the v1. Because normal hex files from the micro:bit editor work for both v1 and v2 micro:bits, this dialog needs to make it super clear that if you recompile for v2 then the hex file will not work on a v1 micro:bit

@Jaqster @jaustin @BeckHaru

### DIFF
--- a/editor/dialogs.tsx
+++ b/editor/dialogs.tsx
@@ -13,3 +13,26 @@ export function cantImportAsync(project: pxt.editor.IProjectView) {
         ]
     }).then(() => project.openHome())
 }
+
+
+export async function showProgramTooLargeErrorAsync(variants: string[], confirmAsync: (opts: any) => Promise<number>) {
+    if (variants.length !== 2) return undefined;
+
+    const choice = await confirmAsync({
+        header: lf("Program too large..."),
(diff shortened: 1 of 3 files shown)


jwunderl (approved):
code looks good pending the wording ~

riknoll:
@Jaqster could you provide me with some copy for this dialog?

Jaqster:
Yup.  Taking inspiration from our other dialogs, how about this:
Title:  Oops, there is a problem downloading your code.
Description:  Great coding skills!  But unfortunately, your program is too large to fit on a micro:bit V1 :-(.  You can go back and try to make your program smaller, or you can select to download your program onto a micro:bit V2 (which has bigger capacity).
Buttons: 
- Go back (grey)
- Download for V2 only (green)

image

riknoll:
Here it is:

image

jwunderl (approved):

PR #4827, opened 2022-08-17
experiment with screen in micro:bit

tballmsft:


tballmsft:
Michal.

To torture myself, I am trying to put together the elements I think I need to get the screen into pxt-microbit (in advance of then pulling it out into an extension). I've made some progress but am getting some odd errors.

-- Tom

mmoskal:
I added `disablesVariant` but now there is a name clash on Image class, which has different definition in micro:bit. 

AFAIK there is no way to rename the one in micro:bit core from an extension, and if you rename the one in game---light, your code won't work with a regular Arcade. Maybe you can define some type aliases.

The pin stuff can be probably fixed by editing the screen code directly (it goes into extension, so no need to introduce anything in common packages).

mmoskal:
now it fails on `mixer` - I think you want to drop it and use micro:bit native sound

tballmsft:
Yes, I thought I had removed the dependence on mixer.

PR #4842, opened 2022-09-01
Extension: add button-clicks

martinwork:
cc @bsiever


abchatra:
Looks good overall. Nice extension.

Minor nit: @bsiever can you remove more category. It is just few blocks and more category is not recommended when you have less than 10 blocks. Put the more category blocks lower in the hierarchy. 
image

abchatra:
@martinwork please add @ me to these extensions PR. I will miss it otherwise.

martinwork:
Will do, thanks @abchatra

bsiever:
@abchatra: I'm open to that, but I was a little afraid that the features I put in Advanced aren't really the core functionality and may add some confusion about which blocks to use.   But I agree that it's a little silly to have a separate toolbox spot just for two blocks.   Thoughts?

bsiever:
@abchatra Ignore my previous post.  I removed the "More" category and used an "Advanced" group for those two blocks (forgot that was possible until I was doing something else).  Update has been committed.

abchatra (approved):

PR #4856, opened 2022-09-17
added the icon of electric guitar to correct path

akshitagupta15june:
Previously the image path in music.md is /static/mb/projects/electric-guitar.png but the image is inside /static/mb/projects/electric-guitar/electric-guitar.png

so i have added the image in /static/mb/projects/electric-guitar.png

image


akshitagupta15june:
@ganicke please merge the request to solve the above error

ganicke (approved):
PNG moved to the correct folder, thanks. The other one should get deleted but oh well.

PR #4867, opened 2022-10-10
Extension: add gatorRTC

martinwork:
@abchatra DO NOT MERGE
I think these problems should be fixed: https://github.com/sparkfun/pxt-gator-rtc/issues/3


martinwork:
@loricrotser Thanks for the updates.
@abchatra I think this is good to go now.

abchatra:
Blocks seems bit broken. Is this expected?
image

loricrotser:
The blocks are initialized to 0 such that the user can set them as they see fit. Or are you referring to the multiple rows? I have not changed anything with this - to my knowledge it has always looked this way.

abchatra:
I am referring to the **: , - -20** before the zero

abchatra:
Oh this is a date format. It was hard to parse. Probably single line would be much better than multiple rows here.

PR #4886, opened 2022-11-18
Extension: add Apprentice Car

martinwork:
cc @resolute-support

@abchatra I have suggested:
- changing parameters in blocks to lower case, though there are already extensions with Capitalised parameters.
- making some of the JavaScript names more meaningful

I don't know why the JavaScript function PH() is highlighted in green.

Many of the JavaScript mode tooltips have sometimes said (+1 overload), but that went away in a new project. Maybe it's a caching issue after updating the extension version in an existing project?

Here's a project with all the blocks, not necessarily in a sensible order.
https://makecode.microbit.org/_0KsTxhEchMTR
image

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -493,6 +493,10 @@ Check out [the accessories pages on microbit.org](https://microbit.org/buy/acces
 
 ```codecard
 [{
+  "name": "Resolute Apprentice Car",
+  "url":"/pkg/resolute-support/pxt-apprentice_Car",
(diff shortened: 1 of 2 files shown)


resolute-support (approved):
@abchatra i have changed parameters in blocks to lower case and made some of the JavaScript names better

abchatra (approved):

abchatra:
Looks great. Thank you!

PR #4888, opened 2022-11-18
Update pxt with hopeful fix for WinJS links to new app

thsparks:


thsparks:
Bypassing check-in gates to expedite testing

PR #4894, opened 2022-12-11
Update pins.md

amosko:


abchatra (approved):

abchatra:
Thanks for the fix.

amosko:
@abchatra I'm not sure how I missed to approve from a year ago, but I was just on the same page and saw this was never merged. It seems like the checks got stuck on one of the builds, so merging is blocked. I guess I don't have access to see what's wrong with that, can you maybe try to trigger another build so this can merge? Thanks!

PR #4911, opened 2023-01-12
Expand timing info for music blocks reference

### LINKED ISSUE #4907: Rest parameter should be in Beats (not ms)
Hi Galen,
This was reported by an educator for micro:bit, but I think this applies to all of PXT docs.
The docs show the Rest block takes a number/millisecond parameter, but the default value in our editor is actually a beat.
https://makecode.microbit.org/reference/music/rest
image
image

Can we add Beats as the default parameter value for the Rest block in the documentation?  Maybe we can show how this function can also be overloaded by passing in a ms number value as well...

ganicke:
Add some clarifying info for the timing parameters in the older music blocks reference pages.

**Note**: The convention is to use the JavaScript parameter names in listings under **Parameters** on the ref pages. The term "duration" is used in the descriptions to associate that with the parameter.

Closes #4907

### DIFF
--- a/docs/reference/music/beat.md
+++ b/docs/reference/music/beat.md
@@ -1,16 +1,10 @@
 # Beat
 
-Returns the duration of a beat in milli-seconds
+Returns the duration of a beat in milliseconds
 
 ```sig
 music.beat(BeatFraction.Whole)
 ```
-
-## ~ hint
-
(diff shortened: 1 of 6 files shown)


Jaqster (approved):
Looks great!  Thanks for updating this Galen.

PR #4927, opened 2023-02-02
Extension: gator-environment v1.1.2 supports V2

martinwork:
@loricrotser

@abchatra
This changes "dv:mbcodal" to "min:v1.1.2" in the "approvedRepoLib" section.

Is that the right way to enable V2 builds and cause existing projects to upgrade?

Does the "upgrades" section need a similar change?


abchatra:
@martinwork I believe this is good enough.

abchatra (approved):

PR #4946, opened 2023-02-16
Extension: add Cytron MOTION:BIT

martinwork:
Arising from https://support.microbit.org/helpdesk/tickets/60301 (private)

@waiweng83

@abchatra The extension duplicates the Neopixel extension code because it is required to use Neopixels at the same time as Bluetooth. There are only 2 Neopixels on the board and their extensive tests have not found any problems. Is there a better way to reuse the Neopixel code?

https://makecode.microbit.org/_M809MrTK4Pjf

image

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -493,6 +493,10 @@ Check out [the accessories pages on microbit.org](https://microbit.org/buy/acces
 
 ```codecard
 [{
+  "name": "Cytron MOTION:BIT",
+  "url":"/pkg/CytronTechnologies/pxt-motionbit",
+  "cardType": "package"
+}, {
   "name": "MAKE&LEARN Didacbot",
   "url":"/pkg/MakeAndLearn/pxt-didacbot",
   "cardType": "package"
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -775,7 +775,8 @@
             "sparkfun/pxt-gator-uv": { "tags": [ "Science" ] },
             "dfrobot/pxt-dfrobot_bosonkit": { "tags": [ "Science" ] },
             "resolute-support/pxt-apprentice_car": { "tags": [ "Robotics" ] },
-            "makeandlearn/pxt-didacbot": { "tags": [ "Robotics" ] }
+            "makeandlearn/pxt-didacbot": { "tags": [ "Robotics" ] },
+            "cytrontechnologies/pxt-motionbit": { "tags": [ "Robotics" ] }
         },
         "upgrades": {
             "tinkertanker/pxt-iot-environment-kit": "min:v4.2.1",


martinwork:
@waiweng83 Including the neopixels code creates problems that I should have thought of before. If another extension loads the neopixels code, as a dependency or I guess by including the code like here, there will be name clashes and duplicate definitions. MakeCode says there are Extension Errors and can't fix them. The project will not build.

Maybe the names could be changed?

@abchatra Can you help?

waiweng83:
@martinwork Good catch. I've never thought of this. I think the best way will be putting all the functions in my namespace right?
I will do it and update here once it's done.

Thanks a lot.

waiweng83:
@martinwork I've updated the motionbit extension with new release.
Basically I just put all the neopixel functions into the motionbit namespace.
Also renamed the enum to have a unique name.

waiweng83:
@martinwork There is still name clash on the assembly file. However, when I changed the function name from sendBufferAsm to anything else (eg: motionbitSendBuffer), the code doesn't work anymore.

FYI, I changed the label name in the sendBuffer.asm and also modified the //% shim in ws2812.ts.

Any idea how to fix that?

waiweng83:
Did some testing and noticed that the label name in the asm file must be "sendBufferAsm" in order to work. it doesn't matter what's the code under that function label.

I modify the code to be like this and it works too:
```
sendBufferAsm:
    push {r4,r5,r6,r7,lr}
    pop {r4,r5,r6,r7,pc}
```

Is it cached somewhere? I totally have no idea how this happened.

waiweng83:
I think I've found out the problem.

Referring to https://github.com/microsoft/pxt-microbit/issues/3415#issuecomment-716525159, there is a new function for sendBuffer.

A shimRenames was also added to [pxt-microbit/pxtarget.json](https://github.com/microsoft/pxt-microbit/blob/abad5e4ec82a5f541963bb0c53fe571a82f3f797/pxtarget.json#L52).

So I just removed the sendBuffer.asm file and point my shim directly to "light::sendWS2812Buffer" in v1.1.2.
Now it's working happily together with the official neopixel extension.

Test code: https://makecode.microbit.org/_d37Wv7ML5M5d

@martinwork @abchatra Do let me know if there is any other issue. Thanks.

martinwork:
@waiweng83 Thanks for figuring that out. I can confirm that, with neopixels connected to pin 8, your example flashes 2 of them red/blue. Tested with micro:bit V2.21 and V1.5.

waiweng83:
Thanks @martinwork . Looking forward for the merge.

abchatra (approved):

abchatra:
This extension looks great. Yes calling sendWS2812Buffer is the right way to do this.  Thanks!

waiweng83:
Thanks @abchatra .

PR #4947, opened 2023-02-17
Enable new tutorial system in Microbit

thsparks:
With recent layout updates, all we need to do is set the flag.

image


jwunderl (approved):
Nit: Probably just delete the line since false should be default value?

PR #4970, opened 2023-03-03
microbit: enable sign-in & cloud sync

eanders-ms:


abchatra (approved):

pxtarget.json
```diff
@@ -25,12 +25,35 @@
         "thumbnails": true,
         "publishing": true,
         "importing": true,
+        "showBadges": false,
```
jwunderl, line 28: In case you want to make an easy way to turn this on for testing (or change pretty much anything in pxtarget.json), we do support query parameters for toggling apptarget fields under queryVariants (e.g. in minecraft here https://github.com/microsoft/pxt-minecraft/blob/master/pxtarget.json#L258, these hit if you e.g. load with `?inGame=1` at end of url)

jwunderl (approved):

pxtarget.json
eanders-ms: Oh that's good to know!

PR #4971, opened 2023-03-03
microbit: Enable sign-in as query variant

eanders-ms:
For microbit, enable sign-in with url parameter `identity=1`. This will allow us to test in beta, etc.

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -31,29 +31,7 @@
         ],
         "githubPackages": true,
         "cloudProviders": {
-            "github": {
-                "id": "github",
-                "name": "GitHub",
-                "icon": "/static/providers/github-mark.png",
-                "identity": false,
-                "order": 3
-            },
-            "microsoft": {
-                "id": "microsoft",
-                "name": "Microsoft",
-                "icon": "/static/providers/microsoft-logo.svg",
-                "identity": true,
-                "redirect": true,
-                "order": 1
-            },
-            "google": {
-                "id": "google",
-                "name": "Google",
-                "icon": "/static/providers/google-logo.svg",
-                "identity": true,
-                "redirect": true,
-                "order": 2
-            }
+            "github": {}
         }
     },
     "compile": {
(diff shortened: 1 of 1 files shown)


jwunderl (approved):

pxtarget.json
```diff
@@ -646,6 +624,35 @@
             "appTheme": {
                 "disableBlobObjectDownload": true
             }
+        },
+        "identity=1": {
+            "cloud": {
```
kimprice, line 629: Why do you need the extra cloud field?

kimprice: oh nvm I see it's the same above. Just needed to expand.

pxtarget.json
```diff
@@ -31,29 +31,7 @@
...
-            },
-            "google": {
-                "id": "google",
-                "name": "Google",
-                "icon": "/static/providers/google-logo.svg",
-                "identity": true,
-                "redirect": true,
-                "order": 2
-            }
+            "github": {}
```
kimprice, line 34: why do you need this line?

pxtarget.json
jwunderl: the way queryVariant works is that it applies a patch to the appTarget (the rest of this json blob) -- if you go to arcade and open the devtools you can see `pxt.appTarget.cloud.cloudProviders` contains similar / the same data as this to enable sign in, what this does is for microbit.makecode.com/beta?identity=1 it'll override the default value (the other part changed in this pr) for `pxt.appTarget.cloud.cloudProviders` with the one inside this object

pxtarget.json
eanders-ms: Reverted my earlier changes, moving them to the "queryVariants" section where they'll be merged into the root of this object when `identity=1` is provided.

pxtarget.json
eanders-ms: It's to match the schema of this config object. The JSON snippet specified in this override will be merged into the root of this object when the `identity=1` parameter is present.

PR #4989, opened 2023-03-15
Audio extension initial pass

srietkerk:
The blocks from the codal audio extension have been added as a built-in extension and modified so the blocks are more user-friendly instead of just surfacing APIs. 

image

image

If you would like to take a look at the original blocks, you can add the blocks as an extension from this repo: https://github.com/lancaster-university/pxt-codal-audio

This PR has all the blocks that I recall getting feedback on in our discussion about them. Please let me know if I left anything out. I also encourage playing with them to see if they function in a way that you expect them to, I would love to edit them if they do not operate in a way that users would expect. 

**Couple of Notes**
- I started working on making the music blocks work on the audio blocks. `stop all sounds` was the most obvious block for me, so I started there, this should work on the micro:bit.
- The main functionality of the extension is in `libs/audio-recording/recording.ts` and `libs/audio-recording/recording.cpp`. I know how most of this works, but I may not be able to answer all of your questions, especially if it has to do with the cpp logic since that was done by the folks at Lancaster. 

**Future work** 
- Surface the audio buffer so the recording can be treated as a Playable type in the music blocks
- Port the music blocks from arcade into micro:bit and make the audio blocks and music blocks work well together
- Get the simulator to record audio as well. 
   > **Note** I've already started some work with this `/sim/state/record-audio.ts` holds no-ops to all the shim functions so that the simulator won't crash if you try to run your program with only the simulator, and most of the logic works, but you just can't actually record yet.

**HERE'S THE TARGET**
https://makecode.microbit.org/app/c63e9c63fb9c8acc19ad1fdb101a82d10fc9eb10-58c0c97e33


docs/static/packages/audio-recording/icon.png
eanders-ms, line 1: Is this image a placeholder? It feels very generic, it doesn't convey micro:bit or recording. I also wonder where it was sourced and if we're clear to use it.

libs/audio-recording/_locales/audio-recording-jsdoc-strings.json
```diff
@@ -0,0 +1,17 @@
+{
+  "record": "Functions to operate the v2 on-board microphone and speaker.",
+  "record.audioDuration": "Get the how long the recorded audio clip is",
```
eanders-ms, line 3:
```suggestion
  "record.audioDuration": "Get how long the recorded audio clip is",
```

libs/audio-recording/_locales/audio-recording-jsdoc-strings.json
```diff
@@ -0,0 +1,17 @@
...
+  "record.audioEvent": "Do something based on what the audio is doing",
+  "record.audioIsPlaying": "Get whether the playback is active",
+  "record.audioIsRecording": "Get whether the microphone is listening",
+  "record.audioIsStopped": "Gets whether the board is recording or playing back",
```
eanders-ms, line 7:
```suggestion
  "record.audioIsStopped": "Get whether the board is recording or playing back",
```

libs/audio-recording/_locales/codalAudio-jsdoc-strings.json
```diff
@@ -0,0 +1,16 @@
+{
+  "codalAudio": "Functions to operate the v2 on-board microphone and speaker.",
+  "codalAudio.audioDuration": "Get the how long the recorded audio clip is",
```
eanders-ms, line 3:
```suggestion
  "codalAudio.audioDuration": "Get how long the recorded audio clip is",
```

libs/audio-recording/_locales/codalAudio-jsdoc-strings.json
```diff
@@ -0,0 +1,16 @@
...
+  "codalAudio.audioIsPlaying": "Get whether the playback is active",
+  "codalAudio.audioIsRecording": "Get whether the microphone is listening",
+  "codalAudio.audioIsStopped": "Gets whether the board is recording or playing back",
```
eanders-ms, line 6:
```suggestion
  "codalAudio.audioIsStopped": "Get whether the board is recording or playing back",
```

libs/audio-recording/recording.ts
```diff
@@ -0,0 +1,279 @@
...
+        _playbackHead = 0
+        _memoryFill = 0
+        erase();
+        return
+    }
+
+    /**
+     * Do something based on what the audio is doing
+     */
+    //% block="on audio %eventType"
```
eanders-ms, line 237:
```suggestion
    //% block="on audio $eventType"
```
I believe `$` is preferred over `%`

libs/audio-recording/README.md
eanders-ms, line 1: Do these instructions still make sense?

docs/static/packages/audio-recording/icon.png
srietkerk: This is the image that Lancaster University used. I was also unsure about whether we are allowed to use it and agree it doesn't do a great job at conveying what the extension does. I was unsure of what to put here in lieu of it, however, so I kept it.

pxtarget.json
```diff
@@ -16,7 +16,8 @@
         "libs/settings",
         "libs/flashlog",
         "libs/datalogger",
-        "libs/color"
+        "libs/color",
```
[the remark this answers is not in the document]
srietkerk: As far as I understand it, this is needed so that the audio recording extension is included in the list of extensions that we surface on the extensions page.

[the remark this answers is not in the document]
eanders-ms: Oh I see it's in the bundledirs section!

docs/static/packages/audio-recording/icon.png
eanders-ms, line 1: Extension thumbnails should be 320x240, ideally. Let's hunt around for someone who can draw an icon for this.

eanders-ms: (this one is fine as a placeholder)

eanders-ms (approved):

libs/audio-recording/recording.ts
```diff
@@ -0,0 +1,273 @@
...
+        _recordingFreqHz = 22000
+        _playbackFreqHz = 22000
+        _micGain = AudioGainEnum.Medium
+        music._onStopSound(stopRecording);
+
+
+        control.runInBackground( () => {
+            while (true) {
+
+                switch (_moduleMode) {
```
jwunderl, line 110: probably later on we can take a look at making this not have to watch; it feels a little unfortuante that if you record something at the start it'll still be continuing to loop for as long as the program runs; that would probably look like making the Stopped case break from the loop and spin up the thread again whenever necessary

libs/audio-recording/recording.ts
```diff
@@ -0,0 +1,273 @@
...
+                            _memoryFill += _recordingFreqHz / (1000 / INTERVAL_STEP)
+                        }
+                        break
+                    case AudioRecordingMode.Stopped:
+                        if (_memoryFill > 0) {
+                            stop();
+                        }
+                }
+
+                //console.log(`Memory fill: ${_memoryFill}/${MAX_SAMPLES}, Playback: ${_playbackHead}/${_memoryFill} mode = ${_moduleMode}`)
```
kimprice, line 136: Delete this?

docs/static/packages/audio-recording/icon.png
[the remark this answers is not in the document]
srietkerk: When I put the image at `/docs/static/libs/[package].png` and change the icon path in `pxt.json`, the image doesn't show up for me at all. I also tried adding a dot to the front of the file path but no luck. Is there any reason why that is?

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 // Place your settings in this file to overwrite default and user settings.
 {
-    "files.autoSave": "afterDelay",
+    "files.autoSave": "off",
```
kimprice, line 3: This change probably shouldn't be checked in, right?

libs/audio-recording/recording.ts
```diff
@@ -0,0 +1,273 @@
...
+    High
+}
+
+enum AudioRecordingMode {
+    Stopped,
+    Recording,
+    Playing
+}
+
+enum AudioStatus {
```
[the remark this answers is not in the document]
srietkerk: Oh, interesting. I put my added enums with the others because I wanted to follow what they were already doing. I would think that just moving all of these enums into the namespace would be the best move, is there a reason we would want to keep some out?

docs/static/packages/audio-recording/icon.png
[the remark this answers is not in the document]
jwunderl: Possibly it didn't pick up the update to pxt.json & you need to stop / restart pxt serve?

libs/audio-recording/recording.ts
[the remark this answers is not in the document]
jwunderl: It does make the names a little longer in the javascript which might be annoying if they become , but that's the main reason. I think there might have been a blocks -> ts compiler issue like 4 years ago when enums got too deeply nested in namesapces? but I don't remember hitting that recently / it'd be obvious once you try it.

docs/static/packages/audio-recording/icon.png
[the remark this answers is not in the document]
srietkerk: I found out how to do this! Not sure what's going on underneath, but you do want to put images where you suggested `/docs/static/libs/[package].png`,  but you can't add that file path in pxt.json, something doesn't resolve well. I took a look at the other packages that keep their images in the `/libs` directory, and none of them have the `"icon":` line in their pxt.json, so something in our cli or otherwise is serving the image based on the fact that the image is named the same as the directory in which the library lives. In my case, the directory is `audio-recording`, so the image is `/docs/static/libs/audio-recording.png`.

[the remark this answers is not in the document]
srietkerk: I like the suggestion of the blocks, so this is what it looks like now.
image

libs/audio-recording/recording.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+#include "pxt.h"
+#include "MicroBit.h"
+#include "StreamRecording.h"
+
+using namespace pxt;
+
+namespace record {
+
+    static StreamRecording * recording = NULL;
```
kimprice, line 31: I'm not positive on this, but should it be written like this for readability?
```suggestion
    static StreamRecording* recording = NULL;
```

libs/audio-recording/recording.cpp
```diff
@@ -0,0 +1,143 @@
...
+        uBit.audio.deactivateMic();
+    }
+
+    void checkEnv() {
+        if( recording == NULL ) {
+            MicroBitAudio::requestActivation();
+            
+            recording = new StreamRecording( *uBit.audio.splitter );
+
+            MixerChannel * channel = uBit.audio.mixer.addChannel( *recording, 22000 );
```
kimprice, line 49: Same question here. 
```suggestion
            MixerChannel* channel = uBit.audio.mixer.addChannel( *recording, 22000 );
```

libs/audio-recording/recording.ts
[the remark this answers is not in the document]
srietkerk: Hmm, this is what happened when I moved the enums into the namespace:
image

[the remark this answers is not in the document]
jwunderl: Hm, that happens when a block's definition is broken, but we can definitely use enums in namespaces -- e.g. https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/game.ts#L11 and using it https://github.com/microsoft/pxt-common-packages/blob/master/libs/game/game.ts#L307 -- did you update all the blocks that use the types to make sure they're still referring it it?

libs/audio-recording/recording.ts
```diff
@@ -0,0 +1,273 @@
...
+    let _recordingFreqHz: number = 22000
+    let _playbackFreqHz: number  = 22000
+    let _micGain: AudioGainEnum  = AudioGainEnum.Medium
+
+    // Track if we have a simulator tick timer to use...
+    let _isSetup: boolean   = false
+    let _memoryFill: number = 0
+    let _playbackHead: number = 0
+
+    function __init__(): void {
```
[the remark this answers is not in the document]
srietkerk: I want to keep some of the underscores as I think it's a nice differentiator for some of the functions (init and setMode). I agree the double underscores are a bit much, so I just kept a single underscore between those that I mentioned, but I think increases readability a bit to know that those functions are solely for internal logic. There are some other functions that I am keeping in JavaScript that I'm not exporting currently because I think they're functionality that we might want to add as blocks later on.

libs/audio-recording/recording.ts
srietkerk: Hmm, yeah. I'm never a fan of while trues running when it's not needed. This evaded me a bit, that sounds like a good idea to me. I can do that as a follow-up PR.

.vscode/settings.json
srietkerk: Oh yeah, thanks for catching that!

libs/audio-recording/recording.cpp
```diff
@@ -0,0 +1,143 @@
...
+        uBit.audio.mic->enable();
+    }
+
+    void disableMic() {
+        uBit.audio.mic->disable();
+        uBit.audio.deactivateMic();
+    }
+
+    void checkEnv() {
+        if( recording == NULL ) {
```
kimprice, line 44: nit
```suggestion
        if (recording == NULL) {
```

libs/audio-recording/recording.cpp
```diff
@@ -0,0 +1,143 @@
...
+            MicroBitAudio::requestActivation();
+            
+            recording = new StreamRecording( *uBit.audio.splitter );
```
kimprice, line 47: nit: there are several instances of spacing like this around the parenthesis. I think this is different from our standard formatting. If you right click in vs code, you can "Format Document" and it should take care of the whitespace/styling.
```suggestion
            recording = new StreamRecording(*uBit.audio.splitter);
```

jwunderl (approved):
Nice job!

libs/audio-recording/recording.ts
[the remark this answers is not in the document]
srietkerk: I found out that, when moving the enums into the namespace, I needed to export them in order for the functions that use them to recognize them. I don't really understand why this is.  I was thinking maybe because if I export functions, then the types they mention should be accessible via a global space, but that doesn't really make sense as the closure the function and namespace creates should include those typings, so I'm at a bit of a loss as to why this happens.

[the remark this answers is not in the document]
srietkerk: I do like the idea from a design perspective, though, to have these enums in the namespace to supply that association, so I made the change.

PR #5025, opened 2023-03-24
update codeql

jwunderl:
Should make this happy again
image


jwunderl:
Just noticed this was still open, it doesn't matter much but just clears out a few errors that you can see if you open any of the codeql actions warning against using this old version

kimprice (approved):

Issue #5023, opened 2023-03-24
Recording blocks: on audio block should have options in lowercase


kimprice:
![image](https://user-images.githubusercontent.com/15070078/227602635-6ebbd2eb-5d58-4647-ba30-6846939f16bd.png)

kimprice:
This too:
![image](https://user-images.githubusercontent.com/15070078/227603621-dda374e8-bee4-4a3e-a48d-028ae4cb812a.png)

PR #5027, opened 2023-03-27
Audio blocks visual changes

srietkerk:
Changed the color of the blocks as requested by the foundation and made the drop-downs on the blocks with enum types lowercase. 

image

image

image

Closes https://github.com/microsoft/pxt-microbit/issues/5023

### DIFF
--- a/libs/audio-recording/_locales/audio-recording-strings.json
+++ b/libs/audio-recording/_locales/audio-recording-strings.json
@@ -1,9 +1,12 @@
 {
-  "record.AudioEvent.StartedPlaying|block": "Starts Playing",
-  "record.AudioEvent.StartedRecording|block": "Starts Recording",
-  "record.AudioEvent.StoppedPlaying|block": "Stops Playing",
-  "record.AudioEvent.StoppedRecording|block": "Stops Recording",
-  "record.AudioStatus.BufferFull|block": "out of space",
+  "record.AudioEvent.StartedPlaying|block": "starts playing",
+  "record.AudioEvent.StartedRecording|block": "starts recording",
+  "record.AudioEvent.StoppedPlaying|block": "stops playing",
+  "record.AudioEvent.StoppedRecording|block": "stops recording",
+  "record.AudioRecordingMode.Playing|block": "playing",
+  "record.AudioRecordingMode.Recording|block": "recording",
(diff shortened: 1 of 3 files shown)


eanders-ms (approved):

libs/audio-recording/_locales/audio-recording-strings.json
```diff
@@ -1,17 +1,20 @@
...
+  "record.AudioStatus.BufferFull|block": "full",
   "record.AudioStatus.Playing|block": "playing",
   "record.AudioStatus.Recording|block": "recording",
   "record.AudioStatus.Stopped|block": "stopped",
   "record.audioEvent|block": "on audio $eventType",
   "record.audioStatus|block": "audio is $status",
   "record.playAudio|block": "play recording",
   "record.setMicGain|block": "set microphone sensitivity to $gain",
-  "record.startRecording|block": "record audio for 3(s)",
+  "record.startRecording|block": "record audio clip",
```
kimprice, line 17: what prompted this change?

kimprice (approved):
Looks good! Love the new color, but how does it look against the rest of the toolbox? Maybe include a picture in this PR description?

kimprice:
The mic icon is a bit hard to see. I wonder if you can add some kind of shadow or something to improve the contrast with the background.

srietkerk:
> Looks good! Love the new color, but how does it look against the rest of the toolbox? Maybe include a picture in this PR description?

Here is what the color looks like in  the toolbox!

image

image

srietkerk:
> The mic icon is a bit hard to see. I wonder if you can add some kind of shadow or something to improve the contrast with the background.

I agree; other categories with dark block colors have this problem as well. If you open the images and led blocks in micro:bit, you have a similar problem. One option is to just remove the icon completely; the Serial category in advanced does this. I'd just have to dig in a bit as to how they do it.

libs/audio-recording/_locales/audio-recording-strings.json
srietkerk: Oh, when meeting with the foundation, they didn't like the idea of stating a time since time elapsed is not really what constrains the recording length and can change depending on the set quality of the audio (a block I will likely be adding back in). 

Clip was suggested because it implies a short amount of time while also not explicitly stating how much time the user will have.

Issue #4941, opened 2023-02-10
play sound block consumes ~48 KBs of 


microbit-carlos:
**Describe the bug**

Breaking out this specific issue from:
- https://github.com/microsoft/pxt-microbit/issues/4934

Looking at the `binary.js` size numbers from this programme, the size difference between removing or including the `[play sound]` block is ~48 KBs, which is quite a lot for something that would be mostly wrapping CODAL functionality.

https://makecode.microbit.org/_bfs6mzETmeUi

<img width="783" alt="image" src="https://user-images.githubusercontent.com/147113/217794697-a92d9106-5dad-4fb3-896d-aca54e04f746.png">

**To Reproduce**
- Load the programme above
- Looking at the `binary.js` file the `total` value is `166831`
- Remove the `play sound` block
- Looking at the `binary.js` file the `total` value is `118995`

**Expected behavior**
A lower memory footprint for the block.

**Screenshots**
N/A

**micro:bit version (please complete the following information):**

N/A

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**

makecode.microbit.org version:  [5.0.12](https://github.com/Microsoft/pxt-microbit/releases/tag/v5.0.12)
Microsoft MakeCode version:  [8.0.7](https://github.com/Microsoft/pxt/releases/tag/v8.0.7)
microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
codal-microbit-v2 runtime version:  [v0.2.40](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.40)

abchatra:
@riknoll any idea

abchatra:
@riknoll can you take a look at this? This came up from ElecFreaks accessory maker again.

riknoll:
sure thing

riknoll:
quick note: the 48kb number refers to binary.js size and not the compiled code size. To get an accurate number, you need to download the hex file and open the mbcodal.asm file that appears under the built folder. You can also add ?compiler=size to the url to get more detailed size information in that file.

PR #5039, opened 2023-03-31
Improve createSoundEffect code size

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/4941

Avoiding using the Sound class with all it's getters/setters. It's a nice API, but I guess it generated more code than I expected. The class is still there, but it's not used in this function anymore.

To get much of an improvement beyond this, we'd probably need a new API from codal instead of the string API. Would be great if we could pass the args in a byte array or something so that we don't have to deal with all of the string formatting. I know @jaustin was talking about getting rid of the string representation at some point...

For this program:

https://makecode.microbit.org/_bwgdsdXPbCcK

Code size before this change:
```
generated code sizes (bytes): 7424 (incl. 5624 user, 946 helpers, 450 vtables, 404 lits)
```

Code size after this change:
```
generated code sizes (bytes): 4016 (incl. 2934 user, 808 helpers, 110 vtables, 164 lits)
```

One thing to note: the linked issue mentions 48kb of data, which isn't quite accurate because that's the binary.js size and not the compiled code size. To get an accurate number, you need to download the hex file and open the mbcodal.asm file that appears under the built folder. You can also add `?compiler=size` to the url to get more detailed size information in that file.

### DIFF
--- a/libs/core/soundexpressions.ts
+++ b/libs/core/soundexpressions.ts
@@ -368,46 +368,58 @@ namespace music {
     //% expandableArgumentBreaks="3,5"
     //% group="micro:bit (V2)"
     export function createSoundEffect(waveShape: WaveShape, startFrequency: number, endFrequency: number, startVolume: number, endVolume: number, duration: number, effect: SoundExpressionEffect, interpolation: InterpolationCurve): string {
-        const sound = new soundExpression.Sound();
-        sound.wave = waveShape;
-        sound.frequency = startFrequency;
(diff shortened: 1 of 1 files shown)


jwunderl (approved):

abchatra:
@riknoll after this change, does **https://github.com/microsoft/pxt-microbit/issues/4941** doesn't repro?

riknoll:
@abchatra correct. that linked program compiles with no issue. That doesn't mean it can't be improved, but like I said we would probably want to do a change in codal for that.

PR #5041, opened 2023-04-04
Audio Blocks: Adding Sample Rate

srietkerk:
In this PR:
- Recording object now updated to work with the new codal version.

- Added block to change the sample rate with a slider from 1000 samples to 22000 samples per second - an important block to have for the foundation. 

- Added a getter for the output sample because the input sample rate is a bit trickier to access. 

- Commented out the `_init()` function for now in all functions in `recording.ts` to play with the consequences of the audio calls being more synchronous. This change has created some bumps, that I am going to smooth over in a follow-up PR, but the first time you run your program and record, the quality should be decent enough that you can understand the words being said.

Improvements to be made: 
- When you record and playback, there is a very high-pitched frequency that interferes with the audio you recorded. I am not sure what this is. This still applies when the microphone sensitivity is set to low.
- Some refactoring in the TS and between the TS and cpp. I know there are places to clean up the code a bit to make it more efficient and cleaner, I just need to make the adjustments. 

**UPDATE**
Added block ids, changed the order of the blocks a bit. 
This is the block ids of those that are currently in beta:
image

Updated order of the blocks:
image


jwunderl (commented):
few nitpicks, I would suggest adding in blockIds now before we forget about it too if you have time (can be separate pr) -- we can probably set them to whatever we want if we're not afraid of people losing programs (and since this is a new extension / only in beta / only really used by us so far that's probably fine).

But if you wanna make them match you can put the blocks into a project, add the `saveblocks=1` query parameter (e.g. `https://makecode.microbit.org/beta?saveblocks=1#editor`), and press the save button to download the .blocks xml file to see what the ids are: 
image
(it'll be the 'type' on the block).

You can also check the source when we are loading it (I believe it's here: https://github.com/microsoft/pxt/blob/master/pxtlib/service.ts#L594 for functions, and we rarely override the blockid for enums / etc but that would be in the same file) but that's a little bit of a pain :)

libs/audio-recording/_locales/audio-recording-strings.json
```diff
@@ -10,10 +10,11 @@
   "record.AudioStatus.Playing|block": "playing",
   "record.AudioStatus.Recording|block": "recording",
   "record.AudioStatus.Stopped|block": "stopped",
-  "record.audioEvent|block": "on audio $eventType",
   "record.audioStatus|block": "audio is $status",
-  "record.playAudio|block": "play recording",
+  "record.getOutSampleRate|block": "playback sample rate",
+  "record.playAudio|block": "play audio clip",
   "record.setMicGain|block": "set microphone sensitivity to $gain",
+  "record.setSampleRate|block": "set sample rate to $hz || for $scope",
```
kimprice, line 17: Is scope where you pick recording, playback, or both?

libs/audio-recording/_locales/audio-recording-strings.json
```diff
@@ -10,10 +10,11 @@
   "record.AudioStatus.Playing|block": "playing",
   "record.AudioStatus.Recording|block": "recording",
   "record.AudioStatus.Stopped|block": "stopped",
-  "record.audioEvent|block": "on audio $eventType",
```
kimprice, line 13: Why remove this block?

libs/audio-recording/recording.cpp
```diff
@@ -29,6 +29,8 @@ using namespace pxt;
 namespace record {
 
 static StreamRecording *recording = NULL;
+static SplitterChannel *splitterChannel = NULL;
+static MixerChannel *channel = NULL;
```
kimprice, line 33: What is mixer and splitter in layman's terms?

libs/audio-recording/_locales/audio-recording-strings.json
srietkerk: Yes! The default is everything.

libs/audio-recording/_locales/audio-recording-strings.json
srietkerk: The foundation questioned its value since we have the booleans. They pointed out that the use case will be that users will press a button to start and stop playing, so it doesn't make a ton of sense to have the event blocks. When I chatted with everyone about the blocks, I think Thomas brought this up as well. Since I can understand how it doesn't bring a lot of value, I decided to go with their recommendation.

libs/audio-recording/recording.cpp
srietkerk: It's still a bit fuzzy for me, and I'm not sure if I can really simply it much, but I might be able to explain it. As far as I understand it, the splitter takes the input from the mic and puts it into it's own data stream. I think this makes it so the micro:bit can receive multiple different inputs from different channels so things can be more parallel? I could be wrong. 

The mixer channel takes that data from the splitter and outputs it to the speaker.

TL;DR from what I gather from the code, the splitter channel handles audio input and the mixer channel handles audio output, but I could be wrong.

kimprice: oh ok thanks!

libs/audio-recording/recording.cpp
```diff
@@ -146,4 +153,34 @@ bool audioIsRecording() {
...
+//%
+void setInputSampleRate(int sampleRate) {
+        if (splitterChannel == NULL)
+            splitterChannel = uBit.audio.splitter->createChannel();
+        splitterChannel->requestSampleRate(sampleRate);
+}
+
+
+/**
+ * Change the sample rate of the mixer channel (audio input)
```
kimprice, line 179: Why does the comment say `audio input` if this one has to do with output?

libs/audio-recording/recording.cpp
```diff
@@ -104,13 +109,13 @@ void erase() {
 void setMicrophoneGain(int gain) {
     switch (gain) {
     case 1:
-        uBit.audio.processor->setGain(0.1);
+        uBit.audio.processor->setGain(0.079);
```
jwunderl, line 112: is `uBit.audio.processor->setGain` intended to pass a 0-1 value to regard as percentage? If that's the case it feels like we should double check with them these are roughly the 'intended' values for good input if we haven't (after this pr / non blocking) / what they would expect a good range to be -- if it's as expected that's totally fine, but just to make sure we're in the same ballpark as they expect

jwunderl (approved):

libs/audio-recording/recording.ts
```diff
@@ -238,21 +232,50 @@ namespace record {
...
+                break;
+
+            case AudioSampleRateScope.Playback:
+                setOutputSampleRate(hz);
+                break;
+
+            case AudioSampleRateScope.Recording:
+                setInputSampleRate(hz);
+                break;
+
```
kimprice, line 273: You could put the `Everything` case here and then just have it fall through to the default case. Then just remove it from the beginning of the switch statement. This would make it clearer that the default is the `Everything` case.
```suggestion
case AudioSampleRateScope.Everything:
```

libs/audio-recording/_locales/audio-recording-jsdoc-strings.json
```diff
@@ -1,17 +1,20 @@
 {
   "record": "Functions to operate the v2 on-board microphone and speaker.",
   "record.audioDuration": "Get how long the recorded audio clip is",
-  "record.audioEvent": "Do something based on what the audio is doing",
   "record.audioIsPlaying": "Get whether the playback is active",
   "record.audioIsRecording": "Get whether the microphone is listening",
   "record.audioIsStopped": "Get whether the board is recording or playing back",
   "record.audioStatus": "Test what the audio is doing",
   "record.erase": "Clear the buffer",
+  "record.getInputSampleRate": "Get the sample rate of the splitter channel (audio input)",
```
kimprice, line 9: Is the reason there is no `record.getSampleRate` here because it's not a block?

srietkerk: Yep! And you pointing this out made me realize that I should really just get rid of this in the C++, I don't want this surfaced in the TypeScript when there's not symmetry with the other option.

libs/audio-recording/recording.cpp
```diff
@@ -146,4 +151,32 @@ bool audioIsRecording() {
 bool audioIsStopped() {
     return recording->isStopped();
 }
+
+/**
+ * Change the sample rate of the splitter channel (audio input)
+ */
+//%
+void setInputSampleRate(int sampleRate) {
+    checkEnv();
```
kimprice, line 160: Is this potentially resetting the output sample rate to the default if the user previously set it to something else? I'm wondering what the need is for this call.

srietkerk: You're right, this could definitely happen if the code was written a bit differently. I was just completely glazing over this before, but the logic in `checkEnv()` will actually will ever be run once in a program using the audio blocks. All of the logic in `checkEnv()` is inside the `if (recording == NULL)` condition, and if `checkEnv()` has been called once already, `recording` will not be null. I was always just testing the blocks with setting the sample rate before using other audio blocks. If I set the output sample rate before setting the input sample rate, calling `checkEnv()` in `setInputSampleRate` actually wouldn't do anything because `recording` is not null since we called `setOutputSampleRate()` and set recording in the call there. 

To answer the question as to why this call is needed, I put the call there as a way to check that `splitterChannel` would get initialized before trying to set its sample rate. I did try just putting a null check in the `setInputSampleRate()` function itself (`if splitterChannel == NULL, create splitter channel`), but I was running into weirdness that I was confused by before but I now understand what it was. Because I was setting the input sample rate before `checkEnv()` was ever called, and thus before initializing `recording`, when `checkEnv()` did end up getting called when recording started, the input sample rate was lost!

TL;DR this would cause problems if the logic in `checkEnv()` was not wrapped in the `if (recording == NULL)` condition.

srietkerk: All that being said, this helped me realize a bug that I just thought was me not understanding how sample rate works. When I just wanted to change the sample rate for output, I tried making a program like the following, but I could never tell a difference between the default and what I changed it to.

image

This was because, since `recording` was already initialized, the call that I make to `checkEnv(sampleRate)` in `setOutputSampleRate()` wouldn't actually _do anything_ so it was always just the default.  

Thank you for helping me find this!!

libs/audio-recording/recording.ts
```diff
@@ -168,91 +171,105 @@ namespace record {
      * Record an audio clip for a maximum of 3 seconds
      */
     //% block="record audio clip"
+    //% blockId="record_startRecording"
     //% weight=70
     export function startRecording(): void {
-        _init()
+        // _init()
```
abchatra, line 177: Please remove the commented code here. They don't serve value.

abchatra: I am also confused, who calls the init?

abchatra:
@JohnVidler  can you please review this PR? @srietkerk  had few questions too.

libs/audio-recording/recording.ts
srietkerk: This was answered in the meeting with the foundation this morning, but I'll put the info here for future use. I was keeping `init()` around because I didn't want to fully get rid of code until I knew that it was no longer valuable. Having `init()` in the TypeScript was important without the codal updates. Now that the codal updates are applied, we can rely more on the C++ logic.

libs/audio-recording/recording.cpp
srietkerk: I couldn't get a super clear answer here, but it sounded like it should be a value somewhere between 0-1. The values that I landed on at this point I think make the most sense in terms of low, medium, and high sensitivity.

Issue #4890, opened 2022-12-01
LED plot bar graph - disable serial output


martinwork:
**Is your feature request related to a problem? Please describe.**

Arising from support ticket https://support.microbit.org/helpdesk/tickets/58910 (private)

> I would like to request a way to disable the feature (99% helpful) that sends a serial message every time some data is plotted via the LED's. My students would frequently like the micro:bit to show a graph on its display, but send other data via serial, and this interferes with their projects.

**Describe the solution you'd like**
> A simple block that could toggle it on/off would be really handy.

**Describe alternatives you've considered**

Using Javascript: `led.barGraphToConsole = false`.
Using an extension with alternative bar graph blocks.

abchatra:
We can add an optional parameter to switch this off.

martinwork:
From the ticket: 
> Perfect, thank you!

Thanks @abchatra!

PR #5053, opened 2023-04-18
Add optional console log parameter to plotBarGraph

kimprice:
closes https://github.com/microsoft/pxt-microbit/issues/4890

### DIFF
--- a/libs/core/led.ts
+++ b/libs/core/led.ts
@@ -21,10 +21,10 @@ namespace led {
     let barGraphHigh = 0;
     // when was the current high value recorded
     let barGraphHighLast = 0;
-    
+
     /**
      * Controls where plotbargraph prints to the console
-     **/ 
+     **/
     export let barGraphToConsole = true
 
     /**
@@ -32,13 +32,17 @@ namespace led {
      * If `high` is 0, the chart gets adjusted automatically.
      * @param value current value to plot
      * @param high maximum value. If 0, maximum value adjusted automatically, eg: 0
+     * @param valueToConsole if true, prints value to the console
      */
     //% help=led/plot-bar-graph weight=20
-    //% blockId=device_plot_bar_graph block="plot bar graph of %value up to %high" icon="\uf080" blockExternalInputs=true
+    //% blockId=device_plot_bar_graph block="plot bar graph of %value up to %high|| console log %barGraphToConsole" icon="\uf080" blockExternalInputs=true
     //% parts="ledmatrix"
-    export function plotBarGraph(value: number, high: number): void {
+    export function plotBarGraph(value: number, high: number, valueToConsole?: boolean): void {
+        if (valueToConsole == undefined){
+            valueToConsole = barGraphToConsole;
+        }
         const now = input.runningTime();
-        if (barGraphToConsole)
+        if (valueToConsole)
             console.logValue("", value);
         if (isNaN(value)) {
             basic.clearScreen()
--- a/libs/core/led.ts
+++ b/libs/core/led.ts
@@ -28,17 +28,6 @@ namespace led {
     export let barGraphToConsole = true
 
     /**
-     * Sets whether the bar graph values print to the console.
-     * @param valueToConsole if true, prints values to the console
-     */
-    //% help=led/set-bar-graph-to-console weight=20
-    //% blockId=device_set_bar_graph_to_console block="log bar graph values to console $valueToConsole" icon="\uf080"
-    //% valueToConsole.shadow=toggleOnOff
-    export function setBarGraphToConsole(valueToConsole: boolean): void {
-        barGraphToConsole = valueToConsole;
-    }
-
-    /**
      * Displays a vertical bar graph based on the `value` and `high` value.
      * If `high` is 0, the chart gets adjusted automatically.
      * @param value current value to plot
--- a/libs/core/led.ts
+++ b/libs/core/led.ts
@@ -35,7 +35,7 @@ namespace led {
      * @param valueToConsole if true, prints value to the console
      */
     //% help=led/plot-bar-graph weight=20
-    //% blockId=device_plot_bar_graph block="plot bar graph of %value up to %high|| console log %valueToConsole" icon="\uf080" blockExternalInputs=true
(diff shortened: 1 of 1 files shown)


libs/core/led.ts
```diff
@@ -21,24 +21,28 @@ namespace led {
...
     /**
      * Displays a vertical bar graph based on the `value` and `high` value.
      * If `high` is 0, the chart gets adjusted automatically.
      * @param value current value to plot
      * @param high maximum value. If 0, maximum value adjusted automatically, eg: 0
+     * @param valueToConsole if true, prints value to the console
      */
     //% help=led/plot-bar-graph weight=20
-    //% blockId=device_plot_bar_graph block="plot bar graph of %value up to %high" icon="\uf080" blockExternalInputs=true
+    //% blockId=device_plot_bar_graph block="plot bar graph of %value up to %high|| console log %valueToConsole" icon="\uf080" blockExternalInputs=true
```
srietkerk, line 38: Not sure if we want to mix the two types of parameter definition for blocks, but if we do, I think it should be `$valueToConsole` rather than `%valueToConsole`.

kimprice: Yea I wasn't sure whether to change it or whether we just keep the old code the way it is and only use `$` for new block definitions. @riknoll @jwunderl?

libs/core/led.ts
```diff
@@ -21,24 +21,28 @@ namespace led {
...
     //% parts="ledmatrix"
-    export function plotBarGraph(value: number, high: number): void {
+    export function plotBarGraph(value: number, high: number, valueToConsole?: boolean): void {
+        if (valueToConsole == undefined){
+            valueToConsole = barGraphToConsole;
+        }
```
kimprice, line 43: Oh should I actually have an `else` case and set `barGraphToConsole = valueToConsole`?

jwunderl: I would not expect one call's optional parameter to affect other calls, so I'd say no.

libs/core/led.ts
jwunderl, line 38:
```suggestion
    //% blockId=device_plot_bar_graph block="plot bar graph of $value up to $high|| console log $valueToConsole" icon="\uf080" blockExternalInputs=true
```

if we're changing the block definition anyways we should change them to $ -- the only reason we left them as is is to avoid messing up the translations, and if we're changing the string the translations won't carry over anyways

libs/core/led.ts
```diff
@@ -21,24 +21,28 @@ namespace led {
...
      * Displays a vertical bar graph based on the `value` and `high` value.
      * If `high` is 0, the chart gets adjusted automatically.
      * @param value current value to plot
      * @param high maximum value. If 0, maximum value adjusted automatically, eg: 0
+     * @param valueToConsole if true, prints value to the console
      */
     //% help=led/plot-bar-graph weight=20
-    //% blockId=device_plot_bar_graph block="plot bar graph of %value up to %high" icon="\uf080" blockExternalInputs=true
+    //% blockId=device_plot_bar_graph block="plot bar graph of %value up to %high|| console log %valueToConsole" icon="\uf080" blockExternalInputs=true
     //% parts="ledmatrix"
```
jwunderl, line 39:
```suggestion
    //% parts="ledmatrix"
    //% valueToConsole.shadow=toggleOnOff
```

? to make it an 'on/off' toggle instead of `true/false`

jwunderl (commented):
Whenever we change blocks we need to make sure they update nicely; trying importing https://makecode.microbit.org/_ACCRJ7i4A7pc into a build with these changes and see how it ends up. My guess / if I'm remembering right is that valueToConsole will be an empty slot instead of having a default, so we'd need to either add an upgrade rule to account for that or deprecate this block and add a new one with the new optional parameter~

libs/core/led.ts
jwunderl: Oh, this comment didn't show up when I was reviewing, response below https://github.com/microsoft/pxt-microbit/pull/5053#discussion_r1170349902

libs/core/led.ts
jwunderl: I'd see that as a reason to make a `set bar graph to console [on/off]`, but just doing it as a side effect feels really weird to me

kimprice: @abchatra @Jaqster should we add another block?

kimprice: Added the block above for now, but can remove. Let me know.

abchatra: My opinion is not to add another bock. This is for advanced users and if they want they will use in each block

libs/core/led.ts
```diff
@@ -21,24 +21,40 @@ namespace led {
...
     export let barGraphToConsole = true
 
+    /**
+     * Sets whether the bar graph values print to the console.
+     * @param valueToConsole if true, prints values to the console
+     */
+    //% help=led/set-bar-graph-to-console weight=20
+    //% blockId=device_set_bar_graph_to_console block="log bar graph values to console $valueToConsole" icon="\uf080"
+    //% valueToConsole.shadow=toggleOnOff
+    export function setBarGraphToConsole(valueToConsole: boolean): void {
```
abchatra, line 37: Lets not add new blocks for this.

jwunderl:
Can you check on https://github.com/microsoft/pxt-microbit/pull/5053#pullrequestreview-1390591301 and see how it shows up when loading a previously made project with this block in it?

kimprice:
> Whenever we change blocks we need to make sure they update nicely; trying importing https://makecode.microbit.org/_ACCRJ7i4A7pc into a build with these changes and see how it ends up. My guess / if I'm remembering right is that valueToConsole will be an empty slot instead of having a default, so we'd need to either add an upgrade rule to account for that or deprecate this block and add a new one with the new optional parameter~

There was no problem when I imported on a local build. Would it possibly be different if I create an upload target?

srietkerk (approved):
As long as what Joey asked about the changes not breaking the block is all good, everything looks good to me!

jwunderl (approved):
Should be fine then, we must have done a fix at some point to make it automatically populate shadow if it's missing

PR #5055, opened 2023-04-19
Recording Extension: using a local stream normalizer to change recording mic gain

srietkerk:
**Context**
Before, when setting mic gain for the microphone, the microphone would be effected globally. Therefore, if someone changed the sensitivity of the mic with the audio extension and tried to use a block like `onLoudSound` the threshold would react differently. We don't want to have this side effect, so we are now using a local `StreamNormalizer` in the data pipeline that will only effect the mic for recording. 

**NOTE**
There is still a bug in the code that exists currently, but I wanted to get some progress in. Right now, if you had a program like this 
image
and you tried to press 'A' a second time, nothing will happen.

I have an inkling that the issue lies with the way that the `StreamRecording` object connects and disconnects to the upstream data source. For context, [this](https://github.com/lancaster-university/codal-core/blob/master/source/streams/StreamRecording.cpp#L139) is where I'm looking. This code should, whenever `record()` in C++ is called, trigger the upstream source and propagate to the mic (the recent changes they did in codal with on-demand data streams). However, this isn't happening. I haven't been able to pin point why; I'm hoping to have another chat with @JohnVidler to fix this. 

That being said, this code works fine upon download or if you just reset the micro:bit program with the button on the back. Because this code is a step in the right direction, I wanted to get a PR up.

### DIFF
--- a/libs/audio-recording/recording.cpp
+++ b/libs/audio-recording/recording.cpp
@@ -31,17 +31,7 @@ namespace record {
 static StreamRecording *recording = NULL;
 static SplitterChannel *splitterChannel = NULL;
 static MixerChannel *channel = NULL;
-
-void enableMic() {
-    uBit.audio.activateMic();
-    uBit.audio.mic->enable();
-}
-
-void disableMic() {
-    uBit.audio.mic->disable();
-    uBit.audio.deactivateMic();
-}
-
+static StreamNormalizer *localProcessor = NULL;
 
 void checkEnv(int sampleRate = -1) {
     if (recording == NULL) {
@@ -50,8 +40,8 @@ void checkEnv(int sampleRate = -1) {
         MicroBitAudio::requestActivation();
 
         splitterChannel = uBit.audio.splitter->createChannel();
-
-        recording = new StreamRecording(*splitterChannel);
+        localProcessor = new StreamNormalizer(*splitterChannel);
+        recording = new StreamRecording(localProcessor->output);
 
         channel = uBit.audio.mixer.addChannel(*recording, sampleRate);
 
@@ -72,7 +62,6 @@ void checkEnv(int sampleRate = -1) {
 //% promise
 void record() {
     checkEnv();
-    enableMic();
     recording->record();
 }
 
@@ -82,7 +71,6 @@ void record() {
 //%
 void play() {
     checkEnv();
-    disableMic();
     recording->play();
 }
 
@@ -92,7 +80,6 @@ void play() {
 //%
 void stop() {
     checkEnv();
-    disableMic();
     recording->stop();
 }
 
@@ -102,7 +89,6 @@ void stop() {
 //%
 void erase() {
     checkEnv();
-    disableMic();
     recording->erase();
 }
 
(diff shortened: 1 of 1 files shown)


libs/audio-recording/recording.cpp
```diff
@@ -111,15 +97,16 @@ void erase() {
  */
 //%
 void setMicrophoneGain(int gain) {
+    checkEnv();
     switch (gain) {
     case 1:
-        uBit.audio.processor->setGain(0.079f);
+        localProcessor->setGain(0.5f);
```
kimprice, line 103: Why did the gain values change?

libs/audio-recording/recording.cpp
```diff
@@ -50,8 +40,8 @@ void checkEnv(int sampleRate = -1) {
         MicroBitAudio::requestActivation();
 
         splitterChannel = uBit.audio.splitter->createChannel();
-
-        recording = new StreamRecording(*splitterChannel);
+        localProcessor = new StreamNormalizer(*splitterChannel);
+        recording = new StreamRecording(localProcessor->output);
 
         channel = uBit.audio.mixer.addChannel(*recording, sampleRate);
```
kimprice, line 47: Do you need a case below for when `recording != null && sampleRate == -1`? Maybe I'm not seeing it, but I don't think you are covering that case.

libs/audio-recording/recording.cpp
srietkerk: I don't know how the different channels impact the gain, but when I had the values near 0.079 with the new Normalizer, it was extremely quiet , so I changed the values back to higher floats and it sounds louder and more noticeable. These are values that can be fiddled with forever, but I think the values I landed on are good.

srietkerk:
> If checkEnv is called when recording is NULL, I think this will always be true, so 2 channels will initially be added to the mixer.
> 
> https://github.com/microsoft/pxt-microbit/blob/69608fd7385e073ffcccc23e0a213348f8b6470b/libs/audio-recording/recording.cpp#L53
> 
> 
> Further calls with sampleRate != -1, will each add another mixer channel. Each extra channel addition is a memory leak.
> https://github.com/lancaster-university/codal-microbit-v2/blob/master/source/Mixer2.cpp#L92
> This program currently causes panic 020 after 20s or so. image

Ah, that's bad. Thank you for pointing this out, @martinwork. 

A solution here would be to just set the global mixer's sample rate instead of the created channel's, like using `uBit.audio.mixer.setSampleRate(sampleRate);` in `setOutputSampleRate` instead of having all the logic in `checkEnv()`.  However, this would mean that the sample rate is changed globally. As I don't know much about microphones and sound, I'm unsure how this will affect the other sound blocks or if it will at all. I'd be up to make this change if the effects on the other blocks is minimal or non-existent. 

I'm not sure how this would work of if this is even possible, but I've also wondered what it would look like if we surfaced a `setSampleRate()` in the `MixerChannel`. Currently, the only things public for the `MixerChannel` are itself and the ability to change the channel's volume, but it has data on a sample rate (https://github.com/lancaster-university/codal-microbit-v2/blob/master/inc/Mixer2.h#L62). What would it look like to set the sample rate in the `MixerChannel`? Would it make everything too distorted? I know making a change to codal would be more involved, but it's just a thought.

abchatra:
What is the status of this PR?

PR #5057, opened 2023-04-19
Recording extension: using blocks now has sim indicators

srietkerk:
Before, when putting the extension blocks into the editor, no changes would happen to the simulator. Now, we show that V2 is needed and that the microphone will be used. 

When the `record audio clip` block is activated, the microphone indicator will light up red like it does with the other microphone-centered blocks. 

When the blocks are in the editor:
image

When recording: 
image

TODO:
- Make it so the microphone stops lighting up when it is not listening
- Make the browser actually prompt the user to use the in-computer mic to record
- Playback recorded media when the `play audio clip` block is active

Last note: I'm not sure if setting the sample rate and setting the mic gain is something that we'd be able to simulate.. Is there any way to communicate this to users without changing the audio quality?


libs/core/_locales/core-strings.json
```diff
@@ -356,7 +356,7 @@
   "input|block": "input",
   "led.brightness|block": "brightness",
   "led.enable|block": "led enable %on",
-  "led.plotBarGraph|block": "plot bar graph of %value up to %high",
+  "led.plotBarGraph|block": "plot bar graph of $value up to $high|| console log $valueToConsole",
```
kimprice, line 359: Oh this change and the one above are from one of my recent PRs. What did I need to do to update these strings?

srietkerk: I'm not sure why this got updated on my side. I think the strings get updated when you run a build.

abchatra:
>> Last note: I'm not sure if setting the sample rate and setting the mic gain is something that we'd be able to simulate.. Is there any way to communicate this to users without changing the audio quality?

I would skip these things.

jwunderl (approved):
fixes https://github.com/microsoft/pxt-microbit/issues/5073

PR #5058, opened 2023-04-19
Update reference of 'plotBarGraph' for new parameter

### LINKED ISSUE #5056: Update help doc for plotBarGraph
An optional parameter was added to turn off/on logging the values. 
See changes here: https://github.com/microsoft/pxt-microbit/pull/5053/files

ganicke:
Update the reference page for **led.plotBarGraph()** to include the parameter description for `valueToConsole`. Plus, a few more edits.

Closes #5056

## Suggestions

* Since `console` is still directed to serial output, I'm thinking we might use 'serial' instead of 'console' in reference to this parameter. I did that in the doc update.

See [line 398](https://github.com/microsoft/pxt-microbit/blob/bf61883bd5e660e4dc3e9142b3560d73fcfc39b3/libs/core/control.cpp#L398) in control.cpp as `control.__log()` (called thru `console.log()` et al) sends to serial:

```typescript
    /**
    *
    */
    //%
    void __log(int priority, String text) {
        if (NULL == text) return;
        pxt::sendSerial(text->getUTF8Data(), text->getUTF8Size());
    }
```

Also, the actual `console` namespace is yet to be exposed:

image

* I modified the text for the block to use 'serial write' rather than 'console log' due to the reasons above.

* It seems that `barGraphToConsole` has no utility anymore so I removed that and set `valueToConsole` to default to `false`. This seems to have the same effect.

You can hate on any of these changes if you want. They are just suggestions.

### DIFF
--- a/docs/reference/led/plot-bar-graph.md
+++ b/docs/reference/led/plot-bar-graph.md
@@ -6,7 +6,7 @@ Display a bar graph for a number value.
 led.plotBarGraph(2, 20);
 ```
 
-A bar graph is a kind of chart that shows numbers as lines with different lengths.
+A bar graph is a kind of chart that shows numbers as lines with different lengths. The value is plotted in LEDs as a ratio of a **value** to the **high** value you set as a maximum range. So, if there are 25 LEDs on the screen, then plotting `9` for the high value of `50` would display about 5 LEDs on the screen.
 
 ## Parameters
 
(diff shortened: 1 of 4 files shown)


jwunderl (commented):
I like changing it to serial, maybe `|| mirror to serial`` to match datalogger block?

image

We need to keep `barGraphToConsole` to avoid breaking programs as kim said, but it should be fine to leave it as is even if the block says serial; it's typescript anyway / advanced / we conflate the two in typescript already

kimprice (approved):
Looks good!

kimprice:
I should've done this in my PR, but could you make the default "on" for serial write?

Issue #5062, opened 2023-04-21
Compile error on sample rate


Jaqster:
Compile exception gets thrown when I set the sample rate value - 
![image](https://user-images.githubusercontent.com/22108954/233698380-22bca90f-16c7-4dc9-8190-938518fefea8.png)

Issue #5006, opened 2023-03-24
Need some indicator in sim when audio is supposed to be recording


kiki-lee:
Using the sim, I did expect to be able to record and play back audio from my computer for the sim, but I couldn't tell if/when audio was actually recording for my program.  There's no indication in the sim (lights, text, flashing "on air" sign...) to let me know that block is running. 

![Screenshot by Dropbox Capture](https://user-images.githubusercontent.com/1633958/227590586-19e7ab7a-7f78-4bbc-8fdc-f78bf2db6d1a.png)

Issue #4610, opened 2022-04-29
Allow simulator to behave correctly when used with audio play/record extensions using new CODAL pipeline


jaustin:
**Is your feature request related to a problem? Please describe.**

There's a new audio pipeline going into CODAL that will allow, amongst other things, playback and record of audio, and some more dynamic behaviour for the audio pipeline. For example, if a user 'listens' for 3 seconds to store a clip in RAM, the microphone will only be powered for those 3 seconds, and consequently the micro:bits microphone LED will only be on during that time. 

As yet, the simulator doesn't implement anything relating to these new CODAL APIs and the recording LED, nor the ability to record sound.

We're developing these APIs as an extension, but would like the simulator to do 'the right thing' so users don't get vastly different experiences on the simulator and the device.

**Describe the solution you'd like**

Ultimately, I'd like the simulator to be able to record audio and play it back. I understand that may be a longer term goal.

However, in the short term, it may suffice for the simulator's microphone LED to activate and deactivate at appropriate times.
* Should this be under the control of the simulator
* Or should we push the responsibility to the extension to have a different codepath under simulation that is responsible for faking microphone LED state? 
* If the latter, is that API already exposed to the extensions?

**Describe alternatives you've considered**

* Doing nothing:
We think this will be confusing for users and make it hard for us to user test the recording blocks

* Having a fixed sample that is always 'recorded' and played back
I think this would work quite nicely but in conjunction with #4609 could be implemented 'extension side' not necessarily in MakeCode itself. (This would allow user changes of sample rate etc to be experimented with even in absence of real recording

@riknoll @johnvidler has more information on the pipeline stuff.

abchatra:
This is asking for simulator support for record playback api

PR #5082, opened 2023-04-27
Audio extension: using computer mic and speakers for recording, playback in simulator

srietkerk:
If a user does not have a micro:bit, we want their experience with the audio extension to be as close to the real thing as possible.

In the simulator, if you are actively recording audio, the browser will now request to use your microphone. It will record data from the user's computer microphone for approximately four seconds. Currently, it is a fixed number. I chose four seconds because I believe this is a close median to how many seconds you can record on the micro:bit. The sim indicators have also updated to reflect this functionality. The microphone indicator will only be illuminated if the sim is actively recording, just like it is on the board. 

If the user plays the audio, the clip will come through their speakers. 

This PR has also updated the booleans in the simulator. They actually represent whether audio is playing, stopped, or recording (full was already fine because that implementation is just in TypeScript).

The sample rate functions were also added to the sim state so if the user tries to set the sample rate in their simulator program, the simulator won't crash. Currently, `setSampleRate` and `setMicGain` will not work in the simulator. This is something we might want to look into for future work. 

You can play with it here: https://makecode.microbit.org/app/cc0b1e751947b5dfa24a287ba0770de1a848ab54-da5a9d1a93

Closes https://github.com/microsoft/pxt-microbit/issues/5062
Closes https://github.com/microsoft/pxt-microbit/issues/5006
Will partially close https://github.com/microsoft/pxt-microbit/issues/4610 there might be more we want to do in the future.


package.json
```diff
@@ -32,6 +32,7 @@
         "docs/static/icons/favicon.ico"
     ],
     "devDependencies": {
+        "@types/dom-mediacapture-record": "^1.0.16",
```
jwunderl, line 35: for reference, this is the part needed to fix the issue Sarah mentioned seeing in stand up; it appears that up until a certain point mediarecorder api wasn't present in typescripts libs by default, and I think our tsconfig's libs is fighting it still / it's hidden away somewhere there. Easiest fix is just pulling it in explicitly from definitelytyped like this. re: microsoft/typescript#34728

sim/state/record-audio.ts
```diff
@@ -4,26 +4,64 @@ namespace pxsim  {
...
+            if (navigator.mediaDevices && navigator.mediaDevices.getUserMedia) {
+                try {
+                    stream = await navigator.mediaDevices.getUserMedia({video: false, audio:true});
+                    recorder = new MediaRecorder(stream);
+                    recorder.start();
+
+                    setTimeout(() => {
+                        recorder.stop();
+                        runtime.queueDisplayUpdate();
+                    }, 4000)
```
jwunderl, line 30: I'd almost be tempted to throw a  `+ Math.random() * 1000` just to 'simulate' that it's not consistent timing wise.

sim/visuals/microbit.ts
```diff
@@ -502,11 +502,17 @@ path.sim-board {
...
         private updateRecordingActive() {
             const b = board();
-            if (!b
-                || !b.recordingState.currentlyRecording)
+            if (!b)
                 return;
 
-            this.updateTheme();
+            let theme = this.props.theme;
+            if (b.recordingState.currentlyRecording) {
```
jwunderl, line 509: hm, should we leave this 'on' when `b.microphoneState.sensorUsed` so that it's showing as 'listening' when you have on loud sound / etc?

sim/state/record-audio.ts
```diff
@@ -4,26 +4,64 @@ namespace pxsim  {
     }
 }
 namespace pxsim.record {
+    let stream: MediaStream;
+    let recorder: MediaRecorder;
+    let chunks: Blob[];
+    let audioURL: string;
+    let recording: HTMLAudioElement;
```
jwunderl, line 11: I don't think this persists past a device restart on hardware, right? if so we need to clear this state during https://github.com/microsoft/pxt-microbit/blob/5cbf534f4dc454103c5b571740a03fd357481b71/sim/dalboard.ts#L102 -- so probably just store all these things on `RecordingState` defined above instead of as global variables

sim/state/record-audio.ts
```diff
@@ -4,26 +4,64 @@ namespace pxsim  {
...
+        }
     }
 
     export function stop(): void {
 
     }
 
     export function erase(): void {
-
+        chunks = [];
```
jwunderl, line 64: chunks is just the value stored while recording and not the recording itself, right? I'd think calling erase would clear `recording` so that `erase(); play();` never does anything.

sim/state/record-audio.ts
```diff
@@ -4,26 +4,64 @@ namespace pxsim  {
...
     export async function record(): Promise<void> {
         //request permission is asynchronous
         let b = board();
         if (!b.recordingState.currentlyRecording) {
             b.recordingState.currentlyRecording = true;
             runtime.queueDisplayUpdate();
+
+            if (navigator.mediaDevices && navigator.mediaDevices.getUserMedia) {
+                try {
+                    stream = await navigator.mediaDevices.getUserMedia({video: false, audio:true});
```
jwunderl, line 23:
```suggestion
                    stream = await navigator.mediaDevices.getUserMedia({ video: false, audio: true });
```
nit spacing

sim/state/record-audio.ts
```diff
@@ -4,26 +4,64 @@ namespace pxsim  {
...
+                    recorder = new MediaRecorder(stream);
+                    recorder.start();
+
+                    setTimeout(() => {
```
jwunderl, line 27: we probably want a `setTimeout` in the `catch` here  / else branch turning off `currentlyRecording` after a few seconds when mediaDevices is not defined or user has rejected permissions to use it.

sim/state/record-audio.ts
```diff
@@ -4,26 +4,64 @@ namespace pxsim  {
...
+                        audioURL = window.URL.createObjectURL(blob);
+                        recording = new Audio(audioURL);
+                        b.recordingState.currentlyRecording = false;
+                        erase();
+                    }
+                } catch (error) {
+                    console.log("An error occurred, could not get microphone access");
+                }
+
+              } else {
```
jwunderl, line 47:
```suggestion
            } else {
```

jwunderl (commented):
Nice job!

Probably should move state from being top level variables to fields in `RecordingState`, see comment

sim/state/record-audio.ts
srietkerk: I could see the value in this. However, it's not truly random, the only thing that changes the length of the recording is if the input sample rate is changed.

srietkerk: One thing that I could do is add a `sampleRate` variable as part of the state. I could then use the variable in this set timeout, and then have `setInputSampleRate` change that value.

sim/visuals/microbit.ts
srietkerk: Oh, crud, yes. 

Hmm, that does make me wonder if there should be some extra indicator.. If you have `onLoudSound` in the workspace, it will be practically impossible to be able to decipher when the sim is recording and when it's done. I might try to play with making the mic indicator blink if recording is happening.

sim/state/record-audio.ts
jwunderl: Yeah, we could play with that in a future update. 

is the recording time actually pretty stable just based off the `sampleRate`? I know `sampleRate` affects it (by virtue of just... changing the rate it reads data) but I was under the impression that on actual physical devices there was something like 100-500ms wiggle room between sample times even with the same sample rate 'consistent', if it's actually consistent (let's say > 100ms) then it'd be fantastic to work out that timing as close as possible in the future (and possible the `Math.random` to include the wiggle room as well :) )

sim/visuals/microbit.ts
jwunderl: Yeah, check hardware behavior when the features are used at the same time and match -- if we show it blinking and it doesn't on hardware that might be confusing as well.

sidenote; 'cute' feature I've tossed around in my head before would be making it so if you hover over an 'additional hardware' portion of the simulator (that is, something that doesn't get shown with a blank project) for a few seconds we gave a tooltip with 'This is here because you're using x feature in your program', as it can get confusing where certain features come from -- e.g. this teacher on the forum who was having trouble helping a student because their project used radio in arcade https://forum.makecode.com/t/2-simulators-running-simultaneously/19433. Feature would probably get annoying real quick though / would need design help to make it not overbearing

sim/state/record-audio.ts
srietkerk: Yes, that's correct. 

Would just setting `recording = null` here suffice? Am I causing a memory leak by creating these audio objects every time I record? Should I be just changing the source?

jwunderl: It's probably fine for testing but we can play around with it. I was gonna say it shouldn't matter / garbage collector should handle it when the refs clear but on a whim I gave it a search and this does appear to hit a weird edge case in chromium right now (probably some circular dependency thing between mediarecorder and it's parameter? idk I didn't look into it past an 'Oh neat') https://bugs.chromium.org/p/chromium/issues/detail?id=899722

sim/state/record-audio.ts
srietkerk: It's possible that there is some 100-500ms wiggle room per recording with the sample rate the same. I haven't tried to log it and trying to test it with my own timer is just chock full of inconsistencies as it is. For what it's worth, though, I could not tell a difference in time between recordings with the same settings.

sim/visuals/microbit.ts
srietkerk: It doesn't blink on the simulator. I'll just change it so that it doesn't get updated if `b.microphoneState.sensorUsed` is true.

sim/state/record-audio.ts
srietkerk: Good to know. I'll try to change it to see if anything bad happens.

srietkerk: Looking into it, I think this will be a bit more involved. I would need a bit more refactoring to use just one media recorder. For the sake of wanting to get this in for testing, I'm going to push this work out a bit.

jwunderl (approved):
let's get this in for testing today, just keep track of previous comments for a follow up pr :)

PR #5103, opened 2023-04-28
Recording sim: Moved global variables into the board recording state

srietkerk:
We don't want to cause memory leaks if the board gets destroyed. By moving the global variables into the recording state object that gets initialized in `dalboard.ts`, the variables should get destroyed if the board gets reconstructed. 

Done in thanks to Joey's comment here: https://github.com/microsoft/pxt-microbit/pull/5082#discussion_r1178571516

Work related to https://github.com/microsoft/pxt-microbit/pull/5082

### DIFF
--- a/sim/state/record-audio.ts
+++ b/sim/state/record-audio.ts
@@ -99,7 +99,7 @@ namespace pxsim.record {
 
     export function audioIsStopped(): boolean {
         const b = board();
-        if (!b) return false;
+        if (!b) return true;
         return b.recordingState.recorder ? !b.recordingState.audioPlaying && b.recordingState.recorder.state == "inactive" : true;
     }
 
--- a/sim/state/record-audio.ts
+++ b/sim/state/record-audio.ts
@@ -1,42 +1,42 @@
 namespace pxsim  {
     export class RecordingState {
         currentlyRecording = false;
+        stream: MediaStream;
+        recorder: MediaRecorder;
+        chunks: Blob[];
+        audioURL: string;
+        recording: HTMLAudioElement;
+        audioPlaying: boolean = false;
     }
 }
 namespace pxsim.record {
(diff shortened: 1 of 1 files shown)


sim/state/record-audio.ts
```diff
@@ -73,24 +77,30 @@ namespace pxsim.record {
...
-        return recorder ? recorder.state == "recording" : false;
+        const b = board();
+        if (!b) return false;
+        return b.recordingState.recorder ? b.recordingState.recorder.state == "recording" : false;
     }
 
     export function audioIsStopped(): boolean {
-        return recorder ? !audioPlaying && recorder.state == "inactive" : true;
+        const b = board();
+        if (!b) return false;
```
jwunderl, line 102: Should this be true? I'm assuming stopped is basically just "is not playing"? Might be wrong on that assumption

jwunderl (approved):

sim/state/record-audio.ts
srietkerk: Ah, yes, this should be true. Thank you!

thsparks (approved):

sim/state/record-audio.ts
```diff
@@ -1,42 +1,42 @@
...
-    let recorder: MediaRecorder;
-    let chunks: Blob[];
-    let audioURL: string;
-    let recording: HTMLAudioElement;
-    let audioPlaying: boolean = false;
-
     export async function record(): Promise<void> {
         //request permission is asynchronous
         let b = board();
+        let recordingState = b.recordingState;
```
kimprice, line 16: nit: do you want to scope down `recordingState`? You only use it within the `try` so you could wait to define it there.

kimprice (approved):
Looks good. Just a suggestion on scoping down `recordingState`.

Issue #5092, opened 2023-04-27
Recording extension: simulator plays old recording


kimprice:
Repro:
1. Add these blocks to your project: 
![image](https://user-images.githubusercontent.com/15070078/234988738-4c614ca4-087d-4cd2-a4c2-36a032a6a0f0.png)
2. Press A in the simulator to record.
3. Press A again to record something new. Then press B before the recording is done.
4. Hear that the first recording is what played.

Expected: Pressing B should end the recording and immediately play the new recording.

Issue #5091, opened 2023-04-27
Record icon in browser tab does not disappear when audio recording is finished


thsparks:
While recording audio, this red icon appears in my browser tab (in Edge):
<img width="103" alt="image" src="https://user-images.githubusercontent.com/69657545/234988561-bef09a1a-b060-4c8d-a883-b3065b9b6c8e.png">

I think the icon is expected during recording, but it doesn't go away once recording is finished, and I'd guess it probably should?

abchatra:
May be fixed by this? https://github.com/microsoft/pxt-microbit/pull/5103

srietkerk:
No, this is something different.

Issue #5089, opened 2023-04-27
Play Audio Clip does not stop when simulator is stopped


thsparks:
If audio from the record extension's "Play Audio Clip" block is playing and you try to stop the simulator, the sound will keep going until it's finished. Only about 4 seconds though, so not a huge issue.

abchatra:
We need to mimic hardware. Play should stop recording.

PR #5108, opened 2023-05-03
Recording extension: fixed stop recording bugs on simulator

srietkerk:
This PR has everything stop recording!! 

- Fixed the thing that Joey mentioned that if there was a failure anywhere in getting the user media or anything else in `record`, we want to make sure the `recordingState` reflects as such. If there are errors or the user denies access to the microphone, `currentlyRecording = false` is now set. (https://github.com/microsoft/pxt-microbit/pull/5082#discussion_r1178573758)

- Audio and recording now stops when the simulator stop event is fired; closes https://github.com/microsoft/pxt-microbit/issues/5089

- If `play` is triggered before the recording takes the length of the timeout, the timeout is cleared and recorder is stopped. I've added 100ms of buffer time for the recording to populate after recorder is stopped, which then allows the newly recorded audio clip to play immediately. Closes https://github.com/microsoft/pxt-microbit/issues/5092

- When `record` is done, the indicator on the browser tab now reflects that recording has stopped. Closes https://github.com/microsoft/pxt-microbit/issues/5091

- The simulator fully mimics what the hardware does with `record` and `play`. This is non-blocking logic. 

You can play with all these things here: https://makecode.microbit.org/app/c2c6876a5e3af38fcc17f5cb4f21e47c388621db-27857f50af


abchatra:
🏆

sim/state/record-audio.ts
```diff
@@ -7,65 +7,133 @@ namespace pxsim  {
...
+                    this.audioPlaying = false;
+                });
+            }
+        }
     }
 }
 namespace pxsim.record {
+
+    function stopRecording(b: DalBoard): void {
+        b.recordingState.recorder.stop();
```
kimprice, line 28: Is it safe to assume `recordingState` has been initialized?

sim/state/record-audio.ts
```diff
@@ -7,65 +7,133 @@ namespace pxsim  {
...
+
+        initListeners = () => {
+            if (this.recording) {
+                this.recording.addEventListener("play", () => {
+                    this.audioPlaying = true;
+                });
+
+                this.recording.addEventListener("ended", () => {
+                    this.audioPlaying = false;
+                });
```
kimprice, line 20: Should we have clean up code to remove the event listeners?

sim/state/record-audio.ts
srietkerk: I'm pretty sure. If the board is initialized, this line will get called https://github.com/microsoft/pxt-microbit/blob/ed40af178811c7e6a40febbbf0dacd91a3db53bd/sim/dalboard.ts#L102 and thus initialize the state. So as long as the board exists, which is assumed in the contexts in which we call `stopRecording`, `recordingState` should be initialized.

sim/state/record-audio.ts
kimprice: Yea, although I'm not sure where you should call it.

sim/state/record-audio.ts
```diff
@@ -7,65 +7,139 @@ namespace pxsim  {
...
+                    runtime.queueDisplayUpdate();
+                }, 4000)
+
+                b.recordingState.recorder.ondataavailable = (e: BlobEvent) => {
+                    b.recordingState.chunks.push(e.data);
                 }
 
-            } else {
-                console.log("getUserMedia not supported on your browser!");
+                b.recordingState.recorder.onstop = () => {
```
kimprice, line 77: nit: how come `onstop` isn't camel cased?

kimprice: oh is this just MediaRecorder?

kimprice (approved):
LGTM!

sim/state/record-audio.ts
srietkerk: Yeah, its lowercase for event handlers. If I try to use camel case, typescript gives me an error. It looks like this is just a pattern for a built-in objects that have event handlers/inherit from `EventTarget` in JavaScript. A little blurb from the Mozilla docs that I got help from:
image

Issue #5063, opened 2023-04-21
No discernable difference in audio when setting microphone sensitivity


Jaqster:
Changing the microphone sensitivity between low, medium and high - doesn't result in any discernable difference in sound output.
Repro - try recording 3 different sound clips on 3 different micro:bits using Low, Medium and High microphone sensitivity levels.  Play back the audio on all 3 micro:bits.  Notice no difference in sound recording.
![image](https://user-images.githubusercontent.com/22108954/233700215-a1c0caa4-3134-4df4-90c8-a28361a47b24.png)

jaustin:
@finneyj @JohnVidler Could you please test the gain values that @srietkerk is using from C++ and confirm you think they're a good set for the range?

srietkerk:
Will do more investigation and resolve this after the changes from this issue: https://github.com/lancaster-university/codal-microbit-v2/issues/324

abchatra:
@srietkerk are you tracking this one?

PR #5112, opened 2023-05-04
Recording extension: greater difference between microphone gain

srietkerk:
We want to make sure that the different gain selections actually sound different to users, so the differences between medium and high have been increased. I've found in my testing that anything above a 0.1f gain value has a decent amount of interference, so the gap between medium and high are larger than that between low and medium. 

Please play with it here and let me know if you can tell a difference between the levels: https://makecode.microbit.org/app/b0641ceafb772707ea5213d20f40a158ef8df6f2-a7767ee627

Draft pull request because there might be other changes to be made depending on possible pipeline changes that I will be discussing with the micro:bit team.

EDIT: There will be more changes here when we get APIs to change the mic gain on the mixer channel. We won't have to create another channel and the `localProcessor` will not be needed.

Closes https://github.com/microsoft/pxt-microbit/issues/5063

### DIFF
--- a/libs/audio-recording/recording.cpp
+++ b/libs/audio-recording/recording.cpp
@@ -116,10 +116,10 @@ void setMicrophoneGain(int gain) {
         uBit.audio.processor->setGain(0.079f);
         break;
     case 2:
-        uBit.audio.processor->setGain(0.2f);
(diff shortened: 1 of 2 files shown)


srietkerk:
Will change this, if needed, when CODAL apis are updated

PR #5117, opened 2023-05-05
get rid of old, unused pxtarget field

jwunderl:
These were added with https://github.com/microsoft/pxt-microbit/pull/250 for user with https://github.com/microsoft/pxt/pull/330, but we've redone that dialog 3 or 4+ times since then so these images only remain in docs images :)

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -448,68 +448,6 @@
         ],
         "hasReferenceDocs": true,
         "usbDocs": "/device/usb",
-        "usbHelp": [
-            {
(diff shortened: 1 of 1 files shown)


thsparks (approved):

kimprice:
Are the images still used in our docs somewhere?

jwunderl:
The first three I checked were used elsewhere, but I went ahead and checked and a couple weren't so now removed~

kimprice (approved):

Issue #5122, opened 2023-05-09
Block "play sound until done" doesn’t work if dragging melody to replace sound


Sheila111:
**Describe the bug**
Block "**play sound until done**" doesn’t work if dragging **melody** to replace **sound**

**Steps to reproduce the behavior**
1. Navigate to https://makecode.microbit.org/beta#
2. Drag block "**play melody at tempo**" and "**play sound until done**"
3. Drag **melody** to replace **sound**

**Actual behavior**
Block "**play sound until done**" doesn’t work if dragging **melody** to replace **sound**

**Expect behavior**
It should work if dragging **melody** to replace **sound**. If it's by design, I think we should forbid user dragging melody to replace sound.

**Screenshot**
![1](https://user-images.githubusercontent.com/59192069/236985434-fbb14746-c586-471b-bd30-f7e27398e467.gif)

**Additional context**
1.OS: Windows11
2.makecode.microbit.org version: 5.1.26
3.Microsoft MakeCode version: 8.6.33

Issue #4906, opened 2023-01-10
Standardize tempo slider


Jaqster:
There are a couple different inputs for tempo
- Play Melody: Label is "Tempo", min is 40, max is 500
- Set Tempo: Label is "Bpm", min is 4, max is 400
- Change Tempo: number field
![image](https://user-images.githubusercontent.com/22108954/211434840-82f02cc3-33a0-45f3-914e-3326703053be.png)
![image](https://user-images.githubusercontent.com/22108954/211434867-9ef6d10c-1e14-4fb7-bc0e-7ef6e910f0a7.png)
![image](https://user-images.githubusercontent.com/22108954/211434902-8c57cfbf-86f9-44c2-a70c-ab5106424fd1.png)

I think it's fine to leave Change Tempo as a number field, but would be good to standardize the Play Melody and Set Tempo field editors to:
- Use the same label "beats per minute"
- Use the same min value (not sure what this would realistically be?)
- Use the same max value (also not sure what's the max value for the micro:bit to play?)

Issue #4899, opened 2023-01-03
Organization of Music Blocks


Jaqster:
As we are adding more and more audio blocks, I think it may be time to take a look at how we're organizing these blocks.
Attaching the current list here - and including the Arcade sound blocks as well, as it would be great to have some consistency across.

It seems to me that we have 2 different sets of blocks:
1.  Sounds and sound effects (based on frequency values)
2.  Music (based on musical note values) 
And blocks that affect both (volume, tempo values).
[Music Blocks.pdf](https://github.com/microsoft/pxt-microbit/files/10338452/Music.Blocks.pdf)

Let's discuss how we can best represent these blocks to users.  This issue may be related...https://github.com/microsoft/pxt-microbit/issues/4482.

jaustin:
It's also worth noting that the language across some of these blocks is a bit confusing in the way they now relate to each other. 
<img width="479" alt="image" src="https://user-images.githubusercontent.com/147113/210671282-4645afc1-0e03-4394-9940-9996394d0e60.png">

What's the difference between "play" a melody or "start" a melody, when we "play" a sound (but we can start it in the background, too...). The melodies we "start" are different to the ones we can "play"!
For example the relationship between these options, and then
<img width="426" alt="image" src="https://user-images.githubusercontent.com/147113/210671330-f5066337-d993-4916-bf2c-3cb59cdee9c0.png">

"until done" and "in background" 

More detail in [#4735 ](https://github.com/microsoft/pxt-microbit/issues/3823)

abchatra:
Lets port the play paradigm from the arcade which solves various issues.

abchatra:
@srietkerk  I am assigning this issue to @thsparks

PR #5123, opened 2023-05-09
Update Music Blocks in Micro:bit

thsparks:
This updates the music blocks for Micro:bit similarly to what https://github.com/microsoft/pxt-common-packages/pull/1399 did for arcade.

The new blocks look like this:
image
image

Upload target: https://makecode.microbit.org/app/1fef66e42c481585ce96eb46f0de278db809df52-53221f84ce

Fixes https://github.com/microsoft/pxt-microbit/issues/4899
Fixes https://github.com/microsoft/pxt-microbit/issues/4906
Fixes https://github.com/microsoft/pxt-microbit/issues/5122
Fixes https://github.com/microsoft/pxt-microbit/issues/4843


libs/core/melodies.ts
```diff
@@ -69,6 +69,23 @@ enum Melodies {
 }
 
 namespace music {
+    export class StringArrayPlayable extends Playable {
```
thsparks, line 72: >StringArrayPlayable [](http://example.com/codeflow?start=17&length=19)

Move to playable

libs/core/music.ts
```diff
@@ -223,7 +224,7 @@ namespace music {
      * @param ms rest duration in milliseconds (ms)
      */
     //% help=music/rest weight=79
-    //% blockId=device_rest block="rest(ms)|%duration=device_beat"
+    //% blockId=device_rest block="rest for |%duration=device_beat"
```
riknoll, line 227: was this intentional? it will invalidate all existing translations

libs/core/music.ts
```diff
@@ -592,8 +630,8 @@ namespace music {
         }
     }
 
-    export function _bufferToMelody(melody: Buffer) {
-        if (!melody) return [];
+    export function _bufferToMelody(melody: Buffer): string[] {
+        if (!melody) [];
```
riknoll: shouldn't this still return?

libs/core/music.ts
```diff
@@ -479,11 +516,12 @@ namespace music {
     //% help=music/stop-all-sounds
     //% blockId=music_stop_all_sounds block="stop all sounds"
     //% weight=10
-    //% group="Volume"
+    //% group="Melody"
```
riknoll, line 519: this applies to all sounds, correct? i think volume makes a bit more sense

libs/core/playable.ts
```diff
@@ -0,0 +1,134 @@
...
+            super();
+        }
+
+        _play(playbackMode: PlaybackMode) {
+            if (playbackMode === PlaybackMode.InBackground) {
+                control.runInParallel(() => music.playTone(this.pitch, this.duration));
+            }
+            else if (playbackMode === PlaybackMode.UntilDone) {
+                music.playTone(this.pitch, this.duration);
+                if (this.duration > 2000) {
```
riknoll, line 88: i'm guessing this was just copy/pasted, but perhaps we should consider fixing this crazy undocumented behavior

libs/core/music.ts
thsparks: Yeah accidental change, I'll undo it

libs/core/music.ts
thsparks: Hmm, it was intentional as a fix for https://github.com/microsoft/pxt-microbit/issues/4843

Should I deprecate the block and make a new one instead?

libs/core/soundexpressions.ts
```diff
@@ -411,6 +424,50 @@ namespace music {
...
+    //% effect.defl=SoundExpressionEffect.None
+    //% interpolation.defl=InterpolationCurve.Linear
+    //% compileHiddenArguments=true
+    //% inlineInputMode="variable"
+    //% inlineInputModeLimit=3
+    //% expandableArgumentBreaks="3,5"
+    //% toolboxParent=music_playable_play
+    //% toolboxParentArgument=toPlay
+    //% group="micro:bit (V2)"
+    export function createSoundExpression(waveShape: WaveShape, startFrequency: number, endFrequency: number, startVolume: number, endVolume: number, duration: number, effect: SoundExpressionEffect, interpolation: InterpolationCurve): SoundExpression {
```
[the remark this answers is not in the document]
thsparks: It's new. It's a replacement for createSoundEffect which is deprecated after this change. I'd originally just updated createSoundEffect but changing the return type broke backwards compat, hence the new block instead.

libs/core/music.ts
riknoll: no, we should either keep this change or revert it. should maybe ask other folks on the team

riknoll:
can you also do some code size comparisons? to check code size:

1. load the editor with `?compiler=size` at the end of the URL (before the `#editor`)
2. switch to ts and download the compiled hex file
3. after the download is complete, open the built section of the file explorer
4. there should be two new `*.asm` files. the comments on top of each file should give you information on code size

as for the program to check, just write some simple code that uses the new blocks in your localhost and the old blocks in either beta or localhost without your changes.

libs/core/playable.ts
thsparks: Huh. Do you know what the goal was here? Seems like we could just remove it?

riknoll: i honestly have no idea. i remember running into this code a long time ago and being similarly puzzled.

maybe it would be a good idea to ask @pelikhan

thsparks:
> can you also do some code size comparisons?

@riknoll - In the current form (with a simpler parseInt and combined StringPlayable and StringArrayPlayable) it adds about 696 bytes. What's a good target? Should I keep chipping away until there's strictly no increase at all?

riknoll:
i think that's good for now! if we have more time, we should try to optimize this more.

libs/core/_locales/core-jsdoc-strings.json
```diff
@@ -521,6 +521,9 @@
   "music.beat": "Returns the duration of a beat in milli-seconds",
   "music.beginMelody": "Use startMelody instead",
   "music.builtInMelody": "Gets the melody array of a built-in melody.",
+  "music.builtInPlayableMelody": "Gets the melody array of a built-in melody.",
+  "music.builtinPlayableSoundEffect": "Get the sound expression string for a built-in a sound effect.",
```
kimprice, line 525:
```suggestion
  "music.builtinPlayableSoundEffect": "Get the sound expression string for a built-in sound effect.",
```

riknoll: ooh, if you change that don't do it in this file (which is autogenerated), do it in the source file

libs/core/soundexpressions.ts
```diff
@@ -433,7 +491,24 @@ namespace music {
     //% toolboxParent=soundExpression_playSoundEffect
     //% toolboxParentArgument=sound
     //% weight=98 help=music/builtin-sound-effect
+    //% deprecated=1
     export function builtinSoundEffect(soundExpression: SoundExpression) {
         return soundExpression.getNotes();
     }
+
+    /**
+     * Get the sound expression string for a built-in a sound effect.
```
kimprice, line 500:
```suggestion
     * Get the sound expression string for a built-in sound effect.
```

libs/core/_locales/core-jsdoc-strings.json
kimprice: ^ I guess this fix won't really matter since it's automated, but I made the same suggestion below.

libs/core/music.ts
```diff
@@ -326,10 +327,25 @@ namespace music {
...
     //% blockHidden=true
     //% group="Melody Advanced"
+    //% deprecated=1
     export function builtInMelody(melody: Melodies): string[] {
         return getMelody(melody);
     }
 
+    /**
+     * Gets the melody array of a built-in melody.
+     * @param name the note name, eg: Note.C
```
kimprice, line 337: Shouldn't this param be melody?

thsparks: Yup! Good catch.

riknoll: also, get rid of the `eg: Note.C`, which actually has meaning in MakeCode (it will attempt to set the default value for that param in the toolbox)

libs/core/music.ts
```diff
@@ -326,10 +327,25 @@ namespace music {
...
+     */
+    //% weight=60 help=music/builtin-melody
+    //% blockId=device_builtin_melody_playable block="%melody"
```
riknoll, line 340:
```suggestion
    //% blockId=device_builtin_melody_playable block="$melody"
```

libs/core/music.ts
thsparks: For future reference, followed up with team in Teams and current recommendation is to continue with the change.

libs/core/playable.ts
```diff
@@ -0,0 +1,135 @@
...
+    //% bpm.min=40 bpm.max=500
+    //% bpm.defl=120
+    export function stringPlayable(melody: string, bpm: number): Playable {
+        return new StringArrayPlayable(music.getMelodyNotes(melody), bpm);
+    }
+
+    /**
+     * Plays a tone through pin ``P0`` for the given duration.
+     * @param frequency pitch of the tone to play in Hertz (Hz), eg: Note.C
+     * @param ms tone duration in milliseconds (ms)
```
kimprice, line 110: Should these match the names of the params in the function signature?

abchatra:
Removing this getter intentional? 
image

abchatra:
Overall nice changes, I expect some push back on changing "start melody" to "play". Lets check-in this and evaluate that later.

thsparks:
> Removing this getter intentional? image

I'm not really sure what value this block adds by itself and Arcade does not expose it, so I took it out. However, if you think it's important to include it for consistency, I can put it back. (That said, I suspect the code to do so will be somewhat ugly, but hopefully not a huge issue.)

abchatra:
One can set multiple sound effects to a variable. This is just exposing the inner field
image

abchatra:
I believe there is content out there which does this.

jwunderl (approved):

Jaqster:
Very nice!  Much cleaner!  
How do people feel about moving the v2 music block section more towards the top of the toolbox menu?

abchatra:
Good idea, Where do you want it?  I would like it to be at least above melody advanced. (Let's not block on this, we can do a subsequent change).

thsparks:
> Very nice! Much cleaner! How do people feel about moving the v2 music block section more towards the top of the toolbox menu?

I'm going to check this change in as-is for now, but I'm open to making this change in a future check-in if the team wants it.

Jaqster:
Yeah.  Maybe even above Tempo?

thsparks:
There is a slight risk that people without v2s will think everything below that line is v2 only, not just the one category. Not sure if that's something we've seen/tested before.

Jaqster:
Ah.  Good point.  I wish there were "v2" labels or something we could put next to the individual blocks, instead of having a separate section...

Issue #4987, opened 2023-03-15
Project thumbnail keeps loading in share dialog


Sheila111:
**Describe the bug**
**Project thumbnail** keeps loading in **share dialog**

**Steps to reproduce the behavior**
1. Navigate to https://makecode.microbit.org/beta#
2. Create a project and share the project

**Actual behavior**
**Project thumbnail** keeps loading in **share dialog**

**Expect behavior**
Show the **project thumbnail** in share dialog

**Screenshot**
![1](https://user-images.githubusercontent.com/59192069/225220010-6f4ded38-a92b-4bee-9840-a0ac00dbcd13.gif)

**Additional context**
1.OS: Windows11
2.makecode.microbit.org version: 5.1.5
3.Microsoft MakeCode version: 8.6.12

abchatra:
@Sheila111  seems like fixed in 5.1.38. If not reopen.

Sheila111:
Fixed

PR #5128, opened 2023-05-10
Fix project thumbnail loading

kimprice:
Now there is a default screenshot instead of a loading screen for the project thumbnail.
Fixes https://github.com/microsoft/pxt-microbit/issues/4987
Instead of having to enable the screenshot and gif experiments, they are now available by default.

image

This also makes the screenshot button always appear in the simulator toolbar.
image


pxtarget.json
```diff
@@ -538,7 +536,9 @@
         "debugger": true,
         "simGifTransparent": "rgba(0,0,0,0)",
         "simGifMaxFrames": 44,
+        "simScreenshot": true,
         "simScreenshotMaxUriLength": 300000,
+        "simGif": true,
```
jwunderl, line 541: Weren't there a few more fields for keyboard shortcuts you'd want in here?

kimprice: The keyboard shortcuts don't currently work so I wasn't sure if I should include the label. I was going to update pxt so that it doesn't just show undef for the keyboard shortcuts when they don't exist.

jwunderl: ah okay. If we're gonna enable it we'll want the keyboard shortcuts anyway so I'd probably add them in even if they require a pxt side fix to work, but fine either way~

eanders-ms (approved):

PR #5144, opened 2023-05-16
change second step in pairing to gif?

jwunderl:
try the webusb pairing flow this build: https://makecode.microbit.org/app/ec6422ff090c4097696498507fd22f4fb41a4edc-99e8a867c2#editor - this changes it from a png showing a generic computer to just show the pairing flow from the browser. (a little tedious to get it to capture, the browser is very sensitive to focus moving or outside clicks while pairing and aggressively exits for to avoid any chance of it happening without direct user action)

the dialog is a little unfortunate; we can't do any resizing of it or anything, so as far as I know it has to stay this big / wasted space / a little zoomed out. I can fiddle with the gif sizing / maybe manually edit the gif and stitch it together so it's "zoomed in"? it would look weird though.

As another option I added in a few circles + numbers for the step, but it's kinda awkward looking / probably needs a second try if we want it, and I'm not sure how much is gained over the clicks themselves
selecting-microbit-take-2-with-numbering


kimprice:
I wonder if it would be better to show the pop-up with more context so they can see where it will show up on the window. Right now, you can't tell that it will show up at the top left.

I'm also slightly concerned about the dark theme. Maybe that would throw people off. From today's classroom visit, all the popups I saw were the default light theme.

jwunderl:
I don't think there's any way to really show the context more -- I did somewhat by leaving the frames where it 'pops in' in front of the sim, but the browser is in control of where it goes (e.g. it'll be on the right hand side for rtl languages most likely) and if we zoom out you won't be able to really see the interaction that it is showing (click device, click 'confirm' button). That said, it is a pop up, so it's always going to be a bit eye-catching / invasive as it will 'pop in', so making sure they see that they have to do something with the pop up feels like the most important part (it's almost unfortunate it shows words at all for translation reasons, ideally it might even just be outlines in the approximate shape of the interaction)

re: light mode, yeah, the image is more of a placeholder to show the idea, not something I'd merge immediately - would want to smooth out cursor movement / etc either way.

I did check and it looks like edge has changed the dialog to be slightly different recently so would probably also want to put in a check to show a different gif for both browsers (but the paradigm is still exactly the same, of choosing the device from a list & clicking confirm)

thsparks (approved):
In the comments, you mentioned some adjustments you're thinking about making, and those might still be helpful, but I think even as-is this would be a huge help.

jwunderl:
@thsparks @kimprice ill go ahead and redo the gif in light mode and merge in to get it testable then, do you two have any preference with / without labels like the gif I left in the pr description for the remade version? I can also make them arrows if that would be better

thsparks:
> @thsparks @kimprice ill go ahead and redo the gif in light mode and merge in to get it testable then, do you two have any preference with / without labels like the gif I left in the pr description for the remade version? I can also make them arrows if that would be better

Hmm, I'm not sure on that one. Maybe good to get @Jaqster's perspective.

kimprice:
I think the labels are helpful.

jwunderl:
Okay, one more question: centered rows:
image
 or columns: 
image

Also dropping the original, un-arrowed gif here for ref, so we can swap over to a nice arrow animation / maybe mp4 later:
selecting-microbit-take-three

srietkerk (approved):
I like the column layout better. Looks great, though!

kimprice (approved):

Issue #5163, opened 2023-05-19
Micro:bit not moving to next icon after recording, but sim does


kiki-lee:
Using app on iPad.  This happens for both "until done" and "in background".

This program appears to run correctly in the sim, but on the device it never passes the recording stage.

![Microsoft Edge - Microsoft MakeCode for micro-bit 2023-05-19 at 10 46 03 AM](https://github.com/microsoft/pxt-microbit/assets/1633958/01c5621d-70a0-4d9c-af7e-c1f1f93ad734)

![Microsoft Edge - Microsoft MakeCode for micro-bit 2023-05-19 at 10 46 10 AM](https://github.com/microsoft/pxt-microbit/assets/1633958/e8760f31-b271-4f5f-b15d-c68b87bdcb90)

![Screenshot by Dropbox Capture](https://github.com/microsoft/pxt-microbit/assets/1633958/39ac478d-f432-4c0f-8e3f-122174302d2f)

abchatra:
Dupe of https://github.com/microsoft/pxt-microbit/issues/5162. But I will keep this bug as it has more details.

kiki-lee:
Ah, that means it's happening on more than the ipad. Good info!

Kim's Repro:

Open https://makecode.microbit.org/_P2fR3zMz6Dsa
Download to microbit
Press A to record, see that recording doesn't stop
makecode.microbit.org version: [5.1.30](https://github.com/Microsoft/pxt-microbit/releases/tag/v5.1.30)
Microsoft MakeCode version: [8.6.39](https://github.com/Microsoft/pxt/releases/tag/v8.6.39)
microbit runtime version: [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
codal-microbit-v2 runtime version: [v0.2.53](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.53)

Issue #5161, opened 2023-05-19
Unable to rerecord on hardware


kimprice:
Repro:
1. Open https://makecode.microbit.org/_HoRe9Lak9cq5
2. Download to microbit
3. Press A to record
4. Press B to play recording
5. Press A again and see that you can't rerecord

makecode.microbit.org version:  [5.1.30](https://github.com/Microsoft/pxt-microbit/releases/tag/v5.1.30)
Microsoft MakeCode version:  [8.6.39](https://github.com/Microsoft/pxt/releases/tag/v8.6.39)
microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
codal-microbit-v2 runtime version:  [v0.2.53](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.53)

PR #5170, opened 2023-05-20
Recording extension: recording, playback fixed

srietkerk:
The updates to the CODAL `StreamRecording` data type made it that we needed to use `recordAsync` and `playAsync` rather than `play` and `record`. Recording and playback now using these updated codal apis, and thus fixing the problems that we were having with recording earlier today.

Closes https://github.com/microsoft/pxt-microbit/issues/5163
Closes https://github.com/microsoft/pxt-microbit/issues/5162
Closes https://github.com/microsoft/pxt-microbit/issues/5161

For playing if anyone's interested: https://makecode.microbit.org/app/5f07e59c4f47400e1b1ff42f6329d302c150effe-f36946ff57

### DIFF
--- a/libs/audio-recording/recording.cpp
+++ b/libs/audio-recording/recording.cpp
@@ -32,17 +32,6 @@ static StreamRecording *recording = NULL;
 static SplitterChannel *splitterChannel = NULL;
 static MixerChannel *channel = NULL;
 
-void enableMic() {
-    uBit.audio.activateMic();
-    uBit.audio.mic->enable();
-}
-
-void disableMic() {
-    uBit.audio.mic->disable();
-    uBit.audio.deactivateMic();
-}
-
-
 void checkEnv(int sampleRate = -1) {
     if (recording == NULL) {
         if (sampleRate == -1)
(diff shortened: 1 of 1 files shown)


jwunderl (approved):

kimprice:
Does `recordAsync` and `playAsync` enable and disable the mic, respectively?

abchatra:
Awesome. Thanks for quick fix @srietkerk

srietkerk:
> Does `recordAsync` and `playAsync` enable and disable the mic, respectively?

Yes. With the CODAL updates that have been happening over the last couple of months, data streams (like `StreamRecording`) that use the mic will enable and disable the mic as the new pattern is activation by request. Most data stream sources/sinks have something like this in CODAL now as far as I understand: https://github.com/lancaster-university/codal-core/blob/94cbd3933c61924677fc1f2255eaf61bd175a0d1/source/streams/StreamRecording.cpp#L68

srietkerk:
> Awesome. Thanks for quick fix @srietkerk

Big thanks to @JohnVidler for the guidance!

Issue #4788, opened 2022-07-01
MakeCode 2022 consumes more memory and some v4 programmes won't download anymore


microbit-carlos:
**Describe the bug**
We've got several reports from users where their micro:bit MakeCode projects stopped working after MakeCode 2022 release date.

**To Reproduce**
1. Download this hex file: [microbit-fits-v4_does-not-v5.hex.zip](https://github.com/microsoft/pxt-microbit/files/9028251/microbit-fits-v4_does-not-v5.hex.zip)
2. Load it in https://makecode.microbit.org
3. Click "Download"
4. The error `program too big by 6184 bytes!` appears

The same hex file in https://makecode.microbit.org/v4 downloads correctly.

Rather than share one of the user programmes, I create this hex file by adding a bunch of extensions and just randomly dropping blocks on the workspace until it stopped working on MakeCode v5 and worked in v4. So the programme in itself will not make much sense.

**Expected behavior**

Existing programmes to still work.

**Screenshots**

<img width="317" alt="image" src="https://user-images.githubusercontent.com/29712657/176880567-72da8d50-1357-4f1d-8a12-4a19f431b2e6.png">

**micro:bit version (please complete the following information):**

N/A

**Desktop (please complete the following information):**
 - OS: macOS 12
 - Browser: Chrome
 - Version: 103

**Smartphone (please complete the following information):**

**Additional context**

makecode.microbit.org version:  [5.0.6](https://github.com/Microsoft/pxt-microbit/releases/tag/v5.0.6)
Microsoft MakeCode version:  [8.0.3](https://github.com/Microsoft/pxt/releases/tag/v8.0.3)
microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
codal-microbit-v2 runtime version:  [v0.2.40](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.40)

microbit-carlos:
With the provided hex file in the MakeCode 2022 editor, if I then add the Data Logger extension, which is configured to be V2 only, then MakeCode successfully compiles/downloads the programme.

So the memory issue might right now mostly affect V1 (or at least more easily in V1), and as V1 DAL hasn't been updated vs MakeCode v4, it suggest all the additional memory comes from pxt updates.

jaustin:
[...]

PR #5171, opened 2023-05-20
fix webusb flashing when disablesVariant has been used

jwunderl:
requires https://github.com/microsoft/pxt/pull/9524 for compileresult to attach the list of successfully built variants (that's why build is red)

Right now if you have a program that has a variant disabled (either pxt.json or from it being too big like https://makecode.microbit.org/_L92VubLgaWU6 and getting rebuilt only for codal) it fails to find the hexfile as we only use the unique names when building multivariant files. This change grabs the correct file in that case.

I don't know if there was a separate bug for this somewhere but thomas mentioned running into it to me over teams while he was testing music changes / it was mentioned in this comment https://github.com/microsoft/pxt-microbit/issues/4788#issuecomment-1425704556 -- so this is a mitigation for that, though other work remains in that issue.

build https://makecode.microbit.org/app/b51df1fe81ae69b332e39f54acebeadba9ac2009-60d9578ef1

If there's time, as a separate fix next week I'll look at making it so we do the v2 only build silently (without the pop up warning) when too big + a microbit v2 is fully connected if i have time (though I'll aim for progress bar first)

### DIFF
--- a/editor/flash.ts
+++ b/editor/flash.ts
@@ -242,6 +242,10 @@ class DAPWrapper implements pxt.packetio.PacketIOWrapper {
     }
 
     get binName() {
+        return `${this.devVariant}-${pxtc.BINARY_HEX}`;
+    }
+
+    get devVariant() {
         if (this.usesCODAL === undefined)
             console.warn('try to access codal information before it is computed')
         return (this.usesCODAL ? "mbcodal-" : "mbdal-") + pxtc.BINARY_HEX;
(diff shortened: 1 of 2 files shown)


eanders-ms (approved):

jaustin:
> If there's time, as a separate fix next week I'll look at making it so we do the v2 only build silently (without the pop up warning) when too big + a microbit v2 is fully connected if i have time (though I'll aim for progress bar first)

@jwunderl I know this was a long time ago now, but did you get around to this bit? If not I guess we could file it as an enhancement for the next release?

jwunderl:
> > If there's time, as a separate fix next week I'll look at making it so we do the v2 only build silently (without the pop up warning) when too big + a microbit v2 is fully connected if i have time (though I'll aim for progress bar first)
> 
> @jwunderl I know this was a long time ago now, but did you get around to this bit? If not I guess we could file it as an enhancement for the next release?

https://github.com/microsoft/pxt-microbit/pull/5184 should be that one & made it into release, are you seeing different behavior?

jaustin:
Ah, thanks! Not seeing different behaviour but was trying to trigger it without a real micro:bit connected as I'm away from usual desk. Apologies.

PR #5173, opened 2023-05-22
Reverting CODAL, v0.2.54 breaks recording, sound level

srietkerk:
Reverting CODAL to v0.2.53 for now. I should have done some testing on the branch before merging v0.2.54 as it breaks recording playback and sound level in a program like the following: 
image

`on loud sound` still having problems in both v0.2.53 and v0.2.54.

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -189,7 +189,7 @@
                 "codalTarget": {
                     "name": "codal-microbit-v2",
                     "url": "https://github.com/lancaster-university/codal-microbit-v2",
-                    "branch": "v0.2.54",
+                    "branch": "v0.2.53",
                     "type": "git"
                 },
                 "codalBinary": "MICROBIT",


thsparks (approved):

JohnVidler:
Strange... My CODAL only test code seems to behave itself.

Let's document the bug here for a mo and track debugging until we have a handle on why its behaving strangely on PXT

srietkerk:
Interesting. With some further investigation, it looks like the problem is with trying to get the sound level right after starting recording. That series of events seems to freeze up the microphone. I tried just having a while loop that displayed LEDs instead of the while loop with the sound level and everything is working fine.

srietkerk:
More weirdness.. The program in the original PR comment causes the mic and recording to lock up, but if you have a program with something in between the while loop start and the plotting of the sound level, it doesn't lock up. The sound level doesn't show in these programs, but it causes less of a problem than causing the whole micro:bit to lock.

Maybe some kind of race condition?

srietkerk:
In a similar vein, the following program works as expected, with the sound level actively updating as the user is recording. 
image

srietkerk:
Another couple test programs. It does look like some sort of race condition.

If you have a program that pauses even for one millisecond in the `while recording` loop, the sound level will behave as you expect. If you take that pause out, you'll get that locking behavior, the correct behavior happened with both of these programs. Pausing or serial writing seems to pause long enough. The following programs behave as you would expect.

image

image

srietkerk:
For added context, I have been testing all of this  using the uBit object for the on sound blocks. The code is based off the branch from this PR: https://github.com/microsoft/pxt-microbit/pull/5178

srietkerk:
Problems I was having was not a problem with the new codal, just with a test program. There is no need to revert.

PR #5175, opened 2023-05-22
Set default blocks for block tutorials

### LINKED ISSUE #5167: Remove default blocks from tutorials that don't need them
Hi Galen, can you go through and remove the default On Start and Forever blocks from the tutorials that don't use them?
Just taking a look, at the first row:
- Flashing Heart: remove On Start
- Name Tag: remove On Start
- Smiley Buttons: remove On Start and Forever
- Dice: remove On Start and Forever
- Love Meter: remove Forever
- Micro Chat: remove Forever

@jwunderl or @kiki-lee - do you know what the markdown is for removing these default blocks from the workspace in a tutorial?

ganicke:
Set the initial block/s for block choice tutorials. This is to not have any unused "on start" or "forever" blocks present by default.

Closes #5167

### DIFF
--- a/docs/projects/dice.md
+++ b/docs/projects/dice.md
@@ -11,10 +11,10 @@ Let's turn the @boardname@ into a dice!
 
 We need 3 pieces of code: one to detect a throw (shake), another to pick a random number, and then one to show the number.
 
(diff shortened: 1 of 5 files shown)


Jaqster (approved):
Thanks Galen!

Issue #5126, opened 2023-05-10
recording extension sim: microphone light doesn't always turns off


microbit-carlos:
The microphone LED might is only meant to be ON while the device is recording.

✅ In the simulator this works correctly with a programme like this. When the audio recording is finished the microphone light turns off:
<img width="390" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/29712657/387b54e9-3ee9-493e-801e-cb0b90ef0719">

❌ However, when using the `[plot bar graph]` block the microphone LED stays turned on after the recording is done. This screenshot was taken after the recording finished, as the `[clear screen]` block has taken place and display is clear:

<img width="1032" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/29712657/24a6b470-a546-487b-b524-ff7990a5c89f">

Maybe this is because the simulator widget to change the light level stays loaded on the simulator, but this is a deviation from the real behaviour on the device, as on the micro:bit the microphone LED goes off as expected.
- [microbit-sim-mic.hex.zip](https://github.com/microsoft/pxt-microbit/files/11444710/microbit-sim-mic.hex.zip)
```ts
input.onButtonPressed(Button.A, function () {
    record.startRecording()
    while (record.audioStatus(record.AudioStatus.Recording)) {
        basic.showIcon(IconNames.Heart)
    }
    basic.clearScreen()
})
input.onButtonPressed(Button.AB, function () {
    record.playAudio()
})
input.onButtonPressed(Button.B, function () {
    record.startRecording()
    while (record.audioStatus(record.AudioStatus.Recording)) {
        led.plotBarGraph(
        input.soundLevel(),
        255
        )
    }
    basic.clearScreen()
})

```

> makecode.microbit.org version:  [5.1.26](https://github.com/Microsoft/pxt-microbit/releases/tag/v5.1.26)
> Microsoft MakeCode version:  [8.6.33](https://github.com/Microsoft/pxt/releases/tag/v8.6.33)
> microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
> codal-microbit-v2 runtime version:  [v0.2.50](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.50)

srietkerk:
Thank you for finding this! I will work on getting the simulator to reflect the correct hardware behavior.

[...]

PR #5199, opened 2023-05-27
Sim mic indicator: sound level now accurately reflects what happens on hardware

srietkerk:
With the new CODAL updates, the mic indicator doesn't turn on for all of the mic blocks on the hardware. 

image

Doesn't trigger the mic indicator at all, which makes sense because this block just sets the threshold programmatically and has nothing to do with the board.

image
Triggers the indicator when it is actively in use, like in a while loop. Once the context is done, the indicator is no longer active

image
Mic indicator is always on as it is always polling for mic data.

This is now reflected in the simulator. 

**IMPORTANT**
Dependent on https://github.com/microsoft/pxt-common-packages/pull/1449

Closes https://github.com/microsoft/pxt-microbit/issues/5126
Upload Target: https://makecode.microbit.org/app/66914a615527944a087106de191146e33eefafd8-d93a826da4


kimprice (approved):
Looks good!

jwunderl (approved):

PR #5201, opened 2023-05-30
Recording extension: Sleeping in boolean blocks to prevent race condition in tight loops

srietkerk:
In the past, programs like the following would freeze the micro:bit board because of a race condition caused by non-preemptive scheduling. 

image

The root of the problem is that after the check is called the first time, the sound level will continue to poll for resources and be at the top of the calling queue, putting calls to the check behind the sound level poll and thus never getting further in the program. With the `uBit.sleep(0)` calll, we are saying that after waiting 0 time, put the function call at the top of the queue, which allows the program to continue running as expected.

### DIFF
--- a/libs/audio-recording/recording.cpp
+++ b/libs/audio-recording/recording.cpp
@@ -117,6 +117,7 @@ int audioDuration(int sampleRate) {
  */
 //%
 bool audioIsPlaying() {
+    uBit.sleep(0);
     return recording->isPlaying();
 }
 
@@ -125,6 +126,7 @@ bool audioIsPlaying() {
  */
 //%
 bool audioIsRecording() {
+    uBit.sleep(0);
     return recording->isRecording();
 }
 


jwunderl (commented):
So the issue is basically that this while loop prevents the actual recording code from running and deadlocks?

Yielding like this does have some downsides; the main one being that it would allow all other user code a chance to run as well, which is kinda the point but can cause hard to find race conditions; e.g. in https://makecode.microbit.org/S36810-38729-65788-56628, as written it's safe (not a meaningful program, but does what it says; always prints out either 6 or 107 - I threw the forever in there to give it a little more chance to 'hit' when yielding), but if we yield automatically, it's entirely possible for the other forever to 'mess up' the code that was running fine synchronously. I don't think there's really a great other option for this one though (without hacking around / deeply special casing it), but it does seem especially subtle since it'd only be applying to 2 out of the 4 options in the drop down. Maybe we just make it so they all behave the same (so full / stopped also gain a pause)? I suppose they could theoretically both run into the same issue as well

abchatra:
While loop is expected to be blocking if user doesn't add an explicit sleep in there. Otherwise control will go to other events. I think we should discard this PR and resolve the bug as by design.

libs/audio-recording/recording.cpp
```diff
@@ -117,6 +117,7 @@ int audioDuration(int sampleRate) {
  */
 //%
 bool audioIsPlaying() {
+    uBit.sleep(0);
```
abchatra, line 120: We shouldn't add sleep here. This API is basically returning the state of the audio playing.

srietkerk:
> While loop is expected to be blocking if user doesn't add an explicit sleep in there. Otherwise control will go to other events. I think we should discard this PR and resolve the bug as by design.

The problem is, though, that if you have a program like the one above, you will freeze the micro:bit completely. Yes, the while loop should be blocking, but having a program like that shouldn't cause the micro:bit to break.

abchatra:
Microbit will  freeze if student adds a true in the while loop condition right? This is no different in my opinion.

srietkerk:
> Microbit will freeze if student adds a true in the while loop condition right? This is no different in my opinion.

Ah, I see. Okay, makes sense. The same thing happens with a program like the following, so it is staying consistent. 
image

Just a note, though, that in the simulator, you can still play with the sound level in the while loop. I'm not sure if this is something we care about being accurate with since the board behavior is not what you would expect, but I wanted to note it in case that is something we might want to change or maybe alert about?

srietkerk:
It's important to inform the foundation of this as well as they might have been wanting to use these patterns in curriculum.

abchatra:
Yes lets review their curriculum. However aware of this semantics, this is true in their python editor as well. We have Forever loop and events for this reason. Forever has implicit sleep in each turn.

jwunderl:
Yeah, the important thing here is that it's easy to turn a synchronous loop into an async loop (just add the pause in the loop body), but it's not really really possible to convert an async check into a synchronous one ~

srietkerk:
Closing as this change would make while loop logic work differently with the recording extension. We want while loop logic to stay consistent no matter the context

Issue #5159, opened 2023-05-19
Share: Record GIF quality is pretty low


srietkerk:
When recording a GIF preview when you're about to share a project, the quality of the preview is reduced upon the recording being done.

**To Reproduce**
1. Go to share your game
2. Click update thumbnail
3. Click record
4. Stop Recording
5. Look at the sim preview versus the GIF preview.

![image](https://github.com/microsoft/pxt-microbit/assets/49178322/53d35332-83b1-42b7-860c-54d6b44815df)

**Expected behavior**
The sim preview and the GIF preview should have the same quality.

**micro:bit version (please complete the following information):**
5.1.30

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome

PR #5203, opened 2023-05-30
Improve gif quality for thumbnail

kimprice:
closes https://github.com/microsoft/pxt-microbit/issues/5159

This is still not perfect, but I think it's an improvement. The quality can get a little better with larger widths but there is a tradeoff because then there's room for less frames and it seems to take longer to render. Thanks to @jwunderl for helping me find the gif width field.

With the changes:
gif240

Before:
gif160

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -539,6 +539,7 @@
         "simScreenshot": true,
         "simScreenshotMaxUriLength": 300000,
         "simGif": true,
+        "simGifWidth": 240,
         "qrCode": true,
         "importExtensionFiles": true,
         "nameProjectFirst": true,


pxtarget.json
```diff
@@ -539,6 +539,7 @@
         "simScreenshot": true,
         "simScreenshotMaxUriLength": 300000,
         "simGif": true,
+        "simGifWidth": 240,
```
srietkerk, line 542: Is there a `simGifHeight` too? Also, why 240 for a value?

I agree, this resolution looks much better.

thsparks (approved):

jwunderl (approved):
does setting that `simGifQuality` field down to like 8 impact at all along side this change?

pxtarget.json
kimprice: There's no `simGifHeight`. The default is 160 (see https://github.com/microsoft/pxt/blob/affff2786d688cb276685a99294029aac46ae66c/webapp/src/components/SimRecorder.tsx#L128), so I tested values from there. I tested 200, 240, 320, 480, 750, and 1000. Numbers larger than 240 barely make any difference and then they limit the number of frames. And at 750 and 1000, the rendering seemed to slow down. So, it was basically trial and error.

kimprice:
> does setting that `simGifQuality` field down to like 8 impact at all along side this change?

Unfortunately, no it doesn't seem to do anything.

PR #5206, opened 2023-05-31
move webusb flashing over to async await

jwunderl:
Most of this has just been sitting / running on my machine for a couple weeks since I tried this a while back to see if there were any potential gains (e.g. if maybe while with async calls could end up being mildly faster than a ~15k deep tail recursive promise chain), but I didn't end up seeing any noticeable diff on my machine (both have a ~3-4 second range they fall into whenever you flash, so it's a little inconsistent). That said, I do think this way's a bit easier to follow along with, so if it doesn't impact perf probably worth merging (Also good to just leave some builds to compare with on some devices that are struggling to see if anything makes a difference)

### Build with this change

links are to the same build, just different queryParams
* normal https://makecode.microbit.org/app/2150b0cb682610a5c55c271580cf0836ccc6c180-2e44d552f4
* with dbg enabled (logs flash duration once it's over; dbg mode does incur a little runtime cost); https://makecode.microbit.org/app/2150b0cb682610a5c55c271580cf0836ccc6c180-2e44d552f4?dbg=1#editor
* and force full flash; https://makecode.microbit.org/app/2150b0cb682610a5c55c271580cf0836ccc6c180-2e44d552f4?dbg=1&webusbfullflash=1#editor

### build build without this change

built off of https://github.com/microsoft/pxt-microbit/pull/5200
* normal https://makecode.microbit.org/app/10404380f9dde2713665b9cd123597b2d079eda5-7ce748e574
* dbg enabled:  https://makecode.microbit.org/app/10404380f9dde2713665b9cd123597b2d079eda5-7ce748e574?dbg=1#editor
* force full flash https://makecode.microbit.org/app/10404380f9dde2713665b9cd123597b2d079eda5-7ce748e574?dbg=1&webusbfullflash=1#editor

for the debug builds, you'll want to add a filter to your console to see the duration:

image

and a note, browser process does start throttling a bit as soon as you click out of the page (e.g. into dev tools) & can take ~20% longer if you do so, so keep focus in the page for the duration of the flash if you want to compare. Can also toggle on preserve log under the cog wheel so they will stick around through a refresh / swapping between pages 
image

After 10 full flash's using each build, on my device with 4x cpu throttling enabled in dev tools each I came out to about 1.25 seconds faster on this branch, but that was over a ~ 38second average so, again, not likely anything noticeable but at least not harmful? Most of the time is spent waiting on usb i/o here https://github.com/microsoft/pxt-microbit/compare/asyncAwaitifyFlashing?expand=1#diff-d4cae301c7fe5ef74f432b0709c9b908c98be497386263cb498c73852bc847f5R495 so there isn't likely too much we can scrape off the full flashes here unless there is an async call we can extract from https://github.com/microsoft/pxt/blob/coreStartLoadingSpinner/pxtlib/webusb.ts#L343 or https://github.com/microsoft/pxt/blob/coreStartLoadingSpinner/pxtlib/webusb.ts#L394 (probably not, they're pretty minimal)

### DIFF
--- a/editor/flash.ts
+++ b/editor/flash.ts
@@ -376,11 +376,10 @@ class DAPWrapper implements pxt.packetio.PacketIOWrapper {
         await this.io.disconnectAsync();
     }
 
-<<<<<<< HEAD
-    async reflashAsync(resp: pxtc.CompileResult, progressCallback?: (percentageComplete: number) => void): Promise<void> {
-=======
-    reflashAsync(resp: pxtc.CompileResult, progressCallback?: (percentageComplete: number) => void): Promise<void> {
->>>>>>> 44344f190d5f30bdfd73cc258f7c207178c5362c
+    async reflashAsync(
+        resp: pxtc.CompileResult,
+        progressCallback?: (percentageComplete: number) => void
+    ): Promise<void> {
         pxt.tickEvent("hid.flash.start");
 
         log("reflash")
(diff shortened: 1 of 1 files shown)


thsparks (approved):
Much easier to follow!

PR #5207, opened 2023-05-31
Add tour steps for monaco editor

kimprice:
Depends on https://github.com/microsoft/pxt/pull/9535

This PR adds the tour steps for JavaScript and Python projects. Only visible elements are added to the tour, so there should be no duplicate toolbox or workspace steps in the tour.

### DIFF
--- a/docs/tours/editor-tour.md
+++ b/docs/tours/editor-tour.md
@@ -20,10 +20,22 @@
 * highlight: toolbox
 * location: right
 
+## Toolbox
+* title: Toolbox
+* description: Drag out snippets of code from the Toolbox categories into the Workspace.
+* highlight: monaco toolbox
+* location: right
+
 ## Workspace
 * title: Workspace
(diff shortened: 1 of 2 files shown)


jwunderl (approved):
lgtm, though we should probably wait to merge until after the pxt side has been merged / bumped into beta~

jwunderl:
It doesn't matter much right now as it's still just in beta, but one thing to be careful of; this is a case where strictly speaking you'd want to separate out the code changes (version bump) from the docs changes in a separate pr -- the code change will always take a little extra time to get out (for beta probably only 20-30 minutes if you bump immediately after merging, but for live would likely be a few days unless you're rushing out a fix), but the markdown changes will be available immediately besides the cdn cache (which is on average ~ 24hrs, but could hit any time if you're unlucky).

Not something to worry about right now, but something to keep in mind for future changes once released / etc - the code has to be released before you commit a change to an existing markdown file.

Issue #3777, opened 2020-11-27
onSound triggers constantly if playing melody on the spseaker


microbit-mark:
**Describe the bug**

This program triggers the sound playback constantly

<img width="479" alt="image" src="https://user-images.githubusercontent.com/31242877/100466052-34102d80-30c8-11eb-85b9-f0bba1c2c64e.png">

Setting a higher threshold resolves this
<img width="371" alt="image" src="https://user-images.githubusercontent.com/31242877/100466130-543fec80-30c8-11eb-8849-73fa02864563.png">

But `set threshold` is not a kid-friendly block for first using the microphone and it's hidden away in Input > more

Could we change the default threshold so it does not trigger so easily.

...or  we could have a mic on/off block
... or lower the volume of the melody

pelikhan:
We should auto turn off the microphone when I paging a sound

________________________________

jaustin:
@finneyj is there an easy way just to ignore mic events during sound playing? I would rather not actually turn off the microphone

finneyj:
Sounds like a "learning opportunity" about feedback loops.. :)

We could suppress the generation of events, for sure... but I think there's a problem here around making too many assumptions... Kids won't expect the microphone to be enabled/disabled without them knowing.

There could be use cases where disabling the mic events could break kids code. e.g. clap to stop sound? Music playing in a game, and a clap to activate a game activity? etc...  

Alternative options:

- Preserve the current behaviour and encourage kids to manually set threshold block anyway. Can we do tooltips if we see a block that makes a sound inside an "onLoud" handler? 

- Try to dynamically adjust the onLoud threshold based on sound output level. We calibrate for this (as all micro:bit are more or less created equal), but only for sound that go through the mixer (anyone sneaking around that to do old school v1 PWM sounds wouldn't work). Also if kids/teachers were using external speakers, that calibration would be no good, and you'll likely still get the same feedback loop. This would be less of a common case though I guess.

- Add some sort of feedback suppression filter in software, like you do on your phone/laptop. I don't think we have the developer resource or CPU cycles to do that though.

pelikhan:
A microphone mute block would match perfectly what they already use in Zoom/Teams/… every day.

finneyj:
yep - mute would work.

jaustin:
Right, Joe and John implemented a function in CODAL to enable this. CODAL now has an 'is sound playing' 

[...]

PR #5218, opened 2023-06-03
Sound is playing block added to the volume category for music blocks

srietkerk:
When using the `on loud sound` block, if you are also having audio play through the speakers, it often triggers the `loud sound` on hardware. The new `is sound playing` block will help in situations like this. For users who want this outcome of playing sounds and triggering `on loud sound` this will not effect their experience. However, if users don't want that experience, they can use this block with some logic in `on loud sound`, and the experience will better align with what is desired.

What it looks like in the bin:
image
 

Example:
image

**Dependent on** https://github.com/microsoft/pxt/pull/9541

Closes https://github.com/microsoft/pxt-microbit/issues/3777
Upload target: https://makecode.microbit.org/app/ec7f196399d22b881359b4b47f07da2d2b50783c-242591ac95


sim/state/music.ts
```diff
@@ -11,4 +11,10 @@ namespace pxsim.music {
     export function setSilenceLevel(level: number) { 
         // ignore in v1,v2
     }
+
+    export function isSoundPlaying(): boolean {
+        const audioActive = pxsim.AudioContextManager.isAudioElementActive();
+        const soundExpressionPlaying = pxsim.codal.music.isSoundExpPlaying();
+        return audioActive || soundExpressionPlaying;
```
jwunderl, line 18: Does this also need to check for whether a microphone recording is playing?

jwunderl (approved):
LGTM, though I think it might be missing a check for whether the microphone recording is active?

sim/state/music.ts
srietkerk: Yes. Thank you for catching this. I remembered at some point today and then completely forgot.

kimprice (approved):
Looks good!

thsparks (approved):

libs/core/music.cpp
```diff
@@ -63,6 +63,21 @@ void setBuiltInSpeakerEnabled(bool enabled) {
...
+*/
+//% blockId=music_sound_is_playing block="sound is playing"
+//% group="Volume"
+//% help=music/volume
+//% weight=0
+bool isSoundPlaying() {
+#if MICROBIT_CODAL
+    uBit.audio.isPlaying();
+#else
+    target_panic(PANIC_VARIANT_NOT_SUPPORTED);
```
abchatra, line 77: I realized we don't support v1. We should probably group this block with v2 only blocks.

PR #5220, opened 2023-06-05
Recording extension: putting more distance in between gain levels

srietkerk:
With the current iteration of the blocks, It's hard to tell the difference between medium and high mic sensitivity. The changes in this PR create a larger difference. There might be more arguments to make the distance between the levels even greater, but I feel strongly that these are the best values to tell the difference between medium and high sensitivity while also making it so the audio isn't blowing out the speaker completely (high is already on the verge of doing so).

Note: for the low sensitivity level, I just took away setting the gain in the switch statement because the default is already low and I don't really want to touch it. I wasn't sure, though, if I can just take out the case completely. I can take it out if it's safe to do so.

Closes https://github.com/microsoft/pxt-microbit/issues/5063

### DIFF
--- a/libs/audio-recording/recording.cpp
+++ b/libs/audio-recording/recording.cpp
@@ -93,13 +93,12 @@ void erase() {
 void setMicrophoneGain(int gain) {
     switch (gain) {
     case 1:
-        uBit.audio.processor->setGain(0.079f);
         break;
     case 2:
         uBit.audio.processor->setGain(0.2f);
         break;
     case 3:
-        uBit.audio.processor->setGain(0.4f);
+        uBit.audio.processor->setGain(1.0f);
         break;
     }
 }
--- a/libs/audio-recording/recording.cpp
+++ b/libs/audio-recording/recording.cpp
@@ -90,17 +90,8 @@ void erase() {
  * Set sensitity of the microphone input
  */
 //%
-void setMicrophoneGain(int gain) {
-    switch (gain) {
-    case 1:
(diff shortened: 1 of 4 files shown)


libs/audio-recording/recording.cpp
```diff
@@ -93,13 +93,12 @@ void erase() {
 void setMicrophoneGain(int gain) {
     switch (gain) {
     case 1:
-        uBit.audio.processor->setGain(0.079f);
```
jwunderl, line 96: This still needs to set the value even if it's setting it to the same value as the codal default I believe? otherwise

```typescript
record.setMicGain(AudioLevel.High);
record.setMicGain(AudioLevel.Low);
```

would leave us at a high gain. (e.g. if someone wanted to loop over and find the lowest support gain that finds noise in environment)

libs/audio-recording/recording.cpp
```diff
@@ -93,13 +93,12 @@ void erase() {
...
         break;
     case 2:
```
jwunderl, line 97: Doesn't need to happen right away, but maybe worth hoisting the switch statement here up to `setMicGain`, and just having `setMicrophoneGain` take and pass a float to the audio processor directly; that way fiddling with these values can occur without having to wait for a full cpp cloud compile, and also leaves an escape hatch for if the provided gain options aren't working for particular device or environment and they need to make e.g. an extension to make it more usable.

abchatra (approved):

PR #5221, opened 2023-06-05
Add block validation for a few tutorials

### LINKED ISSUE #5169: Add code validation rules
Hi Galen, can you help us add code validation to a few of the micro:bit tutorials?
@kiki-lee and @thsparks know how to do this and can show you...

ganicke:
Block validation is added for some select tutorials present in the "Tutorials" and "V2 Tutorials" galleries on the home screen.

Closes #5169


thsparks:
Just a note - before we check this in, I want to validate that it won't break the tutorials on our live site (where they're still using the old tutorial infrastructure, which doesn't do validation). I'm fairly confident it should be fine (code validation will just get ignored), but I want to test it to make sure.

docs/projects/v2-pet-hamster.md
kiki-lee, line 1: Should //@highlight be one line lower here?

kiki-lee: (After line 28 instead of before?)

kiki-lee: Okay, I opened the accordion and it looks like it's needed in both places.

kiki-lee, line 1: I believe it needs //@highlight both before and after line 28 to check for both blocks.

thsparks:
Just verified, and unfortunately with the current live version of Micro:bit (which doesn't have the code to parse the validation), it does not actually get ignored like I thought it would. Instead, we'll see "#blocksexistvalidator" text at the end of the tutorial (inside a hint on the final step). We'll either need to do a hotfix to change that, or we should wait to commit this until after the Micro:bit release.

image

ganicke:
Thanks @thsparks and @kiki-lee for reviewing! This PR could go on ice until the release in order to resolve that render problem.

An interesting observation is that it seems the first `// @highlight` will short circuit the check on any remaining blocks specified in the snippet. For example, in the **Giggly Cyrus** step in `v2-pet-hamster`, the second validation isn't checked...

````
```blocks
//@highlight
input.onLogoEvent(TouchButtonEvent.Pressed, function () {
    //@highlight
    basic.showIcon(IconNames.Happy)
})
```
````

...resulting in a pass even if `basic.showIcon(IconNames.Happy)` is missing.

kiki-lee:
I believe the update will take care of that, since I think that the fix that happened for Arcade was in the PXT.

abchatra:
Good time to update this PR and release.

thsparks (approved):
Thanks for going through all of these!

Issue #5219, opened 2023-06-05
migration path from `start melody <builtin melody>` to `play <builtin melody>` is not clear


jaustin:
**Describe the bug**

The refactoring of the various `play <stuff>` blocks doesn't leave a clear and obvious choice for people following tutorials that use the old `start melody` block (not the 'play melody' block, that's fine).

There are a lot of tutorials, not least on microbit.org including videos that use the existing 'start melody' block. We have asked a number of authors about this and one reason for it's popularity is the ease of conveying a specific tune in text/pictures - something that's much harder with the 'play melody' block that is entirely visual. (Eg, if you want a 'yes' and a 'no' sound people use 'jump up' and 'jump down', but it's much harder to quickly convey that with the visual 'play melody' sequencer)

It's important to us that people following older books and tutorials can still find their way after the changes, and there are two things that I think stand out in making this difficult.

1. The wording ('melody' is now lost)
2. The default behaviour is changed from async to sync

**proposed solutions**

**1. the word 'melody' is not present** in the replacement for the old 'start melody <named melody>' block, so both 'start' and 'melody' are missing. 
--> I propose we resolve this by finding a way to insert 'melody' into the new block structure:

Original:
<img width="375" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/147113/475a8741-d5a3-4de5-9eab-09d97018e6f4">
New:
<img width="289" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/147113/6dd57416-902e-473a-800a-0d9de43e2037">

Proposed to resolve this:
<img width="293" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/147113/494dd95e-4fe4-4e98-b16d-d4f468bb86a3">

**2. The default options are different (blocking vs no-blocking)**
As per #3823 we have long had a weird difference in default behaviour between 'start melody' and 'play melody' ( I think justified by the different language - starting (in background)?). Ultimately it was a bit unclear.

This change causes more confusion now though, because the default for the new `play melody <built in melody>` block is that it blocks, not plays in the background. 

Here's an example of something that will change:
OLD - dancing note animation during melody:
<img width="216" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/147113/c009161d-82bf-43d8-942a-65bfa2503e5b">
NEW - animation plays after song.
[...]

PR #5228, opened 2023-06-08
Melody Block Updates to Align with Previous Blocks

thsparks:
The goal is to align the new melody blocks added in https://github.com/microsoft/pxt-microbit/pull/5123 (specifically the "built in melody" block) to something that resembles the previous blocks a bit more closely. See https://github.com/microsoft/pxt-microbit/issues/5219 for more details on why.

Specific changes were to include "melody" before the melody name and to change the default to "in background". Since we don't want to change all blocks to use "in background" by default, I created a copy of the play block and used that instead. I don't love this approach, as it makes the javascript look pretty weird, but it gets the job done and a cleaner solution would require more time than we'd like to commit to this change.

Original block (before https://github.com/microsoft/pxt-microbit/pull/5123):
image

Block after https://github.com/microsoft/pxt-microbit/pull/5123 but without these changes:
image

Block with these changes:
image
image

Upload target: https://makecode.microbit.org/app/6634ff49a8e295843baba5736a89f2f4861fa8b0-cdb36cab85

Fixes https://github.com/microsoft/pxt-microbit/issues/5219.

### DIFF
--- a/libs/core/playable.ts
+++ b/libs/core/playable.ts
@@ -86,6 +86,17 @@ namespace music {
         toPlay._play(playbackMode);
     }
 
+    //% blockId="music_playable_play_default_bkg"
+    //% block="play $toPlay $playbackMode"
+    //% toPlay.shadow=music_string_playable
+    //% playbackMode.defl=music.PlaybackMode.InBackground
+    //% group="Melody"
+    //% help="music/play"
+    //% blockHidden
+    export function _playDefaultBackground(toPlay: Playable, playbackMode: PlaybackMode) {
+        return play(toPlay, playbackMode);
+    }
+
     //% blockId="music_string_playable"
     //% block="melody $melody at tempo $bpm|(bpm)"
     //% weight=85 blockGap=8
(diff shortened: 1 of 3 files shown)


jwunderl (approved):

libs/core/playable.ts
```diff
@@ -86,6 +86,17 @@ namespace music {
         toPlay._play(playbackMode);
     }
 
+    //% blockId="music_playable_play_default_bkg"
+    //% block="play $toPlay $playbackMode"
```
abchatra, line 90: I am wondering if we could just say "start $toPlay $playbackMode" here?

thsparks: Do you mean instead of adding "melody"?

I think the concern, as I understood it from the bug, was the lack of the word "melody" more so than the change from "play" to "start". But @jaustin may have more thoughts on it.

abchatra: Can we do both? Like change play to start and add the word "melody"?

thsparks: Yeah, we can do that.

abchatra: However I agree it would be ugly if someone drops "sound effect" in the "start" block and gets confused how they arrived at that stage.

abchatra (approved):

Issue #5223, opened 2023-06-07
Audio always is full no matter how long the recording takes


Sheila111:
**Describe the bug**
Audio always is full no matter how long the recording takes

**Steps to reproduce the behavior**
1. Navigate to https://makecode.microbit.org/beta#
2. Create a project and add extension '**Record**'
3. Drag record block or import this project: https://makecode.microbit.org/_dhJAT3gF17CV

**Actual behavior**
Audio always is **full** no matter how long the recording takes.

**Expect behavior**
The audio will **stop automatically** when the buffer is **full**. In another word, audio would be full only when you didn't interrupt it. 

**Screenshot**
![1](https://github.com/microsoft/pxt-microbit/assets/59192069/6420a06d-d3e8-4131-8c32-3efe1be282f4)

**Additional context**
1.OS: Windows11
2.makecode.microbit.org version: 5.1.37
3.Microsoft MakeCode version: 8.6.48

PR #5232, opened 2023-06-12
Recording extension: Changed audio is full to audio is empty

srietkerk:
Tracking whether the audio buffer is full is a bit nuanced as it doesn't mean anything to the user. When starting a recording, the previous recording is erased. A user can't add to the current audio buffer.  The metric that might be useful to some users is to check if there is something in the audio buffer at all, which is what `audio is full` was tracking with confusing wording.

Because `audio is full` doesn't really provide value to the user, the block is now `audio is empty` for a usage where the user might want to check if the buffer doesn't have anything in it before recording something. The following is an example program:
image

Closes: https://github.com/microsoft/pxt-microbit/issues/5223

### DIFF
--- a/libs/audio-recording/_locales/audio-recording-strings.json
+++ b/libs/audio-recording/_locales/audio-recording-strings.json
@@ -12,7 +12,7 @@
   "record.AudioSampleRateScope.Everything|block": "everything",
   "record.AudioSampleRateScope.Playback|block": "playback",
   "record.AudioSampleRateScope.Recording|block": "recording",
-  "record.AudioStatus.BufferFull|block": "full",
+  "record.AudioStatus.BufferEmpty|block": "empty",
   "record.AudioStatus.Playing|block": "playing",
   "record.AudioStatus.Recording|block": "recording",
   "record.AudioStatus.Stopped|block": "stopped",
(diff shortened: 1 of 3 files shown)


jwunderl (approved):

abchatra (approved):

abchatra:
@jaustin fyi

jaustin:
@srietkerk does this mean you can't (or in fact, could never) use a while loop 'until recording is full'? How would you propose someone finds the exact length of their recording in the current setup? Something like this?
image

srietkerk:
> @srietkerk does this mean you can't (or in fact, could never) use a while loop 'until recording is full'? How would you propose someone finds the exact length of their recording in the current setup? Something like this? image

@jaustin Yes, that's correct, with how it was implemented, you would not have been able to `record until full` at any point of the extension being live. And the blocks you have look like a good solution for what you're looking for. 

There are a couple of things to note, as well. As I explained above, `isFull` is not a very valuable metric to users as what you can do with the buffer whether it is completely full or partially full does not change. Experimentation will allow users to know how long they can record if they want to record for as long as they possibly can.

As for getting recording length, yes, those blocks look like a good solution. I know that there is a CODAL API to get the duration/length of the recording, but I'm not sure I see the value in adding a block that gets the duration of the recording. I would prefer to get some user testing and receive feedback to see if that is a desired block.

PR #5234, opened 2023-06-13
Thumbnails updated for recording, radio broadcast, and servo extensions

srietkerk:
Replacing old thumbnails for the above extensions because we got new art for them from the foundation.

Need to do the same in neopixel, microturtle, and sonar repos.


srietkerk:
This is what they look like in the extensions grid.
image

jwunderl (approved):

PR #5251, opened 2023-06-16
Make another cpp change

eanders-ms:
... to generate a new sha


abchatra:
Do it directly in stable6.0?

eanders-ms:
Is that easier for testing? Needs to be in both places anyway, no?

abchatra:
Yes needs to be there both places. I was suggesting just in case you were in a hurry.

Issue #5252, opened 2023-06-16
offline app: built-in extensions should work offline (audio recording, for example)


eanders-ms:
1. Run the offline app without internet connection.
2. Make a simple program using the audio recording blocks.
3. Click download.

Notice the compilation error message.

The audio recording/playback blocks are not usable without an internet connection.

Suggestion: precompile and include hex files for builtin extensions.

jwunderl:
Ahh I think it's probably not checking for https://github.com/microsoft/pxt-microbit/blob/master/libs/audio-recording/pxt.json#L14 when building or something, that's the only thing I can think of as different vs datalogger; code is here https://github.com/microsoft/pxt/blob/master/cli/cli.ts#L2441, can point you to where it builds in offline app devops build.

jwunderl:
Combination of https://github.com/microsoft/pxt-microbit/pull/5259 and https://github.com/microsoft/pxt/pull/9579 fixed this (5259 for removing disables variant, 9579 for packing in a matching hex file correctly)~

PR #5259, opened 2023-06-16
audio recording: remove disable variant & just give error on v1

jwunderl:
I was looking at this re: https://github.com/microsoft/pxt-microbit/issues/5252 but I believe we want this independent of that issue / haven't tested with that yet as it'd be more of a 'bonus' if it is the cause. Right now the audio recording extension forces into v2 only compilation mode, instead of just erroring when the blocks are used with the known error code; this should fix that. (lemme know if anyone has strong opinions on if we should ignore some of the errors or make any of the ones that are currently just giving a trivial return into an error instead)

build: https://makecode.microbit.org/app/b356e9e3ceca7810a104edb6259d9274a83cad6b-554c09a379#editor

It'll still give the incompatible code if you plug into v1 microbit and include recording blocks, but if that codepath is never reached (e.g you put them in a+b pressed or behind an if that doesn't run on v1 microbits) they'll run just fine.


abchatra (approved):

srietkerk (approved):
Everything looks great; everything compiled and worked perfectly on the micro:bit, too. Thanks, Joey!

PR #5274, opened 2023-06-20
Hide step headings in some tutorials

### LINKED ISSUE #5266: With the new micro:bit layout, we need to hide all the step titles using { }
@ganicke  Would you be willing to go through the prominent micro:bit tutorials and hide all of the step titles?  They're too large for the small amount of space in the instructions (and generally, they are only "Step 1", "Step 2", etc.)

Screenshot by Dropbox Capture

## {Step 1}

We need 3 pieces of code: one to detect a throw (shake), another to pick a random number, and then one to show the number.

Use the ``||input:on shake||`` block you see in the editor workspace. It runs code when you shake the @boardname@.

ganicke:
Ghost the step headings in the featured tutorials in order save room for other text to display in description pane.

Closes #5266

### DIFF
--- a/docs/projects/7-seconds.md
+++ b/docs/projects/7-seconds.md
@@ -1,6 +1,6 @@
 # 7 seconds game
 
-## Introduction @unplugged
+## {Introduction @unplugged}
 
 The goal of this game is press a button after **exactly** 7 seconds!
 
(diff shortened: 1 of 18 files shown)


kiki-lee (approved):
Awesome. Thanks!

PR #5276, opened 2023-06-20
Extension: add FWD Edu Climate Action Kit Gen. 2 Kit

martinwork:
Arising from https://support.microbit.org/helpdesk/tickets/64120 (private)
@ross-inksmith

@abchatra
This extension is in beta. I'm not sure how that should be flagged.

https://makecode.microbit.org/_VULexMCbyd4W

image

image


abchatra:
@martinwork I wouldn't put any tags if it is still in beta, so this doesn't come up when someone clicks "Science"
Folks still can search it and use it.

martinwork:
@abchatra OK, thanks, I removed the tag.

abchatra (approved):

Issue #5158, opened 2023-05-19
Clicking mute button in safari is not unmuting the simulator


riknoll:
Seems like Safari is getting even more strict about this! If you click inside the iframe (like on the microbit itself) then the sound starts playing. To work around this, we will likely need to add an unmute button that is inside the iframe itself.

Tested in Safari Version 16.4 (18615.1.26.110.1)
On macOS 13.3.1 (a)

abchatra:
I vaguely remember this being done before in arcade.

PR #5277, opened 2023-06-20
Add a mute button inside the simulator iframe

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5158
Requires https://github.com/microsoft/pxt/pull/9568

This PR adds an unmute button inside of the simulator iframe in Safari. Unmute only needs to be hit once per iframe; after you click it the outside mute button will be enabled and the inner button will disappear. I put the button up in the upper-right hand corner (every other inch of the micro:bit sim is taken at this point). Also, this button does not appear on the mini-sim, only the fullscreen and normal sized sims. 

Screenshot 2023-06-20 at 4 22 44 PM

The build will fail until that required pxt PR is merged and bumped.


sim/visuals/mute.ts
```diff
@@ -0,0 +1,63 @@
+namespace pxsim {
+    const icon = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 512 512"><!-- Font Awesome Pro 5.15.4 by @fontawesome - https://fontawesome.com License - https://fontawesome.com/license (Commercial License) --><path d="M215.03 71.05L126.06 160H24c-13.26 0-24 10.74-24 24v144c0 13.25 10.74 24 24 24h102.06l88.97 88.95c15.03 15.03 40.97 4.47 40.97-16.97V88.02c0-21.46-25.96-31.98-40.97-16.97zM461.64 256l45.64-45.64c6.3-6.3 6.3-16.52 0-22.82l-22.82-22.82c-6.3-6.3-16.52-6.3-22.82 0L416 210.36l-45.64-45.64c-6.3-6.3-16.52-6.3-22.82 0l-22.82 22.82c-6.3 6.3-6.3 16.52 0 22.82L370.36 256l-45.63 45.63c-6.3 6.3-6.3 16.52 0 22.82l22.82 22.82c6.3 6.3 16.52 6.3 22.82 0L416 301.64l45.64 45.64c6.3 6.3 16.52 6.3 22.82 0l22.82-22.82c6.3-6.3 6.3-16.52 0-22.82L461.64 256z"/></svg>`;
```
jwunderl, line 2: Did you grab the right variant here? when I check out the mute it shows font awesome free in the licensing (though it's showing the v6 license even though I'm pointing at v5, so maybe just something weird)

jwunderl (approved):

sim/visuals/mute.ts
jwunderl: Okay yeah, I get the same result when I press download even though it shows as free / not pro when I click the sg preview right next to it, must just be their default license generation (it doesn't let me even see the svg or download for the non free one so definitely the right thing). Seems good!

microbit-carlos:
This unmute button appears in Firefox as well, is it necessary on that browser too?

riknoll:
@microbit-carlos I don't think so... Firefox must have changed it's user agent to be closer to Safaris

Issue #5268, opened 2023-06-19
Recording extension: play request interrupted error 


srietkerk:
<img width="289" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/49178322/20494052-8524-4c41-aa27-62a3db4c4ce7">

Having a program like this and spamming "A" and then spamming or sometimes even just pressing "B" a couple of times will give you a `The play() request was interrupted by a call to pause()` error. We need to make sure that this error is handled, because if it doesn't get handled, recording and playback in the simulator stops working in certain browsers.

<img width="483" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/49178322/97b97514-ac5f-42c1-8f34-34cf5397224a">

**To Reproduce**
1. Make the above program
2. Open the console 
3. Spam click "A" to trigger a lot of recordings in a row
4. Spam click "B" 
5. See error

**Expected behavior**
Spam clicking should not cause any problem. Everything should clean up nicely each time a recording is triggered to start and triggered to play.

**micro:bit version (please complete the following information):**
6.1.2

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome, Firefox

PR #5283, opened 2023-06-23
Recording extension: simulator errors fixed

srietkerk:
There were a couple of errors happening when recording that we want to avoid. 

The first was `The play() request was interrupted by a call to pause`. This would basically cause recording ability to crash, and maybe the whole simulator which isn't great. I was neglecting the fact that a call to `play` would return a promise, and thus wouldn't make sure that play is resolving which is largely what the `createPlayPromise` logic handles. There is some weirdness here, though, where we really shouldn't mind that the play call gets interrupted by pause, so, in those cases, we just resolve and continue about our day.

Another was a race condition between playing and erasing that is now resolved by making sure to wait for the erase to finish before creating another recording to reference. With spamming, the calls to erase, record, and playback would get really jumbled and there were some times when a call to play would trigger after the resources for that recording to be played were getting removed. This would cause an error on a network request for the blob url. This isn't great because we can't just catch those errors, even though they likely won't cause a problem to the user but are kinda scary to see in the console. With awaits and a couple of null checks, this error is removed.

I also caught an error in Mozilla where the recorder would be nullish when trying to remove it. I added a null check to prevent this as well.

Just a note, I'm satisfied at where this implementation is in terms of not erroring, cleaning up state, and giving feedback to the user. However, as of now, the simulator implementation and micro:bit behavior differ when you spam click buttons. I haven't been able to investigate if I can get it closer to micro:bit implementation, but I think the fact that there are timeouts in the simulator code will make it difficult to get the simulator closer to the micro:bit behavior. 

Closes https://github.com/microsoft/pxt-microbit/issues/5268

### DIFF
--- a/sim/state/record-audio.ts
+++ b/sim/state/record-audio.ts
@@ -6,7 +6,6 @@ namespace pxsim  {
         chunks: Blob[];
         audioURL: string;
         recording: HTMLAudioElement;
-        playPromise: Promise<any>;
         audioPlaying: boolean = false;
         recordTimeoutID: any;
         currentlyErasing: boolean;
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
Did a few minor edits; not persisting the promise that we never look at again / just inlining it, and only adding `onStopAll` handler once since that was leaking a bit.

While looking it over I found a few places I think we can simplify a bit (e.g. in record audio we should be able to remove the `onstop` listener, make `populateRecording` synchronous again, and just call at the end of `stopRecorder`, and then we should be able to get away without e.g. the setTimeout in play and mute having to be async-ish so it doesn't get a race condition with play. If I end up with some extra time I can mock that out real quick this week, but doesn't feel necessary to push that into the release since it's mainly just refactoring.

here's a build / program to test with https://makecode.microbit.org/app/13046bc4c210251dbfc4e9785c3490aa652c55d3-c09041c406#pub:_dHC0tvLkji0m

thsparks (approved):

Issue #5273, opened 2023-06-20
Recording extension sim: use more generally accepted encoding


srietkerk:
On Safari, the playback for the audio recording doesn't work. This is likely because the encoding that is used currently, `audio/ogg` , is not fully supported on Safari. We should change it to something like `audio/mpeg` that is more universally supported.

PR #5288, opened 2023-06-26
fix recording on safari

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/5273

Tested on my ipad and works fine when mp4~ build, live site not working https://makecode.microbit.org/app/6d848023ecb33036f3280056dfd998945686ca56-0a6a88b1df

(note that this build does not include https://github.com/microsoft/pxt-microbit/pull/5283, so racing through the buttons has some issues / just let it record on this build.)

### DIFF
--- a/sim/state/record-audio.ts
+++ b/sim/state/record-audio.ts
@@ -40,7 +40,8 @@ namespace pxsim.record {
     }
 
     function populateRecording(b: DalBoard) {
-        const blob = new Blob(b.recordingState.chunks, { type: "audio/ogg; codecs=opus" });
+        const recordingType = pxsim.isSafari() ? "audio/mp4" : "audio/ogg; codecs=opus";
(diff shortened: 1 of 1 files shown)


eanders-ms (approved):
LGTM! I'm curious, why do we prefer ogg on other platforms?

jwunderl:
My understanding is ogg is typically used for better quality (though that may be considered a downside if we want to match actual hardware) -- would have to ask Sarah if we had a particular reason, but coverage is pretty good for it / likely came from following https://developer.mozilla.org/en-US/docs/Web/API/MediaStream_Recording_API/Using_the_MediaStream_Recording_API#grabbing_and_using_the_blob, but it feels like every example I've seen for mediarecorder api ends up using a different encoding. Main reason I'm keeping it as is for now is to avoid last minute change when it appeared to be working everyone else it needed to~

Issue #5294, opened 2023-06-27
Audio recording playback occurs even when muted


jwunderl:
**Describe the bug**

playback from audio-recording occurs even when simulator is muted, and is not stopped when simulator becomes muted.

**To Reproduce**
1. Go to https://makecode.microbit.org/_ahh3vYRmmiKU (or open in stable, both repro)
2. mute the simulator
3. record audio (press a, confirm allowed, and say a few things)
4. play back audio (press b)

**Expected behavior**

No sound should play, though ideally could still have a timeout for the duration to simulate actual behavior

PR #5295, opened 2023-06-27
respect muted state when playing audio recording

jwunderl:
fix https://github.com/microsoft/pxt-microbit/issues/5294

This change only checks when it starts, so if you mute while it's going the sound will still continue; the `onStopAll` in `registerSimStop` above is not called in that case, since that's for stopping sound (e.g. via the block or restarting program) vs. just muting them -- I don't see a good place to latch on right away for muted events - for multiplayer I added https://github.com/microsoft/pxt/blob/stable9.0/pxtsim/simlib.ts#L240, and can either use that here or clean it up and reuse, but this feels like a small problem to me (at most a few seconds) for now / can fix more generally in a followup rather than making release wait on it? 

test build / project https://makecode.microbit.org/app/aafb0a471874194c4e2855f8411ab679f0d96b37-630803537c#pub:_TyjP6eLE06b7

### DIFF
--- a/sim/state/record-audio.ts
+++ b/sim/state/record-audio.ts
@@ -130,7 +130,7 @@ namespace pxsim.record {
                 b.recordingState.recording.removeEventListener("play", b.recordingState.handleAudioPlaying);
                 b.recordingState.recording.removeEventListener("ended", b.recordingState.handleAudioStopped);
             }
-        })
+        });
     }
 
     export function play(): void {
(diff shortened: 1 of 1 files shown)


eanders-ms (approved):
Is `b.recordingState.recording.volume` in the range [0..1] ?

thsparks (approved):

jwunderl:
Yes, here's the docs: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/volume

PR #5301, opened 2023-06-30
support clever by default

jwunderl:


eanders-ms (approved):

jwunderl:
Yes, going to push this out just to beta for now (no clever=1 flag) and we can consider hotfixing it out later

PR #5310, opened 2023-07-09
Extension: add Cutebot Pro

martinwork:
Arising from https://support.microbit.org/helpdesk/tickets/64164 (private)

@elecfreaks1
Looking at the picture below, I noticed some blocks say "get". For example, I think "get speed" could be simply "speed".

@abchatra
Do you think the Cutebot Pro category has too many blocks?

https://makecode.microbit.org/_cW0bjY0rXEvr

image


elecfreaks1:
Yes, this is indeed a good suggestion, which can make the block very concise, and I will modify it later

abchatra (approved):

PR #5316, opened 2023-07-20
Bumped CODAL to v0.2.56 and the samples repo to v0.2.13

JohnVidler:
Updated to the latest CODAL with bugfixes for some outstanding issues with sound recording, the level detector and other smaller bugs. Assuming this all builds just fine, we would like to take this one directly to live.

This also includes the BLE utility service, but is disabled by default. I'll create a new PR with it enabled for Beta.

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -197,12 +197,12 @@
                 "codalTarget": {
                     "name": "codal-microbit-v2",
                     "url": "https://github.com/lancaster-university/codal-microbit-v2",
-                    "branch": "v0.2.55",
+                    "branch": "v0.2.56",
                     "type": "git"
                 },
                 "codalBinary": "MICROBIT",
                 "githubCorePackage": "lancaster-university/microbit-v2-samples",
-                "gittag": "v0.2.12",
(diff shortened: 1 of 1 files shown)


jwunderl (approved):
diffs: 

* https://github.com/lancaster-university/codal-microbit-v2/compare/v0.2.55...v0.2.56
* https://github.com/lancaster-university/codal-core/compare/bf65ebed2003049bd026e1f685bb756e189c9258...763a2d1a916d4db3fe3f3f2b1eaa9947cc8d0a7e
* https://github.com/lancaster-university/microbit-v2-samples/compare/v0.2.12...v0.2.13

@srietkerk do you have context on what was fixed for audio recording here? I can push this out to stable so you can test on hardware this afternoon if so

JohnVidler:
@jwunderl - The major fix here was with memory management, moving from a fixed-number-of-blocks approach to a fixed-amount-of-RAM approach. Functionally similar for constant sample rates, but changing the sample rate now should result in longer/shorter recording times, as would be expected.

PR #5261, opened 2023-06-17
Audio Recording extension reference pages

ganicke:
Add a base set of reference pages and a top level reference page to satisfy the "Learn more" link destination.

Closes #5222, closes #5240

PR #5322, opened 2023-07-24
Add some edits to clarify parms for audio settings

ganicke:
Add a few edits as requested in the comments for #5261.

Related to https://github.com/microsoft/pxt-microbit/pull/5261#discussion_r1269720366, the reason for the two copies of the extension main page is to satisfy the link for the "Learn More":

image

...which will use "./audio-recording" from the extension name while the reference doc space will use the "./record" path as per the namespace.

I don't think a _audio-recording-ref.json_ will work here with something like:

```json
{
    "redirect": "./record"
}
```

### DIFF
--- a/libs/audio-recording/docs/reference/record/audio-status.md
+++ b/libs/audio-recording/docs/reference/record/audio-status.md
@@ -1,12 +1,12 @@
 # audio Status
 
-Check to see if an audio buffer status is true or not.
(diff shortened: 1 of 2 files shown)


jwunderl (approved):

srietkerk (approved):
Looks great, thanks so much, Galen!

PR #5362, opened 2023-09-28
Bumped codal version to v0.2.62

JohnVidler:
- A patch to prevent out-of-memory errors in certain situations when using the audio recording API

Once this PR is built and tested, I'll mark this for review to go to beta and live.


JohnVidler:
Tag built with errors. Disregard this build.

PR #5401, opened 2023-11-08
Remove stale sections approveRepos & preferredRepos

abchatra:
We have new section approvedRepoLib which handles this. 

This will break extensions in old editors such as v4/v3 etc. We don't care about it at this point of time as they have workaround of adding extensions manually.


jwunderl (approved):
LGTM, I don't think this will actually break old editors though cause we append version info to targetconfig request and I think it grabs based of major version? e.g. 
https://makecode.microbit.org/api/config/microbit/targetconfig/v2.0.18

ramseyball:
@abchatra  and @pelikhan  - I noted that this commit removed pxt-jacdac (and other pxt-jacdac extensions). We should discuss - any chance it affects forwardedu extension?

abchatra:
Those are old sections and not read by the client. New sections should have all pxt-jacdac.

PR #5441, opened 2024-01-18
Arcadeshield

tballmsft:


pxtarget.json
```diff
@@ -174,7 +175,7 @@
         "githubCorePackage": "lancaster-university/microbit",
         "gittag": "v2.2.0-rc6",
         "serviceId": "microbit",
-        "dockerImage": "pext/yotta:gcc5"
+        "dockerImage": "tball_yotta"
```
abchatra, line 178: tball_yotta expected here?  Won't this break backend?

tballmsft: this is just for my local work... don't worry...

PR #5443, opened 2024-01-29
Add initial catalog and validator plans for Micro:bit

thsparks:
I expect these will be iterated on. I just wanted to have something here to get started.


eanders-ms (approved):
LGTM ~ when you get a chance can you update the schema in OneNote?

PR #5448, opened 2024-01-31
add "accent smoke" theme color

eanders-ms:
Used by teacher tool. I only added it to header bar color set for now. Will add to the others later.

### DIFF
--- a/theme/themepacks.less
+++ b/theme/themepacks.less
@@ -7,6 +7,7 @@
 @pxt_headerbar_background_glass: rgba(@pxt_headerbar_background, 0.25);
 @pxt_headerbar_foreground: @white;
 @pxt_headerbar_accent: darken(@pxt_headerbar_background, 8%);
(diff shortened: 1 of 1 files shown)


srietkerk (approved):
The change looks fine, but what is an "accent smoke" is this a common term in design? Or are we using the term "smoke" because the color would be the result of smoke going on top of the specified color?

eanders-ms:
> The change looks fine, but what is an "accent smoke" is this a common term in design? Or are we using the term "smoke" because the color would be the result of smoke going on top of the specified color?

The borrowed the term "smoke" from the Xbox Design System, which borrows a lot from Fluent. Smoke is one component that makes up the Acrylic effect in Xbox and Fluent. I chose the term because it describes the mood this color should express, a smokey version of the given root color. To make a color smokey: desaturate it but not all the way to grayscale, then lighten or darken as appropriate to bring it's luminosity into the .65-.75 range.

The complement of smoke is fog. I don't know if we'll need to introduce fog, but it's just brightened smoke.

image

thsparks (approved):

PR #5478, opened 2024-02-08
Set teacher tool toast colors for Micro:Bit

thsparks:
Follow up on https://github.com/microsoft/pxt/pull/9857#issuecomment-1934967487

Had to desaturate and lighten the colors a bit so they weren't too bright. Final result:
image


eanders-ms (approved):
The colors seem a little dark overall, but we can keep tweaking them.

PR #5508, opened 2024-02-22
Add a "Light Foreground" Color to the Themepack

thsparks:
This is useful for description text in the catalog:
image

Confirmed color contrast is okay against a white background using accessibility checker:
image

### DIFF
--- a/theme/themepacks.less
+++ b/theme/themepacks.less
@@ -3,6 +3,8 @@
         source vars: pxt/theme/themepacks.less
 ******************************************/
 
+@pxt_page_foreground_light: #767676;
+
 @pxt_headerbar_background: @blue;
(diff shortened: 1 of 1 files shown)


theme/themepacks.less
```diff
@@ -37,6 +39,7 @@
 :root {
     /// Page
     --pxt-page-font: @pageFont;
+    --pxt-page-foreground-light: @pxt_page_foreground_light;
```
eanders-ms, line 42: Just checking.. will you be opening another PR to add this to the pxt repo?

thsparks: Yes, that's the plan. Just got delayed :)

eanders-ms (approved):

theme/themepacks.less
thsparks: Here it is: https://github.com/microsoft/pxt/pull/9883

PR #5509, opened 2024-02-23
teacher tool carousel content

eanders-ms:
Content for the teacher tool carousels. There's a lot here just to see what the home screen will look like filled out. We will want to whittle it down before we ship.

### DIFF
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -540,6 +540,22 @@
             "youTube": true
         }
     },
+    "teachertool": {
+        "carousels": [
+            {
+                "title": "Rubrics for Tutorials",
+                "cardsUrl": "teachertool/carousels/rubrics-for-tutorials/cards.json"
+            },
+            {
+                "title": "Rubrics for V2 Tutorials",
+                "cardsUrl": "teachertool/carousels/rubrics-for-tutorials-v2/cards.json"
+            },
+            {
+                "title": "Rubrics for Games",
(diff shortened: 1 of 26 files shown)


targetconfig.json
```diff
@@ -540,6 +540,22 @@
             "youTube": true
         }
     },
+    "teachertool": {
+        "carousels": [
+            {
+                "title": "Rubrics for Tutorials",
+                "cardsUrl": "teachertool/carousels/rubrics-for-tutorials/cards.json"
```
srietkerk, line 547: Not a blocking question, I don't have a strong opinion either way, I just wanted to ask why we have each of the cards for the different carousels in separate folders instead of files. As in, keep the `carousels` directory, but in that directory, you just have `games-rubric-cards.json`, `v2-tutorials-rubric-cards.json`, and `tutorial-rubric-cards.json` instead of the folders. I think I just have a hard time imagining something else will go in the separate folders other than `cards.json`.

srietkerk (approved):
One comment, but not blocking. Looks great!

thsparks (approved):

targetconfig.json
eanders-ms: I wanted to model it after how the editor home page works, but it is inconsistent. I chose to go with the per-folder layout   because it seemed more organized. Planning for a time when we have lots of rubric content, I wanted to make sure content editors would be able find the files they're looking for if/when we have lots of rubrics and carousels.

PR #5511, opened 2024-02-27
Teacher tool: blow away rubric

srietkerk:
As the title says, this PR has everything to do with the Blow Away rubric that I made. The approach I took has criteria that are making sure that the program looks almost exactly like what we expect it to look like when someone has completed the Blow Away tutorial if they didn't make any additions or changes. This might not be the route we take for all tutorials, but I think it would be nice to have different rubric design types when we show these to teachers. 

image

Here it is in action:

**Success**
Screenshot 2024-02-26 173958

**Success but the sound level isn't in the right place**
Screenshot 2024-02-26 174404

**Failure since the col + 1 bit is missing**

Screenshot 2024-02-26 174146

**Failure because 'row' wasn't set to a random number**
Screenshot 2024-02-26 174100

As you can see from the success with the sound level in the wrong side of the comparison, one validator that would be valuable to make is one that checks the order of inputs; I need to prioritize other work, and not having a validator that checks order shouldn't be a blocker, but one that will be nice to add.


thsparks:
So, even if we do decide to go with this approach of checking that every block is precisely where we expect it to be, I think our first consideration still needs to be: "Does this map to how a teacher is actually going to grade the assignment?" and I think we'll need to adjust this rubric a fair bit to make that work here.

If we don't do that, it's quite possibly more work for the teacher to read, understand, and act upon the rubric than it is to just grade the assignment by hand. Criteria like "Sound level is detected in the program" and "Variable named 'col' is set to a random value" are great, easy to read, and useful (disregarding my own feelings about requiring specific variable names), but the more specific ones like "forever loop nests a repeat loop that holds all main program logic" and "If an LED point is truthy, then uplot that point and plot a new point" just don't really reflect how a teacher is thinking about grading (I suspect), and there's a significant mental mapping that needs to happen between this and the gradebook. Arguably, this is basically the same as just having them look at the code but with more steps.

With all this in mind, I think we would still benefit from considering the main learning objectives of the tutorial and simplifying the rubric down to those few points, then move the more specific checks surface as just one (or maybe two at most) additional criteria ("Code closely matches expected tutorial code" or something, though we'll need to find a better way to phrase it since we don't account for extra blocks).

tl;dr: I think we really need to simplify this rubric to the learning objectives and then consolidate the more specific checks into only one or two criteria, really abstracting the specific content of the check away so the teacher doesn't have to think about it.

If you want, we can set up a meeting and figure out what that looks like. I don't think this rubric is too far off, we just need to consolidate a few criteria into one.

docs/teachertool/catalog.json
```diff
@@ -12,6 +12,106 @@
...
+            "use": "show_icon_on_start",
+            "template": "Show an icon on the LED screen when the program starts",
+            "description": "An icon block is used inside the on start block",
+            "hideInCatalog": true,
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "0B2BC680-D79D-42C8-BA13-B6B189E2BF7D",
+            "use": "point_condition",
+            "template": "If an LED point is truthy, then uplot that point and plot a new point",
```
thsparks, line 111: I think we would benefit from using more casual, user-oriented language. Some of these are a bit too javascripty, I feel.

```suggestion
            "template": "If an LED point is lit, then unplot that point and plot a new point",
```

thsparks: ditto for the description here

docs/teachertool/catalog.json
```diff
@@ -12,6 +12,106 @@
...
+            "use": "soundlevel_greater_than_check",
+            "template": "Check that the sound level is greater than a number",
+            "hideInCatalog": true,
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "B835FAA0-8CA6-4E4A-95B4-7FC6D15231BD",
+            "use": "soundlevel_gt_condition",
+            "template": "If sound level is greater than a number, then two variables are set to random values",
+            "description": "In the if chunk of code that checks that sound level is greater than a number, variables 'row' and 'col' are set to random values",
```
thsparks, line 34: Let's chat with the team about wording choices here. I think "the if chunk of code" may not be the best way to refer to this. Perhaps "Inside the if block" or something would be clearer. I'm sure we've had to refer to this sort of thing in tutorials before.

thsparks: Perhaps @kiki-lee could provide some good insights on what language to use?

thsparks (changes requested):
Left a few comments. I think we should meet up and figure out how we can trim this down and simplify the wording.

docs/teachertool/validator-plans.json
```diff
@@ -55,6 +55,213 @@
...
+                    "validator": "blockFieldValueExists",
+                    "fieldType": "OP",
+                    "fieldValue": "GT",
+                    "blockType": "logic_compare",
+                    "childValidatorPlans": ["get_sound_level", "math_num_exists"]
+                }
+            ]
+        },
+        {
+            ".desc": "variable with name col is set to random number",
```
eanders-ms, line 87: > with name col

This level of specificity will be problematic for localized tutorials.

docs/teachertool/validator-plans.json
```diff
@@ -55,6 +55,213 @@
...
+                    "validator": "blocksExist",
+                    "blockCounts": {
+                        "controls_if": 1
+                    },
+                    "childValidatorPlans": ["soundlevel_greater_than_check", "row_variable_set_random", "col_variable_set_random"]
+                }
+            ]
+        },
+        {
+            ".desc": "point block used for boolean check //no catalog entry",
```
eanders-ms, line 155: What is meant by `//no catalog entry`?

docs/teachertool/validator-plans.json
```diff
@@ -55,6 +55,213 @@
...
+            ".desc": "variable with name col is accessed //no catalog entry",
+            "name": "col_add_num",
+            "threshold": 1,
+            "checks": [
+                {
+                    "validator": "blockFieldValueExists",
+                    "fieldType": "OP",
+                    "fieldValue": "ADD",
+                    "blockType": "math_arithmetic",
+                    "childValidatorPlans": ["col_variable_accessed", "math_num_exists"]
```
eanders-ms, line 192: Where is the definition of `math_num_exists`? Maybe I'm just missing it...

docs/teachertool/validator-plans.json
srietkerk: It means that it has a validator plan, but does not have a corresponding JSON entry in the catalog. I thought it might be useful to include these types of comments, but I think we landed on that they're not necessary and that I can remove them.

docs/teachertool/validator-plans.json
srietkerk: It's on the pxt side, in `validator-plans-shared.json`. https://github.com/microsoft/pxt/blob/master/common-docs/teachertool/validator-plans-shared.json#L211-L220

docs/teachertool/validator-plans.json
srietkerk: Hmm, I didn't consider that. I think that there are plenty of people that are skeptical of the value of checking variable names, anyway, but I would still love to get feedback from teachers of whether or now they would find a check like this valuable.

thsparks: I think a compromise might be when we have parameterized criteria, we can have the teacher type in the name (or names) they want to check for. Until then, I'd be inclined to remove this criteria.

docs/teachertool/catalog.json
eanders-ms: "In the section of code that checks if the sound level is above a certain number, it randomly assigns values to 'row' and 'col'."

docs/teachertool/catalog.json
eanders-ms: "If an LED is on, turn it off and light up a different LED."

docs/teachertool/catalog.json
```diff
@@ -12,6 +12,112 @@
...
+            "use": "soundlevel_greater_than_check",
+            "template": "Check that the sound level is greater than some number",
+            "hideInCatalog": true,
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "B835FAA0-8CA6-4E4A-95B4-7FC6D15231BD",
+            "use": "soundlevel_gt_condition",
+            "template": "If a detected sound level is greater than some number, then two variables are set to random values",
+            "description": "An if statement checks a detected sound level is greater than a predetermined number. If this is true, variables 'row' and 'col' are set to random values",
```
thsparks, line 34: What does a description this long look like in the UI? Would love a screenshot. We may need to make some adjustments, either to how we display those or to have a max length here.

docs/teachertool/catalog.json
```diff
@@ -12,6 +12,112 @@
...
+            "use": "row_variable_set_random",
+            "template": "Variable named 'row' is set to a random value",
+            "hideInCatalog": true,
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "01A65046-D4BE-44B6-8273-7A878A71B3D0",
+            "use": "point_bool_check",
+            "template": "Check that an LED is lit on the screen",
+            "description": "At the spot 'col', 'row' on the LED screen, the LED is on",
```
thsparks, line 56: nit:
```suggestion
            "description": "At the spot ('col', 'row') on the LED screen, the LED is on",
```

thsparks: If you go with this formatting (which I think is more consistent with how points are written in school), we also need to adopt it anywhere else we reference "spot col, row"

docs/teachertool/catalog.json
```diff
@@ -12,6 +12,112 @@
...
+            "use": "col_add_num",
+            "template": "Sum the value of the 'col' variable and a given number",
+            "hideInCatalog": true,
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "6BA59761-9464-4C8D-9D8F-E8985273FB92",
+            "use": "plot_vars_used",
+            "template": "Plot a point on the LED screen at position 'col' + 1, 'row'",
+            "description": "Uses the block that turns on an LED",
```
thsparks, line 79: Is this description necessary?

docs/teachertool/catalog.json
```diff
@@ -12,6 +12,112 @@
...
+            "hideInCatalog": true,
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "1B331451-081E-4F26-9CB6-F5A649432BF9",
+            "use": "repeat_loop_used",
+            "template": "Repeat loop is used",
+            "description": "A repeat loop is found in the program",
```
thsparks, line 87: I think this description is redundant.

thsparks: ditto for forever_used below

docs/teachertool/catalog.json
thsparks: I think we still need to update this one. I like Eric's suggestion.

docs/teachertool/catalog.json
```diff
@@ -12,6 +12,112 @@
...
+            "template": "Show an icon on the LED screen when the program starts",
+            "description": "An icon block is used inside the on start block",
+            "hideInCatalog": true,
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "0B2BC680-D79D-42C8-BA13-B6B189E2BF7D",
+            "use": "point_condition",
+            "template": "If an LED point is truthy, then uplot that point and plot a new point",
+            "description": "An if statment checks that a specific LED is on/1/true. If this is true, turn that light off and turn on a new light",
```
thsparks, line 111:
```suggestion
            "description": "An if statment checks that a specific LED is lit, turns that light off, and then turns on a new light",
```

docs/teachertool/catalog.json
```diff
@@ -12,6 +12,112 @@
...
+            "hideInCatalog": true,
+            "docPath": "/teachertool"
+        },
+        {
+            "id": "ECC79C65-56DC-44A9-98FC-147F5EED87CC",
+            "use": "blow_away_completeness",
+            "template": "Project completeness",
+            "description": "The project contains all the required blocks by the end of the tutorial",
```
thsparks, line 119: nit:
```suggestion
            "description": "The project contains all the blocks required at the end of the tutorial",
```

docs/teachertool/catalog.json
srietkerk: I'm totally fine with changing it, this is one spot in particular that I chatted with Kiki about, though. She agreed that describing the LED as lit would be a sound way forward, but also proposed that we could use the opportunity to teach through context here. By saying "we're checking that a specific LED is on/1/true", we're able to explain what it means to be truthy. 

Again, totally fine with changing it, but I wanted to present it in that light before doing so.

thsparks: Personally, I don't think this is the place to try and teach teachers this sort of thing. We could put something like that in the curriculum itself, but the goal of this tool is to simplify their process and reduce cognitive load. This adds to it (quite significantly).

docs/teachertool/catalog.json
srietkerk: **Big screen, about 50% split**
image

**Big screen, smaller rubric panel**
image

**Smaller screen**
image

On the smaller screen, it definitely looks a bit funky, but I didn't set it to a specific size with debugging, just readjusted my window, so it might not be an accurate screen size.

docs/teachertool/catalog.json
eanders-ms: spelling: statment -> statement

docs/teachertool/catalog.json
srietkerk: The top is a bit technical, so I think the description is nice in that way, but I can combine the two a bit. Like: "Turn on an LED at point ('col' + 1, 'row')"

thsparks: I like that!

docs/teachertool/catalog.json
thsparks: What about inside the catalog popup? (Not that we need to worry too much about it right now.) This probably isn't a blocker, but it'd be good to know if we'll need to account for this going forward.

docs/teachertool/catalog.json
eanders-ms: Referring to an LED as a "light" seems inconsistent here, It would also conflict with the usage of that word in my suggestion above, should you choose to accept it. Maybe:

```suggestion
            "description": "Checks to see if a certain LED is on, and if so, turns it off and turns a different LED on.",
```

srietkerk: I'm thinking that we should just get rid of the description for this one. They sound too similar for the description to be valuable.

thsparks (approved):
👍

PR #5517, opened 2024-03-01
Micro Chat Rubric

thsparks:
Happy to iterate, but I thought just sending the PR would be a good way to capture feedback.

Rubric is here:
image

I would've liked to add two "bonus" checks as well (see below), but we don't currently have the ability to check for nested blocks when more than 1 block is required, which means we can't support the first check (yet), and it felt weird to have one and not the other.

**Bonus: Sends additional strings on different button events**
The program has more than one "on button pressed" event which sends a string.

**Bonus: Can send secret messages by changing the radio group**
The program contains logic to change the radio group beyond the initial set on startup.

Depends on https://github.com/microsoft/pxt/pull/9896 for parameter checks

### DIFF
--- a/docs/teachertool/catalog.json
+++ b/docs/teachertool/catalog.json
@@ -17,7 +17,7 @@
             "id": "18C44CC4-497F-45EC-90FA-CE7DB117AD03",
             "use": "set_radio_group_on_start",
             "template": "Sets the radio group on startup",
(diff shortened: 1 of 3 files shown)


eanders-ms (approved):
Note to self/others: After my PR is merged, this rubric will need to be re-added to cards.json. I made all empty rubrics hidden.

Issue #5434, opened 2023-12-29
Unable to get tooltip for some LEDs


ofaurax:
**Describe the bug**
On hover-ing LEDs in the simulator, one can get their coordinates.
Like this:
![Capture d’écran du 2023-12-29 11-01-40](https://github.com/microsoft/pxt-microbit/assets/271723/10abc2be-a302-41ca-96c4-2cd3f2953613)

But for certain LEDs, this is not possible, as it seems there is a sim-head forward:
![Capture d’écran du 2023-12-29 11-02-02](https://github.com/microsoft/pxt-microbit/assets/271723/2c16ecd7-c456-40d6-bca6-98d7d7480207)

**To Reproduce**
1. Create a project from a computer
2. Put your mouse on the first LED, you'll see the tooltip (0,0)
3. Put your mouse on the second LED (same line), you won't see the tooltip

**Expected behavior**
See the (1,0) tooltip

**micro:bit version (please complete the following information):**
V2.21, but it's not hardware related

**Desktop (please complete the following information):**
 - OS: Linux Mageia 9
 - Browser: Firefox
 - Version 115.5.0esr

**Smartphone (please complete the following information):**
None

**Additional context**
I need to mention microbit-support

ofaurax:
I forgot to mention microbit-support

abchatra:
Here is the sample program where one can repro: https://makecode.microbit.org/_h7xiwA6Pqe9t
Some of the top row leds don't have tooltips when hovered.

PR #5578, opened 2024-04-02
Changed bounding box of microbit logo

srietkerk:
The svg box for the micro:bit logo was too large. It was interfering with our coordinate tooltips on the LEDs. To fix this, I changed the sim-head svg container to be an ellipse to better fit the shape of the logo and fit the shape closer to the logo. Because of this, some of the interactions with the compass were a bit wonky, so that's where the `moveHeadingOnClick` function comes in. Because we're adding an event listener to the body of the board, we also need to make sure that we remove that event listener when the board is killed, but the BoardView is only accessed within the context of the BoardHost, so I needed to make some modifications to the BoardHost as well (PR below).

Required for this PR: https://github.com/microsoft/pxt/pull/9946
Fixes https://github.com/microsoft/pxt-microbit/issues/5434

Upload target: https://makecode.microbit.org/app/cbd4697446a349f835837d3cb85d0943d5e65bc2-2310ea9f83#editor 
Two things to try: 
1. hover over the first row of LEDs to see that the tooltips are now showing up
2. Play with the compass to make sure that it's interactions are as expected

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,6 @@
     },
     "dependencies": {
         "pxt-common-packages": "12.0.1",
-        "pxt-core": "10.0.15"
+        "pxt-core": "10.0.16"
     }
 }
--- a/package.json
+++ b/package.json
@@ -46,6 +46,6 @@
     },
     "dependencies": {
         "pxt-common-packages": "12.0.1",
-        "pxt-core": "10.0.11"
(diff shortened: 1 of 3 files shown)


eanders-ms (approved):
Nice improvement.

PR #5592, opened 2024-04-11
Teacher Tool: Clean Micro:Bit Catalog

thsparks:
Now that parameterized criteria are supported in beta, we can reduce some of the duplication in our catalog and sample rubrics.

Kept:
- Show an icon on the display
- Read a GPIO pin

Hid from main catalog but kept for sample rubrics:
- Runs code when the micro:bit is shaken
- Show an icon on display when a condition is met

Removed (can use `${Block} used ${count} times` for all of these):
- Utilizes the pick random block
- Utilizes the if/else block
- Forever loop is used
- Repeat loop is used
- Sound level is detected in the program

Related pxt change: https://github.com/microsoft/pxt/pull/9965

While testing, I noticed a bug where parameters seem to be getting overwritten when we have multiple instances of the same catalog criteria, but I'll look into that separately.

### DIFF
--- a/docs/teachertool/validator-plans.json
+++ b/docs/teachertool/validator-plans.json
@@ -268,38 +268,6 @@
             ]
         },
         {
-            ".desc": "'pick random' block is present",
-            "name": "pick_random_exists",
-            "threshold": 1,
-            "checks": [
-                {
-                    "validator": "blocksExist",
-                    "blockCounts": [
-                        {
-                            "blockId": "device_random",
-                            "count": 1
-                        }
-                    ]
-                }
-            ]
-        },
-        {
-            ".desc": "'if/else' block is present",
-            "name": "conditional_exists",
-            "threshold": 1,
-            "checks": [
-                {
-                    "validator": "blocksExist",
-                    "blockCounts": [
-                        {
-                            "blockId": "controls_if",
-                            "count": 1
-                        }
-                    ]
-                }
-            ]
-        },
-        {
             ".desc": "read value of 'hand' variable",
             "name": "hand_variable_accessed",
             "threshold": 1,
--- a/docs/teachertool/validator-plans.json
+++ b/docs/teachertool/validator-plans.json
@@ -17,6 +17,29 @@
             ]
         },
         {
+            ".desc": "Shows something on the LED screen.",
+            "name": "output_to_led_display",
+            "threshold": 1,
+            "checks": [
+                {
+                    "validator": "blocksInSetExist",
+                    "blocks": [
+                        "device_show_number",
+                        "device_show_leds",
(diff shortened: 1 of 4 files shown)


docs/teachertool/validator-plans.json
```diff
@@ -267,38 +267,6 @@
                 }
```
srietkerk, line 267: Great changes. In addition to the other modifications, but I have some thoughts on the criteria that were elft.

Here are my thoughts:
1. We remove custom target plans/catalogs (except for tutorials or for specific curriculum). With the parameterized criteria, users can specify any block that they need. I'm not sure what the value of having these target-specific criteria.
2. We have a target-specific criteria that looks like: `show ${block} on display` where, when the user clicks on the "block" area, the pop-up for the blocks just has the possible show blocks. This would be more work, though. We could have something similar for the "GPIO pin pressed" block where it can be something like `interaction block ${block} used`, and the blocks are just the `on ___ pressed` blocks.

thsparks: I think there's still value in custom target catalog entries. Things like "GPIO pin pressed" can be useful, since that actually encompasses several possible blocks that'd be a pain to pick out by hand. (There could be similar scenarios for "Code has inputs" and "Code has outputs" type validation).

Not opposed to having intelligent filtering in the block picker, but it is already broken into the toolbox categories which I suspect could be sufficient. We can see if this is an area where we get feedback.

srietkerk: Yeah, the extra filtering would be great to get feedback on. 

I think the code has inputs, code has outputs criteria will be important ones, but I don't think it's part of this change. I do think the `show_icon` criteria should be changed, though. Either to using parameters or just getting rid of it to be absorbed into the `block ${block} used ${count} times` criteria. I could be wrong about this, but I feel like having that specific criteria, especially worded  differently with "show an icon on the display", makes it seem like this criteria will behave differently, like it may do a check at run time or something.

thsparks: I'm not sure parameterizing that one is going to be all that helpful. However, I do think we should update it to reference more than just the show icon block. Perhaps "Outputs to LED screen display" and we check for show icon, show string, the LED grid block, show arrow, etc... Would that address your concern?

srietkerk: Yeah, that would be great! 

And I think you're right about it not being helpful. I think I'm just stuck in the "a teacher will want to make sure that the student has the right specific block" mentality, even though we learned that teachers wouldn't want to make rules like that. And for teachers that _do_ want that specificity, they can just use the parameterized "block used" criteria.

srietkerk (approved):
Looks great!

PR #5621, opened 2024-04-25
Teacher Tool: Add Max Count for Micro:bit Criteria

thsparks:
Field added in https://github.com/microsoft/pxt/pull/9986, need to include for micro:bit criteria as well.


srietkerk (approved):
Nice

PR #5656, opened 2024-05-07
Add tags to visible criteria

thsparks:
Tag support added in https://github.com/microsoft/pxt/pull/10007, need to add for microbit criteria as well.


srietkerk (approved):
I'm not sure where this would live, but it might be nice to have documentation that outlines standards that we should use when applying a tag to a criteria entry. Looks good to me, though!

Issue #5558, opened 2024-03-21
Simple Teacher Request: Allow pin parameters to be variables instead of (as well as?) drop-down.


kiki-lee:
Would love the pin blocks to be parameterized.  Would love to be able to have the Pin be a variable…instead of a drop-down.  
![image](https://github.com/microsoft/pxt-microbit/assets/1633958/c00885a3-6099-42d3-b5e0-0f23bd883c06)

nevergone:
I would be very happy to have a general solution that can also be used for extensions. Maybe wherever there is a dropdown.

nevergone:
Maybe [/microsoft/pxt-microbit/milestone/7 ?](https://github.com/microsoft/pxt-microbit/milestone/7) ?

abchatra:
This is fixed in /beta @jaustin  fyi

PR #5700, opened 2024-05-29
Make pin parameters variable

sae220:
https://github.com/microsoft/pxt-microbit/issues/5558
https://github.com/microsoft/pxt/issues/10015
variablePinParam


nevergone:
Just checking for equality doesn't seem to work.

```
let test1 = DigitalPin.P0
let test2 = DigitalPin.P1
if (test1 == test2) {
	
}
```
Screenshot_20240529-210920

sae220:
@abchatra 
Could you review this?

carlosperate:
Is there any user facing impact to the blocks and/or the javascript API? If so, can we get a before and after screenshot?

sae220:
@abchatra 
I fixed test blocks.

sae220:
@carlosperate 
variablePinParam

nevergone:
Do you know any good methods for testing equality that children can use?
https://github.com/microsoft/pxt/issues/10015#issuecomment-2138091719

microbit-carlos:
Thanks for the screenshot @sae220! What about the Javascript pin API, how does the code change?

sae220:
@nevergone
> Do you know any good methods for testing equality that children can use?

Sorry, but I can't understand well. You mean you want to check for equality of different pins?

@microbit-carlos 
> What about the Javascript pin API, how does the code change?

As far as I know, there are no code changes in JavaScript.

nevergone:
@sae220 

I have two variables with pin.
First I want to check that they contain the same pin. And then that the variable contains Pin2. Just as the contents of other variables can also be compared.
You have done a lot of work, for which I thank you very much.
If this can be solved, I think it will be perfect.

sae220:
@nevergone 

 So you mean you want to compare obviously inequal pins (like `DigitalPin.P0` and `DigitalPin.P1`)?
[As I shew before](https://github.com/microsoft/pxt/issues/10015#issuecomment-2138510694), we can compare pins when we use array (like `DigitalPin` and `DigitalPin.P1`).
Isn't it enough for your purpose?

Sorry for my poor English.

nevergone:
@sae220 

Some pins are stored in a variable. How do we know afterwards which pin is in the variable? How can we check if a specific pin is present in the variable?

sae220:
@nevergone 

I'm very sorry, but I can't understand the situation well.
Could you give me some example of code (blocks or JavaScript)?

nevergone:
@sae220 
I tried it and now it works perfectly. Thanks for all your work! :)

Screenshot_20240607-233604

```
    if (var1 < var2) {
    }
    if (var1 < DigitalPin.P10) {
        basic.showIcon(IconNames.SmallHeart)
    }
})
input.onButtonPressed(Button.AB, function () {
    if (var1 == var2) {
        basic.showIcon(IconNames.Duck)
    }
    if (var1 == DigitalPin.P10) {
        basic.showIcon(IconNames.Tortoise)
    }
})
    if (var1 > var2) {
        basic.showIcon(IconNames.Yes)
    }
    if (var1 > DigitalPin.P10) {
        basic.showIcon(IconNames.Happy)
    }
})
let var2 = 0
let var1 = 0
var1 = DigitalPin.P0
var2 = DigitalPin.P1
```

sae220:
@abchatra 
I fixed some test code. Could you review this again?

riknoll (approved):
LGTM!

This needs upgrade rules to ensure backwards compat with old projects
Also needs new docs pages for the two new blocks

...but we can just file an issue for that and do it later

abchatra:
@sae220 thanks for contribution. There are few follow ups: upgrade rules for existing projects and doc pages update. If you can file issues for the same it would be great. If you want to tackle the upgrade rules feel free to do so as well.

sae220:
@abchatra @riknoll 
OK. I'll upgrade soon, but I don't know how I can upgrade rules for existing projects to ensure backwards compat.
Could you let me know the helpful commit?

riknoll:
@sae220 it's all in this file: https://github.com/microsoft/pxt-microbit/blob/master/editor/patch.ts

basically just need to add some code that converts the old xml to the new one, just like you did for the test code

abchatra:
Also @sae220  you should be able to see the new block in /beta now. Greatly appreciate the work.

sae220:
@jaustin 
You've got a point. This pull request may be thoughtless.

But it isn't caused by this pr that the getter method `digital read pin` includes `read`. I don't have any idea to solve this.

To solve other problem, what do you think about the unification of `digital pin` and `analog pin`?
Using blocks error list (but experimental), we can restrict usable pins.

sae220:
@abchatra 
What do you think about this?

microbit-carlos:
@abchatra looks like the change in the blocks is causes issue loading older projects, like in https://github.com/microsoft/pxt-microbit/issues/5728.

Even if the generated code is identical the change in the blocks themselves seems to be causing this issue. 
Has MakeCode already implemented a way to "convert old code" for previous block updates? When other blocks, like the music blocks, have been refactored, how did we managed importing older programmes into newer versions of MakeCode?

abchatra:
We can fix the update issues.

sae220:
@microbit-carlos @abchatra 
I created pull request to convert old code.
https://github.com/microsoft/pxt-microbit/pull/5712

microbit-carlos:
I have a few suggestions, which might be worth considering as I think it'd likely affect https://github.com/microsoft/pxt-microbit/pull/5712 as well.

To reduce the block visual difference, and word redundancy (`pin [P0]` vs `pin [digital pin P0]`):

| MakeCode v6 | This PR |
|--------|--------|
| image | image | 

We can probably create a separate block with just the pin name, that is only used as a shadow block for other pin blocks.
These blocks wouldn't be available on their own, and they could not be "dragged out" from the other pin blocks (the currently implementation doesn't let you drag-out the shadow block, so that's already done).

So my suggestion to this PR would be something like this:

|This PR | Suggestion |
|--------|--------|
image | Screenshot 2024-07-17 at 16 25 10 |

And so the change between MakeCode v6 and v7 would be very small:

| MakeCode v6  | Suggestion |
|--------|--------|
| image | Screenshot 2024-07-17 at 16 25 10 |

@abchatra having this smaller digital/analog pin block to serve only as a shadow block something you'd consider accepting?

-----

We can still keep the new pin blocks, as these are new and they also need to be self explanatory on their own:

image

-----

Another change that would be needed is to revert back the pulse event block, or to change it in a way that doesn't let the user add a variable for of the pin:

image

This is because unfortunately the code generation can end up placing the event handler code on top of the "on start" code, and so we might end up with an undefined variable:

image image

-----

As a separate conversation we might need to consider if the "analog pin" block can be improved in the future as well, as something like this code snippet should not work, but MakeCode doesn't show any errors and it even works in the simulator:

image


(To clarify, this was not introduced in this PR, it has been like this in MakeCode for a very long time I assume).

sae220:
@microbit-carlos 
About the first suggestion, I'll create new pull request to shorten new blocks after discussion in #5733.

To solve second point, how about fix compiling code to bring `on-start` code top?

PR #5702, opened 2024-06-02
Fix reaction-time code

martinwork:
Arising from support ticket https://support.microbit.org/helpdesk/tickets/75716 (private)

If restarted without a result, or after a false start, a press during the countdown may be processed as a success, because **running** is true from the previous start.

This PR moves initialisation of the control variables before the countdown.
image


abchatra:
@ganicke can you please review this code?

ganicke (approved):
P0 preset seems to work OK.

PR #5713, opened 2024-06-27
Make pins.map block inline

sae220:
I accidentially closed previous pull request (https://github.com/microsoft/pxt-microbit/pull/5697).

@abchatra 
Could you review this simple pull request? Or could you let me know why it should not be inline?


abchatra:
@sae220 can you attach the screenshot of change in blocks?

sae220:
inlinePinMap_2
inlinePinMap_1
I don't know why pins_map is needed anyway.

abchatra:
@sae220 we won't be taking this change as we don't want to change the look of this block due to legacy reasons. (Too many teachers have screenshots).

PR #5725, opened 2024-07-12
add font extension

tballmsft:
Adding a font extension for future use by pxt-arcadeshield. This extension needs to be included in pxt-microbit because the font is too large to be included in a github hosted extension.


tballmsft:
@riknoll - poke

riknoll (approved):

PR #5752, opened 2024-07-29
Bring event block after "on-start"

sae220:
#5700 
> This is because unfortunately the code generation can end up placing the event handler code on top of the "on start" code, and so we might end up with an undefined variable:

block
block

before
before

after
after

@microbit-carlos @abchatra

### DIFF
--- a/libs/core/control.cpp
+++ b/libs/core/control.cpp
@@ -299,7 +299,7 @@ namespace control {
      * Registers an event handler.
      */
     //% weight=20 blockGap=8 blockId="control_on_event" block="on event|from %src=control_event_source_id|with value %value=control_event_value_id"
-    //% help=control/on-event
+    //% help=control/on-event afterOnStart=true
     //% blockExternalInputs=1
     void onEvent(int src, int value, Action handler, int flags = 0) {
         if (!flags) flags = ::EventFlags::QueueIfBusy;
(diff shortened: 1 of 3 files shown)


sae220:
@abchatra 
Could you review this?

abchatra:
Sorry for the delay. I will review this week

riknoll:
we hoist event handlers above on-start intentionally because people can write code that pauses inside of on-start.

for example, take a look at this program:

microbit-screenshot (31)

if we were to register the event handler after on-start, it ends up breaking this program because that while loop will never finish executing.

i agree that both error scenarios are confusing on this one, but i think not hoisting the event handler is worse. i think the better fix would be to add a warning when variables are used in event handlers that don't have afterOnStart, similar to how we [give a warning](https://github.com/microsoft/pxt/blob/390bcc6d61ef0417a5c51c17eb1acafeefbe1e63/pxtblocks/compiler/compiler.ts#L180) if a variable is never assigned in blocks

microbit-carlos:
Alternatively we could leave the pin event handler with a "normal pin dropdown" (what we had in MakeCode v6), instead of the shadow block. That way users cannot drop a variable into the handler that could be undefined.

abchatra:
Let's close this PR based on Richard's comments. Thanks for @sae220  for all the work.

PR #5754, opened 2024-07-29
CI: Update testghpkgs workflow to be able to run it again.

microbit-carlos:
It's been a while since this workflow was run, so it needed some minor updates.

### DIFF
--- a/.github/workflows/testghpkgs.yml
+++ b/.github/workflows/testghpkgs.yml
@@ -2,6 +2,11 @@ name: pxt-testghpkgs
 
 on:
   workflow_dispatch:
+    inputs:
+        branch-or-commit:
+          description: Branch or commit to checkout for test
+          required: false
+          default: 'master'
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/testghpkgs.yml
```diff
@@ -10,15 +15,14 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x]
-        branch: [stable3.0]
```
microbit-carlos, line 14: Rather than have the branch in the yaml file, I've added it to as a workflow form input, so that it can be set to whatever it's needed at the point the workflow is manually triggered:
image

.github/workflows/testghpkgs.yml
```diff
@@ -2,6 +2,11 @@ name: pxt-testghpkgs
 
 on:
   workflow_dispatch:
+    inputs:
+        branch-or-commit:
+          description: Branch or commit to checkout for test
+          required: false
+          default: 'master'
```
microbit-carlos, line 9: I've set the default to the main branch, the workflow has testing on `stable3`, so if needed we could change this to one of the stable branches, but since the form has to be triggered manually I think it's probably fine to input that at the time it's triggered.

microbit-carlos:
I'm manually triggered this workflow and it's currently running agains the `srietkerk/codal-bump-0.2.67` branch:
https://github.com/microsoft/pxt-microbit/actions/runs/10149336339/job/28064053783

microbit-carlos:
@abchatra this is running correctly, so it's ready to be reviewed.

abchatra:
@riknoll to review

abchatra:
@carlosperate  is this still relevant?

Issue #5286, opened 2023-06-26
Recording extension: Set Sample Rate block not implemented in simulator


microbit-carlos:
**Describe the bug**
I believe the sample rate blocks have not implemented in the simulator in the 2023 release yet.

~~If that's the case there should probably be some kind of UX to let the user know this? Otherwise it looks like the code is not working as expected, which can be quite frustrating if you think there is a bug in your code.~~

2024 edit: Can we implement this feature in the simulator for the 2024 release?

**To Reproduce**
1. Add recording extension
2. Add the code below
4. Try it on the sim
5. The playback is not at twice the speed

<img width="530" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/29712657/ae8c412a-3a7c-4f11-849c-53d05c93fd37">

```ts
input.onButtonPressed(Button.A, function () {
    record.startRecording(record.BlockingState.Blocking)
    record.playAudio(record.BlockingState.Blocking)
})
record.setSampleRate(22000, record.AudioSampleRateScope.Playback)

```

**Expected behavior**
The sample rate features to work on the simulator.

**Screenshots**
Already added.

**micro:bit version (please complete the following information):**

V2.x

**Desktop (please complete the following information):**
- OS: macOS
- Browser: Chrome 114

**Smartphone (please complete the following information):**
N/A

**Additional context**
makecode.microbit.org version:  [6.0.10](https://github.com/Microsoft/pxt-microbit/releases/tag/v6.0.10)
Microsoft MakeCode version:  [9.0.7](https://github.com/Microsoft/pxt/releases/tag/v9.0.7)
microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
codal-microbit-v2 runtime version:  [v0.2.55](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.55)

Edit:
Tried it again in:
makecode.microbit.org version:  [6.0.25](https://github.com/Microsoft/pxt-microbit/releases/tag/v6.0.25)
Microsoft MakeCode version:  [9.0.15](https://github.com/Microsoft/pxt/releases/tag/v9.0.15)
microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
[...]

PR #5757, opened 2024-07-29
Add simulator support for sample rate

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5286

This adds simulator support for the sample rate blocks in the audio recording extension. 

The new behavior includes:

* Worse audio quality at lower sample rates (and better at higher)
* Longer recording times at lower sample rates (and shorter at higher)
* The ability to mess with playback speed by changing the output sample rate

The recording times vary from 3 seconds at the highest sample rate to 20 seconds at the lowest. I can adjust this range if need be, but those were the values @srietkerk got when testing on hardware.

Also, the browser has limits on how fast/slow you can play audio, so I capped it within a reasonable range; I've tested recording and playback in Chrome + Firefox on Windows and Safari on macOS.

Here's a test URL:

https://makecode.microbit.org/app/8ac1af74a11ed81b3d41f1a0d0eee068d2158fdf-3acc60290d

### DIFF
--- a/sim/state/record-audio.ts
+++ b/sim/state/record-audio.ts
@@ -4,11 +4,18 @@ namespace pxsim  {
         stream: MediaStream;
         recorder: MediaRecorder;
         chunks: Blob[];
+
         audioURL: string;
+        // The inputBitRate when the current audioUrl was recorded
+        audioURLBitRate: number;
+
         recording: HTMLAudioElement;
         audioPlaying: boolean = false;
         recordTimeoutID: any;
         currentlyErasing: boolean;
+
+        inputBitRate = record.defaultBitRate();
+        outputBitRate = record.defaultBitRate();
 
         handleAudioPlaying = () => {
             this.audioPlaying = true;
@@ -27,6 +34,16 @@ namespace pxsim  {
     }
 }
 namespace pxsim.record {
+    // Arbitrarily chosen lower bound. Can't go much lower than this without bugs cropping up
+    const MIN_BIT_RATE = 3000;
+    // This is double the default in chrome (128000)
+    const MAX_BIT_RATE = 256000;
+
+    const MAX_SAMPLE_RATE = 22000;
+    const MIN_SAMPLE_RATE = 1000;
+
+    const MIN_RECORDING_TIME = 3000;
+    const MAX_RECORDING_TIME = 20000;
 
     let _initialized = false;
     function init() {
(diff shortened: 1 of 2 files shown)


sim/state/record-audio.ts
```diff
@@ -139,20 +165,42 @@ namespace pxsim.record {
...
-            if (!b.recordingState.currentlyErasing && b.recordingState.recording) {
+            if (!state.currentlyErasing && state.recording) {
                 try {
                     const volume = AudioContextManager.isMuted() ? 0 : 1;
-                    b.recordingState.recording.volume = volume;
-                    await b.recordingState.recording.play();
-                } catch (e) {
+                    state.recording.volume = volume;
+
+                    const minPlaybackRate = 0.15
```
srietkerk, line 178: Is there a reason that these aren't defined with the other default variables at the top of the file?

sim/state/record-audio.ts
```diff
@@ -139,20 +165,42 @@ namespace pxsim.record {
...
+
+                    let maxPlaybackRate = 15;
+                    if (isFirefox()) {
```
srietkerk, line 181: I know that you said that you were experiencing weirdness with Firefox. Is there a reason why setting this to 8 works for Firefox or was it with testing that this was found to be the value that works on Firefox?

If there is a tangible reason, could we maybe add a comment here for a quick explanation?

sim/state/record-audio.ts
```diff
@@ -139,20 +165,42 @@ namespace pxsim.record {
...
+                        maxPlaybackRate = 8;
+                    }
+
+                    const playbackRate = Math.max(minPlaybackRate,
+                        Math.min(
+                            maxPlaybackRate,
+                            bitRateToSampleRate(state.outputBitRate) / bitRateToSampleRate(state.audioURLBitRate)
```
srietkerk, line 188: This question might be best asked as what is the relationship between bit rate and sample rate, but I'm going to put here what I'm thinking through.

I know you say below that the browser APIs don't allow us to adjust the sample rate directly and can map in between sample rate and bit rate. If that's the case, why do we need the playback to be the sample rate? Can we just have playback rate be the bit rate?

sim/state/record-audio.ts
```diff
@@ -221,14 +269,44 @@ namespace pxsim.record {
...
+        return mapRange(bitRate, MIN_BIT_RATE, MAX_BIT_RATE, MIN_SAMPLE_RATE, MAX_SAMPLE_RATE);
+    }
+
+    function mapRange(value: number, inMin: number, inMax: number, outMin: number, outMax: number) {
+        value = Math.min(Math.max(inMin, value), inMax);
+
+        return ((value - inMin) / (inMax - inMin)) * (outMax - outMin) + outMin;
+    }
 
+    export function defaultBitRate() {
```
srietkerk, line 309: Does this function need to get exported?

riknoll: yes, it's used in the state above which is in a different namespace

sim/state/record-audio.ts
riknoll: bit rate is the sample rate times the bit depth. I couldn't immediately find what the bit depth of the browser's media recorder is, but in any case the sample rate should have a linear relationship with the bit rate. personally, i don't think it's super important that we accurately match the sample rate of the hardware since your computer has a different mic and is going to sound better no matter what; all that matters is that lower sample rate sounds worse and higher sounds better.

the reason i convert back and forth is because we still want the playback rate to be based on the sample rate that the user set. for example, if they set the recording rate to be 11000 and the playback rate to be 22000, the playback rate should be 2.0, but because the bit rate scales differently it would end up being much higher if we used the bit rate instead.

sim/state/record-audio.ts
riknoll: they're only used in this one spot

sim/state/record-audio.ts
riknoll: i'll add a comment, but yes i found it by experimenting. 15 was the max playback rate in chrome and 8 was the max in firefox. i didn't try to figure out the max in safari, but 15 worked and that's higher than anyone will be able to detect anyhow.

sim/state/record-audio.ts
srietkerk: Ohh, that makes sense. Thanks!

srietkerk (approved):
LGTM!

jwunderl (approved):

PR #5767, opened 2024-07-30
Regenerate dal.d.ts

riknoll:
Regenerating the dal.d.ts file for libs/core.

I really just want to get these constants: https://github.com/lancaster-university/codal-microbit-v2/commit/9a03c691c12474852819843f21e32908ab405a2f


thsparks (approved):
Not sure I fully understand the implications of setting corePackage in pxt.json but seems fine overall

riknoll:
@thsparks it's used by the daldts command in the pxt cli. not sure how it ever worked without that being set tbh, possible that this has been broken for a long time

https://github.com/microsoft/pxt/blob/f53bd26c9ede40b1502a2d8cfd9d2f0bc1211449/cli/cli.ts#L4785

PR #5775, opened 2024-08-01
Update soil moisture code and connect

martinwork:
Arising from support ticket https://support.microbit.org/helpdesk/tickets/80167 (private)

Update for V2 compatibility.
- Replace analogWritePin with digitalWritePin, so the output changes more quickly.
- Add a 1ms pause to ensure the changed input voltage has been sampled by the ADC.
- Remove a semicolon after radio.sendNumber.

@abchatra 
@microbit-carlos

### DIFF
--- a/docs/projects/soil-moisture/code.md
+++ b/docs/projects/soil-moisture/code.md
@@ -78,9 +78,10 @@ This saves electricity and also avoids corrosion of the probes.
 led.setBrightness(64)
 let reading = 0
 basic.forever(() => {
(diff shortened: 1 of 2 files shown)


microbit-carlos:
> * Replace analogWritePin with digitalWritePin, so the output changes more quickly.

This is because of the PWM driver design in V2 it can take a full PWM period to change the pin output, so a digital write is quicker and behaves more similarly in V1 vs V2.
The support issue that Martin linked discovered that the programme needed a pause block after setting the pin output.

abchatra (approved):

PR #5792, opened 2024-08-05
Update unsupportedParts function to include "v2"

microbit-carlos:
Issue description in comment below:
- https://github.com/microsoft/pxt-microbit/pull/5792#issuecomment-2269418798

Test deployment:
- https://makecode.microbit.org/app/9257a550143ce31c0f8614758d8253188ce79522-af4dc12bcb

CC @abchatra

### DIFF
--- a/editor/flash.ts
+++ b/editor/flash.ts
@@ -253,7 +253,7 @@ class DAPWrapper implements pxt.packetio.PacketIOWrapper {
         if (this.usesCODAL === undefined)
             console.warn('try to access codal information before it is computed')
         if (!this.usesCODAL) {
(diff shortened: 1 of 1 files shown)


microbit-robert:
When attempting to flash a micro:bit V1 with code containing blocks that have the `//% parts="v2"` annotation, the "Incompatible Code" dialog is not shown when it should be. This PR amends that behaviour. 

Currently, only [the simulator respects this annotation](https://github.com/microsoft/pxt-microbit/blob/master/sim/dalboard.ts#L140-L146). 

This will benefit extensions already using this annotation, such as the [microbit-v2-power extension](https://github.com/microbit-foundation/pxt-microbit-v2-power).

abchatra (approved):

Issue #5796, opened 2024-08-06
Audio recording length went from 4.5 seconds to 1.3 seconds in beta


microbit-carlos:
This simple programme measures how long we can record audio until hitting the default limit:

<img width="805" alt="image" src="https://github.com/user-attachments/assets/97280be2-8e56-4395-8eea-4de60b04e73d">

```ts
let t = 0
input.onButtonPressed(Button.A, function () {
    t = input.runningTime()
    record.startRecording(record.BlockingState.Blocking)
    basic.showNumber(input.runningTime() - t)
})
input.onButtonPressed(Button.B, function () {
    record.playAudio(record.BlockingState.Blocking)
})
```

On MakeCode live it records around 4.5 seconds of audio: [microbit-recording-live.hex.zip](https://github.com/user-attachments/files/16510401/microbit-recording-live.hex.zip)
- makecode.microbit.org version:  [6.0.28](https://github.com/Microsoft/pxt-microbit/releases/tag/v6.0.28)
- Microsoft MakeCode version:  [9.0.19](https://github.com/Microsoft/pxt/releases/tag/v9.0.19).
- microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6).
- codal-microbit-v2 runtime version:  [v0.2.63](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.63)

<br>

On MakeCode beta less than 1.5 seconds: [microbit-recording-beta.hex.zip](https://github.com/user-attachments/files/16510406/microbit-recording-beta.hex.zip)
- makecode.microbit.org version:  [7.0.15](https://github.com/Microsoft/pxt-microbit/releases/tag/v7.0.15)
- Microsoft MakeCode version:  [10.2.4](https://github.com/Microsoft/pxt/releases/tag/v10.2.4)
- microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
- codal-microbit-v2 runtime version:  [v0.2.67](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.67)

riknoll:
is this on hardware or in the simulator?

abchatra:
@carlosperate is this on hardware? Can CODAL update be the reason if on hardware?

microbit-carlos:
This difference is on hardware.

[...]

PR #5805, opened 2024-08-08
audio recording: Set recording splitter channel rate during setup.

microbit-carlos:
Fixes https://github.com/microsoft/pxt-microbit/issues/5796.

Test deployment:
- https://makecode.microbit.org/app/f650a417b2495e02bc009c72ccd5578b96c91a9d-9b51dbe296

### DIFF
--- a/libs/audio-recording/recording.cpp
+++ b/libs/audio-recording/recording.cpp
@@ -38,18 +38,18 @@ static MixerChannel *channel = NULL;
 #endif
 
 
-void checkEnv(int sampleRate = -1) {
+void checkEnv() {
 #if MICROBIT_CODAL
     if (recording == NULL) {
-        if (sampleRate == -1)
-            sampleRate = 11000;
+        int defaultSampleRate = 11000;
         MicroBitAudio::requestActivation();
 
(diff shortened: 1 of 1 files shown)


abchatra:
Just curious, how was this bug introduced?

microbit-carlos:
There are two sample rates to set, one for the microphone with the splitter channel, and another for the sound output via the mixer channel.

CODAL before v0.2.65 had set the ADC sample rate as a magic number in the `uBit.audio` initialisation to 11K. As part of a couple of bug fixes (originally raised in MakeCode and discussed in the codal repo) in the audio pipeline, CODAL updated the splitter to be able to correctly change the upstream (i.e. microphone) sampling rate dynamically, as requested in the audio pipeline. It changed this 11K magic number to a config macro and set the default value to the highest rate (44.1K), as starting high and then going low afterwards yielded better results.

MakeCode depended on this default ADC rate being 11K, which we didn't notice, and we didn't catch it during testing because I think all our test programmes set up the splitter channel rate as soon as it's created.

srietkerk (approved):

Issue #5403, opened 2023-11-10
MakeCode programme can overlap reserved flash pages for BLE and CODAL storage


microbit-carlos:
Originally reported as a comment in:
- https://github.com/microsoft/pxt-microbit/issues/5396#issuecomment-1805425699

In that example, this programme can be correctly flashed to the micro:bit V2:
https://makecode.microbit.org/_9KKHE4WyyUM3
[microbit-largeSwitch.hex.zip](https://github.com/microsoft/pxt-microbit/files/13320016/microbit-largeSwitch.hex.zip)

But the programme goes all the way in flash up to address 0x752B0, writing to flash memory areas reserved for CODAL storage usage and Nordic SDK BLE bond data: https://github.com/lancaster-university/codal-microbit-v2/blob/master/docs/MemoryMap.md

A programme that doesn't use CODAL storage nor BLE will probably not trigger a problem, but a soon as those features are used it can cause issues.

microbit-carlos:
I have tested this in the latest MakeCode beta with this programme (a modification of the example programme included in https://github.com/microsoft/pxt-microbit/issues/5396):
https://makecode.microbit.org/_hWEJ3rMpeYKY

This is the larger programme I was able to successfully build for V2, which is at the edge of the 0x73000 limit (flash pages after that are used by CODAL and bootloader). So, I'm guessing somebody has implemented this already.

This issue can be closed as resolved.

Issue #5396, opened 2023-11-08
this program never works on V2


tballmsft:
The following program never works as expected when loaded successfully (no error messages) over WebUSB or file download to micro:bit V2. I suspect it is too large for the V2 as well as V1:

https://makecode.microbit.org/_9KKHE4WyyUM3

I reduced the program size to this:

https://makecode.microbit.org/_Chq7538rzeDL

This program works as expected when successfully loaded.

abchatra:
I can reproduce the issue. We should throw error in v2 if the size is too long for v2

microbit-carlos:
We have some documented memory map info in:https://github.com/lancaster-university/codal-microbit-v2/blob/master/docs/MemoryMap.md

Essentially the programme data needs to stop at address 0x73000, as anything after that is either used by CODAL or the bootloader.
In this case, the large switch programme is writing up to address 0x7D3EC, so it's about 40 KBs larger than would fit in V2.

PR #5814, opened 2024-08-13
add dialog for when program too big for v2

jwunderl:
Give an error dialog when a program is too big for v2: 

image

wording intended to be similar to the 'too big for v1' one just below, but any wordsmithing appreciated.

re: https://github.com/microsoft/pxt-microbit/issues/5396, moving off milestone with this but not closing -- this addresses the intent of the original issue / why it was marked, but the specific problem there is basically https://github.com/microsoft/pxt-microbit/issues/5403; if we have time before release we should address that (e.g. add a 'parts override' field to hardware in pxtarget json, which lets you specify `flashEnd` per enabled part) but not guaranteed.

'way too big' version of program https://makecode.microbit.org/_6oaTMe08Yc7b

requires https://github.com/microsoft/pxt/pull/10090


srietkerk (approved):
Looks good!

PR #5842, opened 2024-08-21
add support for arcadeshield to pxt-microbit

tballmsft:
## TODO (core)

- [x] change pxtarget.json to build both variants
- [x] review all "shim=" and "helper="
- [x] move pure STS code out to pxt-arcadeshield
- [x] make sure all shimmed functions have simulator support
- [x] test on hardware (MicroData)

## Blocks and Docs

- [x] move bitmap.d.ts and all other block defining stuff out to pxt-arcadeshield
- [x] all docs go to pxt-arcadeshield
- [x] remove "game" terminology
- [x] review the various blocks and APIs and add comments
   - [x] Bitmap (used to be Image) interface
   - [x] bitmap namespace
   - [x] screen (is a Bitmap)
   - [x] screenhelpers


tballmsft:
some code

```
controller.A.onEvent(ControllerButtonEvent.Pressed, function () {
    if (controller.A.isPressed()) {
    	
    }
})
let ff = bmp`
    . . . . . . . . . . . . . . . . 
    . . . . . . . . . . . . . . . . 
    . . . . . . . . . . . . . . . . 
    . . . . . . . . . . . . . . . . 
    . . . . . 3 3 3 3 3 3 3 . . . . 
    . . . . 3 3 . . . . . . 3 . . . 
    . . . . . 3 . . . . . . 3 . . . 
    . . . . . 3 . . . . . . 3 . . . 
    . . . . . 3 . . . . . . 3 . . . 
    . . . . . 3 3 . . . . 3 . . . . 
    . . . . . . 3 3 3 3 3 3 . . . . 
    . . . . . . . . . . . . . . . . 
    . . . . . . . . . . . . . . . . 
    . . . . . . . . . . . . . . . . 
    . . . . . . . . . . . . . . . . 
    . . . . . . . . . . . . . . . . 
    `
ff.drawRect(0, 0, 0, 0, 0)
ff.setPixel(0, 0, 0)
let p = ff.getPixel(0, 0)
ff.fill(5)
ff.fillRect(0, 0, 1, 1, 6)
ff.drawLine(0, 0, 10, 10, 5)
ff.drawRect(2, 2, 5, 5, 5)
let c = ff.clone()
ff.flipX()
ff.flipY()
ff.replace(3, 4)
const b = ff.equals(c)
const w = ff.width
const h = ff.height
const b2 = ff.isMono
ff.copyFrom(c)
const buf = Buffer.create(100)
ff.getRows(3,buf)
ff.setRows(3,buf)
ff.transposed()
ff.scroll(2,2)
const dbl = ff.doubledX()
ff.doubledY()

```

sim/state/bitmap.ts
```diff
@@ -0,0 +1,1135 @@
+namespace pxsim {
+    export class RefImage extends RefObject {
```
eanders-ms, line 2: Suggestion: rename RefImage -> RefBitmap (unless you've already tried this and it didn't work for some reason). Avoiding duplicate global symbols is good practice, even if the `screen` and `bitmap` libs are never loaded together.

sim/state/bitmap.ts
```diff
@@ -0,0 +1,1135 @@
...
+    export function doubledIcon(buf: RefBuffer): RefBuffer {
+        let img = ofBuffer(buf)
+        if (!img)
+            return null
+        img = BitmapMethods.doubled(img)
+        return toBuffer(img)
+    }
+}
+
+namespace pxsim.pxtcore {
```
eanders-ms, line 1113: Everything in this namespace is more screen related than bitmap. To separate concerns, do you think it makes sense to move it into a separate file called `sim/arcadeshield.ts`?

libs/bitmap/bitmap.ts
```diff
@@ -0,0 +1,292 @@
...
+            let r = img.transposed();
+            r.flipX();
+            return r;
+        } else {
+            return null;
+        }
+    }
+
+    //% shim=pxt::setScreenBrightness
+    function _setScreenBrightness(brightness: number) { }
```
eanders-ms, line 281: Move brightness functions to arcadeshield?

sim/state/bitmap.ts
tballmsft: as discussed, can't do because of special support in compiler for RefImage

sim/state/bitmap.ts
tballmsft: will do

libs/bitmap/bitmap.ts
tballmsft: removed

eanders-ms (approved):

eanders-ms:
"arcadeshield" and "bitmap" are showing up as available extensions. Should they be hidden?

image

Issue #5530, opened 2024-03-07
Workaround WebUSB DAPLink firmware issue


microbit-carlos:
There is at least one known issue with WebUSB failing that we can apply a workaround in MakeCode.
If the micro:bit has a battery pack connected, when the USB cable is unplugged (in the middle of a WebUSB transaction) it can leave DAPLink in a bad state and WebUSB might stop working when the micro:bit is reconnected:
- https://github.com/microbit-foundation/DAPLink/issues/17#issuecomment-1555034429

This will need a DAPLink firmware update, but a workaround can be applied to MakeCode as described in the DAPLink issue, which is desirable, as many user will never really update their firmware.

microbit-carlos:
@abchatra The general idea of the workaround (from https://github.com/microbit-foundation/DAPLink/issues/17#issuecomment-1555034429):

* When the response for the wrong DAP command is received, start tracking a history of received and expected commands
* Send a couple of innocuous different DAP command, the response value is not that important as we only need to track the command ID in the response message. For example it could read some DAP info
  * https://arm-software.github.io/CMSIS_5/DAP/html/group__DAP__Info.html
* If DAPLink sends a response for the previous command instead we can detect we are in this "one off" state
  * Might be worth sending a couple of different commands to be 100% sure
* From this point forward on this WebUSB session we can send a dummy command in-between real commands, to counteract this "one off" issue
  * So it looks like this:
    * Send command 01 -> get bad response
    * Send dummy command -> get response for command 01
    * Send command 02 -> get dummy response
    * Send dummy command -> get response for command 02

abchatra:
Thanks @microbit-carlos

microbit-carlos:
We talked about it with @abchatra and we think we should first do the following:
- Detect the issue and add it to the metrics
- Create a UX workflow to instruct the user to unplug and replug all micro:bit power sources

https://github.com/microsoft/pxt-microbit/assets/29712657/d9c210c3-aede-4e93-9854-05f6a76da0f1

abchatra:
@jwunderl lets discuss this before starting the work.

microbit-carlos:
Hi @jwunderl, have you had a chance to start looking into this? Let me know if there is anything I can clarify or assist from the DAPLink side. We are keen to have the UX workflow improvements ready for this release.

microbit-carlos:
Another way to trigger this is by refreshing the MakeCode browser tab, as described in:
[...]

PR #5843, opened 2024-08-21
webusb: when connecting, send a few commands to flush commands

jwunderl:
On my machine with a teams call & full server (watched pxt build / pxt serve) I was able to repro the 5 -> 131 almost 100% of the time when refreshing the same page (turns out having an old device helped a bit!). After poking at it for a bit, this fix of moving "command flushing" up / repeating it a few times to ensure it's cleared seems to have fixed that 100% on my machine in that scenario. I don't have as good of repro cases for the other issues but trying to get repros on the other scenarios now.

(test program i was using, modified from a previous one to be more spammy https://makecode.microbit.org/_5j14oYbeTcCC)

I believe this should actually cover the cases that the more comprehensive suggestion in https://github.com/microsoft/pxt-microbit/issues/5530 handles; our typical tactic for errors in operations is 'throw up hands and reconnect when things misbehave', so clearing through this does largely the same without spreading the logic throughout the app (and if not, first step would be to add reconnect attempts in those scenarios.)

### DIFF
--- a/editor/flash.ts
+++ b/editor/flash.ts
@@ -315,7 +315,9 @@ class DAPWrapper implements pxt.packetio.PacketIOWrapper {
         // before calling into dapjs, we use our dapCmdNums() a few times, which which will make sure the responses
         // to commends from previous sessions (if any) are flushed
         for (let i = 0; i < 3; i++) {
-            await this.getDaplinkVersionAsync()
+            try {
+                await this.getDaplinkVersionAsync();
+            } catch (e) {}
         }
 
         // halt before reading from dap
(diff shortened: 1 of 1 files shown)


editor/flash.ts
```diff
@@ -312,15 +312,21 @@ class DAPWrapper implements pxt.packetio.PacketIOWrapper {
 
         await this.io.reconnectAsync()
 
+        // before calling into dapjs, push through a commands to make sure the responses
+        // to commands from previous sessions (if any) are flushed
+        for (let i = 0; i < 5; i++) {
```
thsparks, line 317: Was 5 just arbitrary? If so, a quick update to the comment above might be helpful indicating that.

thsparks (approved):
One nit, otherwise LGTM!

PR #5856, opened 2024-08-23
Clear previous commands before starting flash as well

jwunderl:
Was seeing some failures along this path with e.g. 5->0 output while reproducing the memory leak, clearing in the same way as https://github.com/microsoft/pxt-microbit/pull/5843 appears to mitigate this path as well on my device as a stopgap.


thsparks (approved):
👍

Issue #5733, opened 2024-07-17
Digital pin vs analog pin blocks


microbit-carlos:
### Summary

The "analog pin" dropdown lets you select invalid pins for the `[analog read pin [P0]]` block, without showing any warning/errors, and it simulates it successfully, even though it will not work on the device.

### Current situation
In MakeCode v6 we have pin blocks with two kinds of pin dropdown widgets:

- The digital pin dropdown, used for digital read, digital write and more blocks like the pulse-in, etc
  <img width="400" alt="image" src="https://github.com/user-attachments/assets/d1e2ab2d-5bc7-4ae9-b98a-28167a9324f3">
- And the analog pin dropdown, used analog read, analog write and more blocks like the servo, etc
  <img width="400" alt="image" src="https://github.com/user-attachments/assets/d56abab6-4489-4222-8931-6c368f5c4b21">

### The problem
This is all good for the digital read, digital write and analog write blocks. However, for the analog read block, the analog pin dropdown let's the user select an invalid pin, it doesn't show any warning on the block, and it simulates the ADC in the simulator:

<img width="400" alt="image" src="https://github.com/user-attachments/assets/3da0a4df-ffc7-4ec9-b744-b813575d92e2"> 

This programme will not work on the device, as P9 is not an ADC pin.

### New blocks
Since the introduction of pin parameters in PR #5700, we now also have the option to set a variable to a "digital pin" or "analog pin".

<img width="150" alt="image" src="https://github.com/user-attachments/assets/60f32b93-c582-450d-bc1a-521664a42575">

So this could add an additional layer of confusion to the user, as you can use the "wrong" pin block type and it will work in some cases, but might not in others.

This this example we are:
- Using an analog pin inside a `[digital write pin[ _ ]]` block, works ✅
- Using digital pin 0 inside a `[analog read pin[ _ ]]` block, works ✅
- Using digital pin 16 inside a `[analog read pin[ _ ]]` block, does not works ❌

<img width="428" alt="image" src="https://github.com/user-attachments/assets/169db94f-0ed7-40db-9b0e-8cea91cb410c">

### Suggestions

These are not final and are open for discussion.

We could:
- Have a "pin" like the current "digital pin" block
- Have a "analog read pin" block with only the ADC pins
- Digital read/write pin blocks could accept either block
- Analog write (PWM) blocks could accept either block
- Analog read (ADC) blocks should only accept the "analog read pin" block
    - A warning could be shown in the block if a different pin is inserted

[...]

PR #5863, opened 2024-08-26
add new pin shadow blocks and field editor

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5733

This PR does a few things:

1. Creates three new abbreviated shadow blocks that only say `P0` instead of `analog pin P0`
2. Changes the shadow block for the `analog read` block so that it only lists the analog-readable pins
3. Adds a new field editor which displays a warning when a non-analog-readable pin is placed inside of the `analog read` block (see GIF below). This field editor is otherwise identical to the gridpicker
4. Renames the `pins.analogPin` and `pins.digitalPin` functions so that they start with an underscore, which means they get filtered out of the TS intellisense
5. Removes the `(write only)` text from the analog pin selector since the new field editor gives warnings instead

Here's what the warning looks like:

writeonly-pin

@sae220 FYI


riknoll:
oh also, this PR relies on https://github.com/microsoft/pxt/pull/10118 but is safe to merge without it!

libs/core/pins.ts
```diff
@@ -8,26 +8,79 @@ namespace pins {
...
      * Returns the value of a C++ runtime constant
      */
-    //% help=pins/digital-pin blockGap=8 advanced=true shim=TD_ID
-    //% blockId="digital_pin" block="digital pin %pin"
-    //% pin.fieldEditor="gridpicker" pin.fieldOptions.columns=4
+    //% help=pins/digital-pin
+    //% shim=TD_ID
+    //% blockId=digital_pin
+    //% block="digital pin $pin"
+    //% pin.fieldEditor=pinpicker
```
srietkerk, line 15: I see that the field editor for the gridpicker had quotes around it, are the quotes not needed?

srietkerk (approved):
One question for my learning, but otherwise looks good!

libs/core/pins.ts
riknoll: nope! they don't hurt anything, i just like no quotes better

kimprice (approved):

abchatra:
@carlosperate @jaustin please take a look at /beta and ensure all issues of pins are addressed as expected.

Issue #5869, opened 2024-08-28
Beta: analog pin block is misplaced


martinwork:
**Describe the bug**

The analog pin block is at the bottom of ...more under micro:bit V2

**Expected behavior**
digital pin and analog pin together

**Screenshots**
![image](https://github.com/user-attachments/assets/e3b03c99-2c82-401e-b49f-484f0813f780)
![image](https://github.com/user-attachments/assets/426c6b85-ed25-48a2-8621-45182e6d29ce)

Issue #5834, opened 2024-08-19
Beta: Importing shared project loses pin choices


martinwork:
**Describe the bug**

Importing into beta a project shared from live loses the pin choices
https://makecode.microbit.org/_FT33VLeqa00v

**To Reproduce**
1. Import https://makecode.microbit.org/_FT33VLeqa00v to beta
2. See error

**Expected behavior**
Pin choices are retained

**Screenshots**
Original
![image](https://github.com/user-attachments/assets/d90e2d5f-05b9-4f2a-9f04-ad972b5af692)

In beta
![image](https://github.com/user-attachments/assets/fa93f4bb-48b0-4fcc-a274-774aaa6ab215)

**micro:bit version (please complete the following information):**
not hardware related

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows 10
 - Browser [e.g. chrome, safari] Chrome

martinwork:
Created in live: https://makecode.microbit.org/_E7sPiPgDJ9a0
![image](https://github.com/user-attachments/assets/10dad2ec-d130-40f2-ac22-b330b9cf77e3)

Imported to beta: makecode.microbit.org version:  7.0.26; Microsoft MakeCode version: 10.2.16
![image](https://github.com/user-attachments/assets/1dbf5136-d4ea-41d1-a24f-fe62ec0b2cdb)

This looks OK, except analog read and write dropdowns are the same (and pins are out of order):
![image](https://github.com/user-attachments/assets/177a43d6-c121-4b46-ab09-53eabac940eb)

A newly created analog read is different
![image](https://github.com/user-attachments/assets/9b669c5d-6f9e-4199-aa97-e4dae04daecf)
![image](https://github.com/user-attachments/assets/eb817d75-2099-419f-ba39-eff299557710)

martinwork:
After sharing the same project in JS mode: https://makecode.microbit.org/_bM66wkK9ECCM
![image](https://github.com/user-attachments/assets/5dcde94c-02b6-4a99-bd48-9d120697ed76)

Import to beta looks OK
![image](https://github.com/user-attachments/assets/2e50e8ab-3633-4eab-951e-0e19e9b9da50)

But blocks mode isn't right
![image](https://github.com/user-attachments/assets/e4a61146-5937-4558-8ed1-2235065592da)

martinwork:
When creating a new project in beta, the instruction is different when added in JS mode (P0) vs blocks mode (P1):
[...]

PR #5871, opened 2024-08-28
Reorder pins, fix pin block location, improve pin patch

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5869
Fixes most of https://github.com/microsoft/pxt-microbit/issues/5834

A few fixes:
* restores the position of the analog pin block which I accidentally deleted
* removes the unnecessary name overrides in the analog pin enum
* moves pin 10 back into numerical order instead of the read-write pins first order that we use in live
* improves the upgrade rule for the analog read pin block so that it uses the new shadow if a read-write pin is selected


libs/core/_locales/core-strings.json
```diff
@@ -7,19 +7,6 @@
   "AcceleratorRange.OneG|block": "1g",
   "AcceleratorRange.TwoG": "The accelerator measures forces up to 2 gravity",
   "AcceleratorRange.TwoG|block": "2g",
-  "AnalogPin.P11|block": "P11",
```
srietkerk, line 10: Why are all of the analog write pins getting removed?

srietkerk: Is this the "removes the unnecessary name overrides in the analog pin enum" part of the PR?

riknoll: correct. they used to be named something like `P11 (write only)` but I removed that in my last PR so these translations are no longer required

jwunderl (approved):

thsparks (approved):

Issue #5878, opened 2024-08-29
"on pin pulse" block and variables as inputs


microbit-carlos:
A continuation of the discussion from:
- https://github.com/microsoft/pxt-microbit/pull/5752#issuecomment-2313579659

With the latest pin changes (thanks @sae220 and @riknoll!), we have the "on pin pulse" block which can take an input to indicate which pin to select.

With the shadow block we can select any available pin:
<img width="383" alt="image" src="https://github.com/user-attachments/assets/155f5066-4233-4ba8-ab5c-a2608794d813">

We can also add a "digital pin" or "analog pin" block as an input:

<img width="423" alt="image" src="https://github.com/user-attachments/assets/8fd328ea-f551-494a-8e81-5abb7cd33aa7">

<br>

But when adding a variable we encounter a warning, for the reasons listed in https://github.com/microsoft/pxt-microbit/pull/5752#issuecomment-2313579659:
<img width="591" alt="image" src="https://github.com/user-attachments/assets/5d2721a6-621d-4bee-ac29-3d41fc3490ed">

So, it's not too clear what might be the advantage to have a shadow block in this block, if dropping a variable cannot work and dropping a "digital pin" or "analog" pin block does not offer any advantages over the shadow block itself.

To avoid confusion with the variable warning perhaps we should revert this block to simply have a dropdown like in Makecode v6?
<img width="325" alt="image" src="https://github.com/user-attachments/assets/683cb36f-2062-4a4d-8c35-7878e12163ba">

riknoll:
yeah, i think it makes sense to revert this single block. i'll open a PR

PR #5879, opened 2024-08-29
Revert the pin pulse event

riknoll:
Fixes https://github.com/microsoft/pxt-microbit/issues/5878

Reverting the on pulse event so that it no longer accepts a variable given that using a variable doesn't actually work.

Looks like someone didn't check in a change to the shims file, you can safely ignore the change to `is-sound-playing`

### DIFF
--- a/libs/core/pins.cpp
+++ b/libs/core/pins.cpp
@@ -241,11 +241,12 @@ namespace pins {
     */
     //% help=pins/on-pulsed advanced=true
     //% blockId=pins_on_pulsed block="on|pin %pin|pulsed %pulse"
-    //% pin.shadow=digital_pin_shadow
+    //% pin.fieldEditor="gridpicker" pin.fieldOptions.columns=4
+    //% pin.fieldOptions.tooltips="false" pin.fieldOptions.width="250"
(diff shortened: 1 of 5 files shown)


srietkerk (approved):

libs/core/pins.cpp
```diff
@@ -241,11 +241,12 @@ namespace pins {
     */
     //% help=pins/on-pulsed advanced=true
     //% blockId=pins_on_pulsed block="on|pin %pin|pulsed %pulse"
-    //% pin.shadow=digital_pin_shadow
+    //% pin.fieldEditor="gridpicker" pin.fieldOptions.columns=4
```
srietkerk, line 244: Just curious, why didn't you update this to use the "pinpicker"?

riknoll: that field editor is only for shadow blocks that can be dragged into other blocks. probably wouldn't hurt, but it wouldn't do anything in this case

srietkerk: Ah, gotcha, thanks for the info!

jwunderl (approved):

PR #5882, opened 2024-08-30
add timeout when retrying failed command

jwunderl:
Make use of https://github.com/microsoft/pxt/pull/10139

Also adds a tick for when retrying fails in that scenario. Still haven't been able to reproduce the errors myself to confirm, but theoretically this should at least not leave it stalled on download (and may allow it to be addressed by existing error handling, if my hunch listed in that pr is correct)


jwunderl:
Going to merge this one and ping to get reviews tomorrow, so that it is in there if anyone tests overnight / it's in testing tomorrow

thsparks (commented):
Retroactive approve 😁

jwunderl:
😁 was just about to ping in chat thanks

Issue #5409, opened 2023-11-17
randint() not random when bluetooth enabled


Amerlander:
**Describe the bug**
When pxt is set to have bluetooth anabled the randint() block always produces the same number. Regardless of v1 or v2.

**To Reproduce**
1. Go to https://makecode.microbit.org
2. Add bluetooth extension
3. Set code to `basic.showNumber(randint(0, 9))`
4. See every time the microbit starts, the same number (one) is displayed

Alternative check this shared project where ble was enabled without loading the bluetooth extension: https://makecode.microbit.org/S91179-28567-56186-10264

**Expected behavior**
I expect the microbit to show a random number that is not predictable.

**micro:bit version (please complete the following information):**
Tested on 1.3, but happens on v1 and v2 afaik.

**Additional context**
On v1 the seedrandom function checks if ble is enabled. If so it tries to use nrf functions for generating a seed and if that fails it is a fix seed. On v2 it is always a fix seed if ble is enabled: https://github.com/lancaster-university/microbit-dal/blob/602153e9199c28c08fd2561ce7b43bf64e9e7394/source/core/MicroBitDevice.cpp#L340

The init function cals `seedrandom()`: https://github.com/lancaster-university/microbit/blob/1d1baa561f82ab0d16d1916b0026955c67d95e74/source/MicroBit.cpp#L118C18-L118C18
But since it gets called before ble is started, it should not matter what value in `BLE_ENABLED` is set.

- I tried to reproduce this error by enabling bluetooth in the microbit samples, but I was not able to reproduce it. I always get random numbers on the samples.
- I tried to replace the static seed fall back with random numbers generated by the sensors, but that still fails.
- `randint()` is [defined in pxt](https://github.com/microsoft/pxt/blob/9ef1988d50f506cdd757c9318ef18360ad264f9d/libs/pxt-common/pxt-core.d.ts#L419), but I didn't figure out how this actually calls the microbit random number functions. Since I was not able to reproduce this issue in the samples, I wonder if there is something on the pxt side I'm missing.

abchatra:
@carlosperate @JohnVidler

abchatra:
I can reproduce the bug on v2 hardware. The sequence of numbers are always 1,7,8 etc when the hardware is reset and random number is generated on button A pressed.

https://makecode.microbit.org/_EaXWh9CzjFWp

microbit-carlos:
[...]

PR #5945, opened 2024-09-13
Remove RNG seed call in platform_init().

microbit-carlos:
This call is not needed, as the random number generator is already seeded in uBit.init().

This also resolves/works-around this issue (more info in https://github.com/microsoft/pxt-microbit/issues/5409#issuecomment-2349284769), which might take a while before we fix it in DAL:
- https://github.com/microsoft/pxt-microbit/issues/5409

Apart from that, this is a costly call, specially in V1 where it takes 50 microsec or more, and in theory could take up to 480 us.

### DIFF
--- a/libs/core/codal.cpp
+++ b/libs/core/codal.cpp
@@ -50,7 +50,6 @@ MicroBitEvent lastEvent;
 bool serialLoggingDisabled;
 
 void platform_init() {
-    microbit_seed_random();
     int seed = microbit_random(0x7fffffff);
     DMESG("random seed: %d", seed);
     seedRandom(seed);


microbit-carlos:
Using the testing deployement :
- https://makecode.microbit.org/app/f7dc50807dd9f63550b1dc9dd555ae6eec6ada7a-fdd3896247#editor

Creating a new project, adding the Bluetooth extension and this code:

```ts
basic.showNumber(randint(0, 9))
basic.forever(function () {
	
})
```

And it shows that it does resolve:
- https://github.com/microsoft/pxt-microbit/issues/5409

mmoskal (approved):

PR #5961, opened 2024-09-18
Match music play hints with toolbox blocks in 'V2 Hamster'

### LINKED ISSUE #5940: Incorrect Code Validation Error in Hamster Tutorial
On step 4 of the hamster tutorial (or any other step expecting `play sound` blocks), we get a code validation failure popup even if the correct block is in place. Probably something to do with the switch to non-deprecated blocks, though the hint looks correct so I'm not sure yet where the issue lies, if it's just updating tutorial content or some issue with the compilation/block parsing.

image

image

ganicke:
The `play()` function in the hints mismatched with `_playDefaultBackground()` from the toolbox blocks causing code validation to fail for valid blocks. The `play()` functions in the hints are changed to match.

Closes #5940

### DIFF
--- a/docs/projects/v2-countdown.md
+++ b/docs/projects/v2-countdown.md
@@ -99,7 +99,7 @@ for (let index = 0; index <= 2; index++) {
     basic.showNumber(3 - index)
 }
 // @highlight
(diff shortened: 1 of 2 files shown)


thsparks (approved):
Interesting...good find! We may need to look into whether or not we actually ought to be decompiling into this function, but that's a different issue. For now, this seems like a good fix to me!

PR #5983, opened 2024-10-09
Bitmap rename

tballmsft:
renaming namespace "bitmap" to "bitmaps" so we can have top-level variables (for blocks) named "bitmap".


eanders-ms (approved):
Nice this hasn't shipped yet. No need to be concerned about breaking existing projects.

tballmsft:
correct. we are in beta for a bit...

PR #5987, opened 2024-10-09
Prepare a test doc folder for 'csintro' updates

ganicke:
Added a new `./test` docs folder for testing course updates for `csintro`. 

This will path at: http://makecode.microbit.org/test/courses/csintro

Note: Pages are path prefixed at `./test/` and will need to be rolled back when merged into the `./courses` folder. New images are staged in `./test/static` as well.


Jaqster (approved):
Thanks Galen!

PR #6001, opened 2024-10-31
Add Slovene as available language

filips123:
This PR adds Slovene as an available language.

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -532,6 +532,7 @@
             "ru",
             "si-LK",
             "sk",
+            "sl",
             "sr",
             "sv-SE",
             "tr",


abchatra:
@filips123  happy to add this language. Can you test the language in the editor by using the following link? 
https://makecode.microbit.org/?forcelang=sl

We don't expect tutorials to be translated, however all blocks should be translated before we enable the language. 
Here are the docs to help translate in crowdin: https://makecode.microbit.org/translate

filips123:
Yeah, there are still some blocks that are not translated yet...

abchatra:
@filips123 let us know once the translation is completed. Happy to merge this and enable Slovene.

PR #6019, opened 2024-11-25
remove old way of doing arcadeshield simulator

tballmsft:


riknoll (approved):

abchatra:
@tballmsft do you need tags? This will show up for everyone in the live. Not having tags it will only show up in the search. 

Also we should discuss with micro:bit foundation and the team before we add new tags.

image

PR #6034, opened 2024-12-11
Add AI Questions for All Rubrics

thsparks:
@kiki-lee and I went through and tried to determine learning objectives for each tutorial, and then tried to encapsulate those into AI questions.

I tried to keep them to a single sentence / question on the hypothesis that this feature will be received better if teachers don't feel they need to construct a whole preamble to their question before asking it. If that turns out to be untrue, we can adjust, but we should also increase the text field to be multi-line, since it currently gets cut off.


docs/teachertool/checklists/level.json
```diff
@@ -1,6 +1,19 @@
 {
     "name": "Level",
     "criteria": [
+        {
+            "catalogCriteriaId": "499F3572-E655-4DEE-953B-5F26BF0191D7",
+            "instanceId": "zBovdISYJKVRHRGeI1hiZ",
+            "params": [
+                {
+                    "name": "question",
+                    "value": "Does this illustrate an understanding of how to use conditionals to respond to data from the micro:bit sensors?"
```
srietkerk, line 10: This might be semantic and one way might be better for the AI than the other, but I just wanted to ask. Is "using conditionals to respond to data" actually what's going on in this tutorial? I just have a hard time grasping what using conditionals to respond to data looks like since conditions usually don't do the responding. But I'm not sure my suggestion improves it much..
```suggestion
                    "value": "Does this illustrate an understanding of how to use conditionals when responding to data from the micro:bit sensors?"
```

docs/teachertool/checklists/morse-chat-mbv2.json
```diff
@@ -1,6 +1,32 @@
...
+                }
+            ]
+        },
+        {
+            "catalogCriteriaId": "499F3572-E655-4DEE-953B-5F26BF0191D7",
+            "instanceId": "3WfHJ3ZrAfD8pUZgBAOfI",
+            "params": [
+                {
+                    "name": "question",
+                    "value": "Does this send different messages in response to different events?"
```
srietkerk, line 23: Do we want both of these questions? They seem to be asking something similar.

docs/teachertool/checklists/love-meter.json
```diff
@@ -1,6 +1,19 @@
 {
     "name": "Love Meter",
     "criteria": [
+        {
+            "catalogCriteriaId": "499F3572-E655-4DEE-953B-5F26BF0191D7",
+            "instanceId": "b6f1ZKD6IyhuZPrrO5yhl",
+            "params": [
+                {
+                    "name": "question",
+                    "value": "Does this illustrate an understanding of how to display a random number in response to input?"
```
srietkerk, line 10: I think this question needs some tweaking. I know showing a number is the concrete end of the tutorial, but is it the most important that a number is displayed? What if a student decided to generate a random number and, depending on the number gotten, displayed different images? 
```suggestion
                    "value": "Does this illustrate an understanding of how to use the display in response to input?"
```

thsparks: @kiki-lee thoughts?

docs/teachertool/checklists/morse-chat-mbv2.json
srietkerk: Yeah, I think that helps!

srietkerk (approved):
The remaining feedback I have is not blocking and I know this would be good to get in before testing. Looks good!

docs/teachertool/checklists/love-meter.json
kiki-lee: I'd say that having something to do with the random number is most important.  It's fine as-is for now and can be tweaked if there are any changes requested by the group this afternoon.

kiki-lee (approved):
These look good.  I did only look at the values, though. I can't speak to formatting :)

PR #6052, opened 2025-01-06
Add link to I2C read/write buffer

mzaffalon:
Add links for the `pins.i2cReadBuffer()` and `pins.i2cWriteBuffer()` reference pages to the Pins main page.


abchatra:
@ganicke can you review this?

ganicke (approved):
These are hidden blocks but it's OK to link to them I suppose.

PR #6076, opened 2025-01-28
Allow-list Code Eval in Microbit

thsparks:
This creates an override for the `robots.txt` file in the `docs` directory to allow for code evaluation in the Micro:Bit experience. My understanding is that this will overwrite the file in pxt's common-docs folder, so only micro:bit will serve this version without --eval blocked. It also adds code-eval as a supported experience in Micro:bit.

Closely tied to changes in https://github.com/microsoft/pxt/pull/10357

### DIFF
--- a/docs/robots.txt
+++ b/docs/robots.txt
@@ -1,0 +1,5 @@
+# robots.txt for PXT Micro:Bit
+
+# Disable crawling in /v2 path for micro:bit
+User-agent: *
+Disallow: /v2
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -438,6 +438,9 @@
             "identity",
             "blocksErrorList"
         ],
+        "supportedExperiences": [
+            "code-eval"
+        ],
         "bluetoothUartFilters": [
             {
                 "namePrefix": "BBC micro:bit"


pxtarget.json
```diff
@@ -438,6 +438,9 @@
             "identity",
             "blocksErrorList"
         ],
+        "supportedExperiences": [
+            "code-eval"
```
srietkerk, line 442: Why is the term here code-eval instead of eval?

srietkerk: Ah, nevermind, I saw where it's defined in the other PR

docs/robots.txt
```diff
@@ -0,0 +1,5 @@
+# robots.txt for PXT Micro:Bit
+
+# Disable crawling in /v2 path for micro:bit
+User-agent: *
+Disallow: /v2
```
srietkerk, line 5: I was looking into this a bit more, we might want to explicitly allow --eval for microbit?
image
Found here: https://developers.google.com/search/docs/crawling-indexing/robots/create-robots-txt#create_rules , scroll down to "Google's crawlers support the following rules in robots.txt files"

thsparks: I didn't really see a reason to do that since there shouldn't be any "disallow" for the --eval ever seen on this domain. (I believe this file will fully override what we serve when going to https://microbit.makecode.com/robots.txt, so the disallow in common-docs should never be seen).

My reading of the docs is that "allow" would only be needed if we disallowed it for * agents but wanted to explicitly allow it for some other agent.

srietkerk: Ohh, I see, that makes sense.

srietkerk (approved):

PR #6108, opened 2025-03-04
Fix Build w/ Theming

thsparks:
The new theming changes in pxt (https://github.com/microsoft/pxt/pull/10391) removed some less variables and set others to css variables, which cannot be used in less color-manipulation functions (like lighten, darken, etc...). As a result, the build for microbit broke when referencing newer versions of pxt (or local).

In this change, I fix those css build errors, add a default theme, and update themepack variables to reference new theme variables. In time, we may want to fully swap over from themepack to this, but this is a stop-gap until we're focused on micro:bit again.

This isn't meant to look perfect. I mostly want to make sure everything can still build. Making it look nice is secondary, and more likely something we'll focus on when it comes time for our next micro:bit release.

Upload target: https://makecode.microbit.org/app/541b9990739c539a2346b9a74c60603d2d52cea2-d3e5aa7aa9


srietkerk (approved):
LGTM! Just want to comment for a record for when we work on micro:bit (I can also open an issue), but the hover for the cloud projects tile gets rid of the text and changes the colors.

image

thsparks:
> LGTM! Just want to comment for a record for when we work on micro:bit (I can also open an issue), but the hover for the cloud projects tile gets rid of the text and changes the colors.
> 
> image

Good catch. I looked into this briefly, and I think it'll be a fix in pxt rather than micro:bit, so I'll fix it over there in a separate change.

PR #6113, opened 2025-03-13
Keyboard accessibility improvements to the micro:bit simulator.

glastonbridge:
Provides the following accessibility improvements in the micro:bit simulator:
* re-structures the DOM so that tab ordering of simulator elements is left-right top-bottom (see videos)
* makes the antenna element interactive using keyboard navigation, simplest approach
    * previously it was an invisible tab stop
* add some preventDefaults to prevent the screen scrolling while you're adjusting values when the simulator has a scrollbar

## Old behaviour

Sequential tabbing through the simulator using the tab key

https://github.com/user-attachments/assets/130633e1-2751-45ea-b9b2-0e5179cfb8b6

## New behaviour
https://github.com/user-attachments/assets/5238ff6b-76bb-4ff0-87cd-fccb321d7157

[You can try it here](https://simulator-keyboard-navigatio.pxt-microbit.pages.dev/).

[Example project](https://makecode.microbit.org/_0qwJ3JKjzerk) (needs JavaScript copy/pasting to correct deployment).

### DIFF
--- a/sim/visuals/microbit.ts
+++ b/sim/visuals/microbit.ts
@@ -143,6 +143,7 @@ namespace pxsim.visuals {
             outline: none;
         }
         *:focus .sim-button-outer,
+        .sim-antenna-outer:focus,
         .sim-pin:focus,
         .sim-thermometer:focus,
         .sim-shake:focus,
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
For context, Alex (@glastonbridge) is working with the micro:bit team on MakeCode accessibility and integrating Blockly keyboard navigation. As the sim is between the toolbar and the workspace in tab order it’s really helpful to get a logical sequence in place.

thsparks (approved):
Thank you for tidying all this up. Makes the code **much** easier to read. The changes look good to me!

PR #6141, opened 2025-04-04
Extension: Add Sound Level in decibels (dB).

microbit-carlos:
A simple software-only extension to add "sound level (dB)" block, currently in beta:

image

One thing to note is that this extension adds a block to the "Input ... more" category, instead of creating a new one. This is not something that we should normally approve for third party extensions, but we'd like to make an exception for this extension from the Foundation.

image


microbit-carlos:
@abchatra any chance you could have a look at this one? This is the software-only extension I mentioned a few months ago that will be used very soon with some teaching resources.

abchatra (approved):

PR #6143, opened 2025-04-14
Recover pin labels for simulator

microbit-grace:
This pin labels went missing from changes in https://github.com/microsoft/pxt-microbit/pull/6113.

The sim-label svg paths are moved to be part of buildPinElements so that the paths can be added on top of the pin elements instead of below.

You can try it here: https://sim-pins.pxt-microbit.pages.dev/#editor


microbit-grace:
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```

@microsoft-github-policy-service agree company="Micro:bit Educational Foundation"

PR #6170, opened 2025-04-30
Update running time api reference doc with overflow note

### LINKED ISSUE #5372: Document `input.runningTimeMicros()` overflows after 18 minutes
This caught me by surprised and it's a MakeCode specific limitation, as CODAL returns a 32 bit unsigned int and MakeCode drops the 2 most significant bits:
https://github.com/microsoft/pxt-microbit/blob/975c890f2b4b1cf6534ee893b91c8e740739824f/libs/core/control.cpp#L239-L245
I guess the signed bit makes sense, but don't know why the second bit is dropped as well.

It was done in commit https://github.com/microsoft/pxt-microbit/commit/15b4b47e6dbeb40026e6211a3f8e5706c456791e as part of PR https://github.com/microsoft/pxt-microbit/pull/2572 (512b4dd2af68171f8843d4c602c061d333b62971).

This should probably be documented, as 18 minutes is a relatively short time that can be easily reached.

Docs:
https://github.com/microsoft/pxt-microbit/blob/975c890f2b4b1cf6534ee893b91c8e740739824f/docs/reference/input/running-time-micros.md#L1-L16

ganicke:
Add a mention for the possibility of overflow of the running time counter.

Closes #5372

### DIFF
--- a/docs/reference/input/running-time-micros.md
+++ b/docs/reference/input/running-time-micros.md
@@ -1,17 +1,33 @@
-# Running Time Micros
+# running Time Micros
 
-Find how long it has been since the program started in micro-seconds.
+Find how long a program has run since it was started, in microseconds.
 
 ```sig
-input.runningTimeMicros();
+input.runningTimeMicros()
+```
+
+### ~ alert
+
+#### Running time maximum count
+
+The program running time counter can only count up to approximately 17 minutes of time (1,073,741,823 microseconds). After that, the running time will reset to zero and start counting up again. If a program is using the running time counter, it should rely on the counter for only this amount of time or less. Otherwise, the program will need to account for the reset and add the difference to any accumulated time it is counting.
+
+### ~
+
+## Example
+
+Pause for one second and then show the program running time on the screen.
+
+```blocks
+basic.pause(1000)
+basic.showNumber(input.runningTimeMicros())
 ```
 
 ## Returns
 
-* the [Number](/types/number) of microseconds since the program started.
-(One second is 1000000 microseconds.)
+* the [number](/types/number) of microseconds since the program started.
+(One second is 1,000,000 microseconds.)
 
 ## See also
 
-[show number](/reference/basic/show-number), [pause](/reference/basic/pause)
-
+[show number](/reference/basic/show-number), [pause](/reference/basic/pause), [running time](/reference/input/running-time)
--- a/docs/reference/input/running-time.md
+++ b/docs/reference/input/running-time.md
@@ -1,9 +1,9 @@
 # Running Time
 
-Find how long it has been since the program started in milli-seconds.
+Find how long a program has run since it was started, in milliseconds.
 
 ```sig
-input.runningTime();
+input.runningTime()
 ```
 
 ## Returns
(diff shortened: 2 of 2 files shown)


docs/reference/input/running-time-micros.md
```diff
@@ -1,17 +1,33 @@
-# Running Time Micros
+# running Time Micros
 
-Find how long it has been since the program started in micro-seconds.
+Find how long a program has run since it was started, in microseconds.
```
ganicke, line 3: Is this really since program start time or power on/reset?

docs/reference/input/running-time.md
```diff
@@ -1,9 +1,9 @@
 # Running Time
 
-Find how long it has been since the program started in milli-seconds.
+Find how long a program has run since it was started, in milliseconds.
```
ganicke, line 3: Is this really since program start time or power on/reset?

abchatra:
@carlosperate  can you review this change?

docs/reference/input/running-time-micros.md
```diff
@@ -1,17 +1,33 @@
-# Running Time Micros
+# running Time Micros
```
microbit-carlos, line 1: @abchatra not sure what the convention is for capitalisation here, some docs start with a capital letter (normal with titles) and others start lower case to follow the camelCase of the API.

docs/reference/input/running-time-micros.md
microbit-carlos: On "power on" and/or reset the timers should be reset, so this should be the time since programme start.

docs/reference/input/running-time-micros.md
```diff
@@ -1,17 +1,33 @@
...
 ```sig
-input.runningTimeMicros();
+input.runningTimeMicros()
+```
+
+### ~ alert
+
+#### Running time maximum count
+
+The program running time counter can only count up to approximately 17 minutes of time (1,073,741,823 microseconds). After that, the running time will reset to zero and start counting up again. If a program is using the running time counter, it should rely on the counter for only this amount of time or less. Otherwise, the program will need to account for the reset and add the difference to any accumulated time it is counting.
```
microbit-carlos, line 13: I found the second part of this sentence a bit harder to follow, maybe we can rephrase to something like this, and let the user manage it?

>> The program running time counter can only count up to approximately 17 minutes of time (1,073,741,823 microseconds). After that, the running time will reset to zero and start counting up again.
>
>Programs using this time counter should take the reset into account, even if they are only measuring short durations, because the counter might reset while the program is running.

docs/reference/input/running-time.md
microbit-carlos: Since the programme started.

microbit-carlos (commented):
Thanks for the PR @ganicke!

The additions look great, just one small comment.

docs/reference/input/running-time-micros.md
ganicke: The page title follows the API signature.

docs/reference/input/running-time-micros.md
ganicke: Good suggestion.

docs/reference/input/running-time-micros.md
microbit-carlos: Ah, great, thanks @ganicke!

microbit-carlos (approved):
Looks great, thanks @ganicke!

Issue #6068, opened 2025-01-17
Approach to externally stored language preference with iframe


microbit-matt-hillsdon:
**Is your feature request related to a problem? Please describe.**
micro:bit CreateAI shows an iframe-embedded MakeCode after significant prior user interaction. So we have our own language menu (for a subset of the languages MakeCode supports) for all the UI prior to MakeCode.

We pass our language to MakeCode on the URL. This seems to be the only way to communicate it at the moment. We reload it when the language changes in CreateAI.

MakeCode's own language menu doesn't integrate with this scenario. It reloads MakeCode but the language stays unchanged.

I think what happens is that the URL always wins and MakeCode reloads on language change. So MakeCode updates its language cookie (e.g. from en to fr) then reloads, notices en is on the URL and copies en to the cookie again. We noticed this doesn't happen in local dev - in this case for us MakeCode concludes cookies aren't working and actually updates the URL before reload.

**Describe the solution you'd like**
Any reasonable approach that allows us to:

1. set an initial language (in a way that won't result in an extra load)
2. be able to change the MakeCode language from the embedding app
3. be notified when the user changes the language inside MakeCode (so we can update our preference and our own UI) and have that language change actually take effect in MakeCode

**Describe alternatives you've considered**
We considered some other options on https://github.com/microbit-foundation/ml-trainer/issues/580.

A really minimal fix could be a way to suppress the MakeCode language menu option. The embedding app could still offer MakeCode-only languages if desired (we do this in micro:bit classroom).

CC @abchatra - as discussed this week

abchatra:
@riknoll opinion on this?

riknoll:
i think adding an editor message for this would be pretty niche, but i'm fine with suppressing the language option in the context menu. we could add a pxtarget flag and a query variant that flips it on

riknoll:
turns out we already have a pxtarget setting for this, so i've opened a pr to add a query variant

PR #6183, opened 2025-05-05
add query variant to hide the language picker from the settings menu

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6068

adds a query variant for removing the language picker. sites that embed us can add `?hidelanguage=1` to the url to prevent the language picker from showing up in the settings menu (e.g. if the embedding site handles the language selection themselves)

@microbit-matt-hillsdon let me know if this works for you!

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -618,6 +618,11 @@
                 "hideMenuBar": true
             }
         },
+        "hidelanguage": {
+            "appTheme": {
+                "selectLanguage": false
+            }
+        },
         "androidapp": {
             "compile": {
                 "webUSB": false


microbit-matt-hillsdon:
> @microbit-matt-hillsdon let me know if this works for you!

Thanks, yes, we can use this with a micro:bit-classroom-style language dialog in CreateAI that has two tiers of translation (CreateAI+MakeCode and MakeCode only for languages we don't have yet).

srietkerk (approved):

thsparks (approved):

Issue #4381, opened 2021-09-23
Microphone level maxes out at ~191 on hardware


microbit-mark:
Writing the value of sound level to serial, the loudest sound level returned seems to be between 191-195. The returned level should be up to 255

![image](https://user-images.githubusercontent.com/31242877/134515393-76b3cbcc-4ba6-40a1-9e05-4c617ec383d6.png)

https://makecode.microbit.org/_LMvK17UKMaeA

Does not reproduce in MicroPython, so I assume this is MakeCode related, not CODAL.

microbit-mark:
@abchatra can you take a look at this one and try and reproduce?

abchatra:
Does this not reproduce in python when writing to serial? Does this reproduce if we don't write to serial but just displayed on the screen? 

I will take a look.

jaustin:
@carlosperate do you think this is still an issue?

microbit-carlos:
This was fixed in CODAL in:
- https://github.com/lancaster-university/codal-microbit-v2/issues/225

As part of that MakeCode should have been updated to use `uBit.audio.levelSPL->setUnit(LEVEL_DETECTOR_SPL_8BIT)` as described in https://github.com/lancaster-university/codal-microbit-v2/issues/225#issuecomment-1241023894, which I think it might not have been implemented yet:
https://github.com/microsoft/pxt-microbit/blob/v6.1.8/libs/microphone/microphone.cpp

microbit-carlos:
@riknoll updating microphone to use the uBit object was already partially done in https://github.com/microsoft/pxt-microbit/commit/b3a01308db7008aa3be3957d239641192a17542c and the only thing left would be to :
- On init, set the sound level output type via `uBit.audio.levelSPL->setUnit(LEVEL_DETECTOR_SPL_8BIT)`
- `setSoundThreshold()` would call `uBit.audio.levelSPL->setLowThreshold()` or `setHighThreshold()` directly with the user input value
- `soundLevel()` would just return `uBit.audio.levelSPL->getValue()`

PR #6194, opened 2025-05-08
call scaled microphone API instead of using our own scaling

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/4381

see linked issue for the discussion here. swapping out our custom scaling for the scaling that's now built in to the ubit object

@microbit-carlos FYI


microbit-carlos:
Thanks @riknoll!

I just realised that to keep the current behaviour exactly the same we need to set the values in the codal.json file:

```
    "config":{
        "LEVEL_DETECTOR_SPL_8BIT_000_POINT": 52.0,
        "LEVEL_DETECTOR_SPL_8BIT_255_POINT": 120.0,
    }
```
This is because the CODAL default values are currently different, so setting these macros configures the sound level with the same range as before.

The underlying issue with https://github.com/microsoft/pxt-microbit/issues/4381 was that the Sound Level detector never return a value above ~100 dB, so having the 8-bit max point at 120 dB means the programmes we showed an 8-bit value above ~191.

If we set 8-bit max point to 100 dB (so that it uses the full 8-bit range) that will change the scaling, so programmes that before outputted an 8-bit value of `~150` now might output something like `~180`. I'll check with the rest of the team if we would consider that a breaking change, but in the meantime the PR can set the min and max to 52 and 120.

riknoll:
@microbit-carlos if we are just going to set the scaling back to the original value, is there a point to merging this at all?

microbit-carlos:
One of the reasons to have this completely managed by CODAL is because we are doing an audio pipeline refactor and that affects the decibel calculations. As we haven't really exposed the dB sound level in the MakeCode nor the MicroPython API that won't affect most users, but we do need to tweak the 8-bit min and max values to match the current implementation. 

Ideally we'd set MakeCode to use both these default values, and have it completely managed by CODAL, but it wasn't until now that I realised that fixing this bug might break some of the existing user projects.

Btw, I've just release codal v0.2.70 to ensure the micro:bit target uses the default min value as 52 and max value as 100 (I thought this was done in release 0.2.44, but looks like we didn't get around to it).
If the codal tag used in MakeCode is updated to v0.2.70 we can leave the default for `LEVEL_DETECTOR_SPL_8BIT_000_POINT` (before this tag the default value was set to 35) and only set `LEVEL_DETECTOR_SPL_8BIT_255_POINT` to 120 in codal.json.

So, it's not ideal to have the max value configured differently in MakeCode, but based initial testing of the audio refactor branch I think we might only need to tweak the min value, and so that would mean that future CODAL release would not need to be accompanied by a pxt code change.

microbit-carlos:
The v0.2.70 mostly includes minor bug fixes and CI updates, although it does also include a couple of patches to ensure float operations don't unnecessarily perform double operations and that saves about half a kilobyte of flash, and should also provide some minor performance improvements as part of that.

https://github.com/lancaster-university/codal-microbit-v2/blob/master/Changelog.md#v0270

microbit-carlos:
@riknoll good news, after careful consideration the team is happy to include the bugfix as well, so we can let CODAL manage both the min and max values. It still needs the latest v0.2.70 tag, but there is no need to set up any additional config options.

riknoll:
@microbit-carlos that's great! i was actually struggling on figuring out where to inject those options in our toolchain anyhow

srietkerk (approved):

jwunderl (approved):

PR #6250, opened 2025-05-23
Adjust images in browser info page

ganicke:
Modify the images slightly to present better in the page.

RE: https://github.com/microsoft/pxt/pull/10599


Jaqster (approved):
Thanks Galen!

Issue #6271, opened 2025-05-28
Fetches out-of-date extension version


ssande-fwd:
**Describe the bug**
I have an extension that depends on another extension. The version of the dependency is not specified in pxt.json, so it should pick the highest version. Instead it selects an earlier version.

**To Reproduce**
1. Add this extension in MakeCode, https://github.com/ssande-fwd/pxt-solar-fork.
2. Observe that the versions of fwd-energy and fwd-servo-positional are 1.0.0.
3. Go to https://github.com/Forward-Education/pxt-fwd-modules.
4. Observe that 1.0.1 is the newest version.

**Expected behavior**
v1.0.1 is used for fwd-energy and fwd-servo-positional dependencies.

**Additional context**
It works if I hardcode the dependency versions in pxt-solar-fork.

PR #6313, opened 2025-06-11
Extension: add Smart Hydroponics Kit

martinwork:
Arising from support ticket https://support.microbit.org/helpdesk/tickets/88893 (private)
@ssande-fwd
@abchatra

Extension: https://github.com/Forward-Education/pxt-hydroponics
Version: v1.0.3
All blocks: https://makecode.microbit.org/_UV3HC43e1E4a
image


abchatra:
@ssande-fwd can you point to the ticket?

ssande-fwd:
https://github.com/microsoft/pxt-microbit/issues/6271#issuecomment-2917246930

abchatra:
@ssande-fwd We do cache the extensions for 24 hours in the CDN and local cache. So hardcoding is good for testing if you want to test immediately after bumping one after another. However, if you are releasing after testing, I would suggest not hardcoding unless you are worried someone else would change those dependencies without testing. Does it make sense? I will merge this for now.

ssande-fwd:
The fetching of the wrong version persisted for days. I can make another issue if it happens again, but I don't see much incentive to not specify the version. I'll take stability and reliability over convenience, given the frustration I've experienced when it randomly doesn't work as expected and can not be fixed. Martin can correct me if I'm wrong but I think it's even part of the coding standards. I recall him asking me to specify a dependency version where I did not. This has been a surprisingly long and rigorous process, requiring several rounds of particular changes before even reaching the pull request. Thank for merging.

PR #6314, opened 2025-06-11
Extension: add Smart Solder Kit

martinwork:
Arising from support ticket https://support.microbit.org/helpdesk/tickets/88895 (private)
@ssande-fwd
@abchatra

Extension: https://github.com/Forward-Education/pxt-solder-3b3l
Version: 1.0.4
All blocks: https://makecode.microbit.org/_TK6JiAVAdWTM
image


abchatra:
@ssande-fwd I am confused. There is no code in this extension and it is just dependent on Forward-Education/pxt-fwd-modules/fwd-solder-3b3l.  Why not just add that extension to approved list?

ssande-fwd:
Our product ecosystem has collections of jacdac modules. In order to allow those collections to exist independently and overlap and intertwine gracefully I am following a 3 layer architecture. It's outlined in the pxt-fwd-modules readme. The user facing extension layer doesn't introduce functionality. It groups together existing functionalities and hosts tutorial content. A side effect of this architecture is this somewhat silly situation where one of our products only depends on one module. I could break the design but then we have non-coders working on tutorials in a repo with code and increased complexity of sharing modules among different products (user facing extensions), to name a couple downsides. Instead I'm following the architecture and keeping my life simple.

Issue #6011, opened 2024-11-15
Feature request: A way to hide an approved extension from the search results


microbit-carlos:
**Is your feature request related to a problem? Please describe.**
We currently have an approved extension ([pxt-micrbit-ml](https://github.com/microsoft/pxt-microbit/pull/5990)), which to be useful needs additional files to be injected by a different environment that embeds MakeCode.
So, this extension doesn’t offer any usable functionality on its own, and doesn't make much sense for a normal MakeCode user to add it to their project.

**Describe the solution you'd like**
It would be useful to have something like a flag in pxt.json file, to be able to hide an extension from the search.

**Describe alternatives you've considered**
There is already a `hidden` key in `PackageConfig`, but it must have a different purpose as it didn't work in our tests.
https://github.com/microsoft/pxt/blob/v11.3.1/localtypings/pxtpackage.d.ts#L82

We also tried removing the `for PXT/microbit` metadata from the extension README file, but that didn't work either.

In this case the extension needs to be approved, as it contain an editor extension as well, so removing it from the approval list would not be an option.

**Additional context**
N/A

ramseyball:
Try this in your pxt.json

```
    "hidden": true,
    "searchOnly": true,
```

microbit-carlos:
Thanks @ramseyball!
I've just tested it with an approved extension and it still appears on the search (and adding the extension to a project links to the latest version, which does include the changes in pxt.json).

microbit-carlos:
One possible use case would be perhaps a "dependency" that we might want to be approved, so that it can take advantage of features like the extension being cached. For example, pxt-ws2812b is a dependency to pxt-neopixel, it currently appears on the search, but when the user selects it it doesn't load any blocks, which can be confusing.

But, in this specific case, this ticket was raised for CreateAI. The ML extensions cannot do much without the CreateAI editor injecting the model and additional info into a project, so it doesn't make sense for the user to load the ML extension on a normal MakeCode project from the extension search screen.

abchatra:
@riknoll  your plan looks good. Lets honor _hidden_ flag.

PR #6353, opened 2025-06-18
hide microbit-foundation/pxt-microbit-ml from extension search results

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/6011

requires https://github.com/microsoft/pxt/pull/10665 but safe to merge without!


microbit-carlos (approved):
Thanks @riknoll! Having the config flag in the targetconfig.json file sounds good to me 👍

Issue #6094, opened 2025-02-15
audio-recording extension overrides set built-in speak <OFF> and set volume


martinwork:
**Describe the bug**

The audio-recording extension turns on the built-in speaker and sets the volume high for speaker and audio pin (e.g. earphone)
https://github.com/microsoft/pxt-microbit/blob/master/libs/audio-recording/recording.cpp#L55

Workaround: set sample rate to 11000 before adjusting speaker and volume, or adjust after recording.

**To Reproduce**
1. Open https://makecode.microbit.org/_CYtDeAAMMX1p
2. Send to micro:bit
3. Press A
4. Press B
5. Hear speaker

**Expected behavior**
Able to turn off built-in speaker and adjust volume.

**Screenshots**

![Image](https://github.com/user-attachments/assets/e015dd27-f8d6-4165-a9e6-a64371fbfbd1)

**micro:bit version (please complete the following information):**
V2

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] 132

microbit-carlos:
Thanks @martinwork! Do you think this could be an issue in the CODAL layer? Would you be able to create a C++ to check and open an issue there as well if it's the case?

martinwork:
@microbit-carlos The cause is in the extension
https://github.com/microsoft/pxt-microbit/blob/924338d0e048d3edcd4d7d8344901cbffd67741a/libs/audio-recording/recording.cpp#L55

microbit-carlos:
@abchatra assuming the fix is simply to remove that line this could be done as part of any other recording & playback tickets for the 2025 release.

srietkerk:
I just wanted to put some extra context here. You're right, removing the `setVolume` and the `setSpeakerEnabled` below that line does fix both of these issues, but these lines were taken from the original implementation of the extension: https://github.com/lancaster-university/pxt-codal-audio/blob/main/codal-audio.cpp#L55. If I remember correctly, the volume value was set in order to get the audio quality better. From the testing that I did, I can't tell if the audio quality is worse with removing this. I'll have to play with it a bit more. I just wanted to put this here in case there's more context for these decisions.

martinwork:
When used with the extension, instructions for connecting an external speaker or headphones, and controlling the volume don't work.
[...]

Issue #5193, opened 2023-05-26
Volume doesn't affect Sound effect blocks in Simulator


Jaqster:
Using the Set Volume block doesn't seem to affect the Sound Effect blocks (editor and built-in) in the Simulator.
Would expect that this didn't make a sound in the Simulator, but it does.  Works on hardware.
![image](https://github.com/microsoft/pxt-microbit/assets/22108954/d8001e93-79a8-4fdb-91fd-54a2b3da4a9a)

Another note - because the Sound Effect editor includes Volume values not shown on the block, there is potential for conflicts/confusion when using the Set Volume block.
![image](https://github.com/microsoft/pxt-microbit/assets/22108954/c92e2771-f25b-4bfc-aa7d-0226a6b22911)

Issue #5281, opened 2023-06-22
Record sound quality


moverstreet007:
The record sound quality is fine when my students use their laptops. When we do payback using the MicroBit. There is lots of static and the volume is really low.

jaustin:
See https://github.com/microsoft/pxt-microbit/issues/5096 for the micro:bit speaker frequency response. I think we could close one of these two issues

Issue #5096, opened 2023-04-27
recording extension: simulator sounds ... too good...


jaustin:
Given what we know about the sound quality of the micro:bit through the speaker, is it possible to reduce the fidelity of the record/playback that comes from the sim?

What sample rate/bit depth are you using at the moment @srietkerk and is it worth applying a bit of a filter to the audio before playing out that matches the speaker resonance profile so that we don't get people thinking it's not working right when it first runs on their real device?

Here's the speaker data sheet
https://datasheet.lcsc.com/lcsc/1811151451_Jiangsu-Huaneng-Elec-MLT-8530_C94599.pdf
And here's the frequency response:
<img width="838" alt="image" src="https://user-images.githubusercontent.com/147113/235004028-e5163c2b-90b3-4483-bfa9-95d62e905370.png">

PR #6362, opened 2025-06-20
sim audio changes

srietkerk:
Fixes https://github.com/microsoft/pxt-microbit/issues/6094 - removing the volume and speaker setting in `recording.cpp`

The micro:bit side of the fix for https://github.com/microsoft/pxt-microbit/issues/5193 - modifications to  `__playSoundExpression` 

Edit: removed the mention of changes for audio recording quality. All of that has been removed as we have done work in separate PRs for a more encompassing change for mic sensitivity.

Upload target: https://makecode.microbit.org/app/0db8794fd2afe9cf4a0b8bf35ece51de828a92bc-696451be58


microbit-carlos:
I'm not familiar with the Web Audio APIs and their limitations (maybe @microbit-matt-hillsdon, @microbit-robert or @microbit-grace can provide a better insight), but would I be right to assume to using that using `echoCancellation` and `noiseSuppression` can result in "cleaner" microphone playback?
The reason I ask is because ideally we'd prefer simulator to play recordings with a quality more similar to the micro:bit device:
- https://github.com/microsoft/pxt-microbit/issues/5096 
- https://github.com/microsoft/pxt-microbit/issues/5281

If I remember correctly, on CODAL the mic sensitivity is just a gain multiplication that pushes everything down (not just the noise), but I guess you probably don't access to the raw data to do that directly.

microbit-carlos:
One quick thing I noticed trying the test deployment with this programme is that changing the volume while the recording is being played back doesn't have an effect until the next 

```ts
    music.setVolume(32)
    basic.showLeds(`
        . . . . .
        . . . . .
        . . . . .
        . . . . .
        # # # # #
        `)
})
    music.setVolume(255)
    basic.showLeds(`
        # # # # #
        # # # # #
        # # # # #
        # # # # #
        # # # # #
        `)
})
input.onPinPressed(TouchPin.P1, function () {
    record.playAudio(record.BlockingState.Nonblocking)
})
input.onLogoEvent(TouchButtonEvent.Pressed, function () {
    record.startRecording(record.BlockingState.Blocking)
})
```

Also noticed this, which is not directly related to this PR, so I've opened a separate issue:
- https://github.com/microsoft/pxt-microbit/issues/6367

srietkerk:
@microbit-carlos, yes, having `echoCancellation` and `noiseSuppression` set does help the audio have the highest quality. I can change it to where I don't set them both to true to lower the quality a bit, but I don't want to tamper with the low mic sensitivity too much because if the audio quality for the micro:bit itself improves, then we would have to make more changes.

In terms of the approach, we can't use the web api `gain` because that just controls volume and not quality. There are other filters that we might be able to apply through different audio nodes, but it's a bit tricky because the way that we handle the recording right now, with an `HTMLAudioElement`, does not play well with audio nodes and contexts. Using the settings was the most direct way to influence the audio quality without having to do a large overhaul. I'm going to poke at this and play with it a bit more to see if I can make some audio changes without a complete refactor. In regards to your other comments, trying to change the volume mid-play has similar implications and I will look into fixing the recording length.

microbit-carlos:
> In terms of the approach, we can't use the web api gain because that just controls volume and not quality

(@finneyj please correct me if I am wrong here) In this case I think the "microphone sensitivity" in CODAL, is applied in the StreamNormaliser gain as a simple multiplication to the microphone raw data, which just scales the amplitude. So, if the web API for `gain` does the same by changing the volume I think that matches the behaviour.

srietkerk:
Thanks for the added context and info @microbit-carlos, with the help of @riknoll, we've applied a distortion on the sound that should get us closer to what we actually want here: https://github.com/microsoft/pxt-microbit/pull/6392

thsparks (approved):

Issue #6367, opened 2025-06-20
Simulator records audio for a longer time than the device


microbit-carlos:
The beta simulator currently records for about 12 seconds, and the device around 5 seconds.

```ts
input.onButtonPressed(Button.A, function () {
    record.startRecording(record.BlockingState.Blocking)
})
input.onButtonPressed(Button.B, function () {
    record.playAudio(record.BlockingState.Nonblocking)
})
```

<img width="683" alt="Image" src="https://github.com/user-attachments/assets/cb784595-522c-42f5-b0fd-007b44ff2a8d" />

Tested with:
makecode.microbit.org version:  [7.1.45](https://github.com/Microsoft/pxt-microbit/releases/tag/v7.1.45)
Microsoft MakeCode version:  [11.4.31](https://github.com/Microsoft/pxt/releases/tag/v11.4.31)
microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
codal-microbit-v2 runtime version:  [v0.3.0](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.3.0)

And test deployment from:
- https://github.com/microsoft/pxt-microbit/pull/6362

abchatra:
@microbit-carlos what about live site? Is this a regression?

microbit-carlos:
Live have similar times as beta, so not a regression.

Issue #5581, opened 2024-04-03
Recording extension: Microphone Sensitivity block not implemented in simulator


microbit-carlos:
**Describe the bug**
Related to:
- https://github.com/microsoft/pxt-microbit/issues/5286

The sample rate blocks has not implemented in the 2023 release yet.

**To Reproduce**
1. Add recording extension
2. Add the code below
4. Press button A to record and play
5. Press button B to change the sensitivity
6. Press button A to record and play again
7. The volume on the playback does not change

<img width="550" alt="image" src="https://github.com/microsoft/pxt-microbit/assets/29712657/12a7458d-48f2-48eb-9028-d878a3a140b7">

```ts
input.onButtonPressed(Button.A, function () {
    record.startRecording(record.BlockingState.Blocking)
    record.playAudio(record.BlockingState.Blocking)
})
record.setMicGain(record.AudioLevels.Low)

```

**Expected behavior**
The gain to be changed in the simulator playback.

**Screenshots**
Already added.

**micro:bit version (please complete the following information):**
V2

**Desktop (please complete the following information):**
 - OS: macOs
 - Browser: Chrome 123

**Additional context**
makecode.microbit.org version:  [6.0.25](https://github.com/Microsoft/pxt-microbit/releases/tag/v6.0.25)
Microsoft MakeCode version:  [9.0.15](https://github.com/Microsoft/pxt/releases/tag/v9.0.15)
microbit runtime version:  [v2.2.0-rc6](https://github.com/lancaster-university/microbit/releases/tag/v2.2.0-rc6)
codal-microbit-v2 runtime version:  [v0.2.63](https://github.com/lancaster-university/codal-microbit-v2/releases/tag/v0.2.63)

PR #6392, opened 2025-06-25
add mic sensitivity support for audio recording

srietkerk:
Big thanks to @riknoll for all the help with this.

This is the micro:bit side of the change and is basically just hooking up to the changes that were made in pxt. These changes makes mic sensitivity work and, in doing so, we are able to better control the quality of the sound. We made it a bit worse to better reflect the hardware experience.

I also modified the min and max recording time to make it so the recording times better matched the times on the hardware. The times still don't exactly match, but it is a lot closer now, about a 2 second difference rather than 5.

Fixes https://github.com/microsoft/pxt-microbit/issues/5281
Fixes https://github.com/microsoft/pxt-microbit/issues/5096
Fixes https://github.com/microsoft/pxt-microbit/issues/5581
Fixes https://github.com/microsoft/pxt-microbit/issues/6367

Edit: upload target: https://makecode.microbit.org/app/c0c10b4531a76d592776e9cbf251b69bdaef9866-127867270f#


jwunderl (approved):

microbit-carlos:
Thanks @srietkerk!
Could you gives us a ping when this is deployed into beta?

Issue #5104, opened 2023-04-29
Servo extension: stop servo doesn't work in simulator


martinwork:
**Describe the bug**
Arising for support ticket https://support.microbit.org/helpdesk/tickets/62402 (private)

**stop servo P0** does not cause a simulated continuous servo to stop moving 

Secondly, in a new project, when “stop servo P0” is placed in “on start” to initialise the servo as stopped, the simulator does not show a virtual servo. No servo is shown unless other servo blocks are present.

**To Reproduce**
1. Go to https://makecode.microbit.org/_8vpFUifK72rj, and click **Edit**
2. Click simulator button A
3. Click simulator button B
4. See error -  simulated servo still moving
5. Remove **continuous servo** block from **on button A pressed**
6. See error - simulated servo disappears 

**Expected behavior**
The simulator behaves like the hardware, and shows a servo when stop servo is added.

**Screenshots**
![image](https://user-images.githubusercontent.com/989126/235308490-707e548f-9f86-4f2c-a646-b1189e1f54c1.png)

**micro:bit version (please complete the following information):**
not hardware related

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows 10
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] 112.0.5615.138

PR #6393, opened 2025-06-25
add ability to stop continuous servo movement in sim

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/5104

the servo stop function does two things to the pin that the servo is connected to:

1. calls digital read (i assume to configure the pin as a digital input) 
2. sets the pin's pull mode to "none"

now i don't have a servo to test so i can't really say which of these two actions actually stops the servo moving, so this sim implementation just checks if both are true and stops the servo in that case


srietkerk (approved):
Looks good to me. 

Just wondering if this also answers the other mention in that issue:

> Secondly, in a new project, when “stop servo P0” is placed in “on start” to initialise the servo as stopped, the simulator does not show a virtual servo. No servo is shown unless other servo blocks are present.

abchatra:
@riknoll good to merge?

PR #6449, opened 2025-07-09
Add initial test procedures for WebUSB & BLE flashing.

microbit-carlos:


docs/release-tests/ble-flashing.md
```diff
@@ -0,0 +1,90 @@
...
+    ```javascript
+    basic.forever(function () {
+        datalogger.log(
+        datalogger.createCV("y", input.acceleration(Dimension.Y)),
+        datalogger.createCV("x", input.acceleration(Dimension.X))
+        )
+    })
+    ```
+2. Open the micro:bit app and connect to the micro:bit.
+3. TODO: Rest of the instructions will be added soon.
```
microbit-carlos, line 90: @martinwork I don't have an iOS app handy (maybe I'll be able to try one tomorrow, not sure). Would you be able to provide test steps for reading the datalog via BLE with the iOS app?

microbit-carlos:
@abchatra is there a way to load MakeCode beta into the electron and the Windows store apps?

abchatra:
No way for electron as it is packaged in. We will have electron build this week.

For windows app, you can type /@beta in the extension search box which will load beta.

docs/release-tests/webusb.md
```diff
@@ -0,0 +1,142 @@
...
+- MakeCode Windows Store App
+    - https://apps.microsoft.com/detail/9nmqdq2xzkwk
+
+## Tests to run on all matrix variations
+
+The following tests should be carried out with all combinations of the
+micro:bit boards, DAPLink versions, and browsers/apps listed in the
+"Test Matrix" section.
+
+### Test 1: WebUSB Download
```
martinwork, line 49: Add a step to remove existing pairings?

docs/release-tests/ble-flashing.md
```diff
@@ -0,0 +1,97 @@
...
+    basic.forever(function () {
+        datalogger.log(
+        datalogger.createCV("y", input.acceleration(Dimension.Y)),
+        datalogger.createCV("x", input.acceleration(Dimension.X))
+        )
+    })
+    ```
+2. Open the micro:bit app, go to "My Programs".
+3. Tap on the "Fetch MY_DATA (V2 only)" button and follow the instructions
+   to connect the micro:bit via Bluetooth.
```
microbit-carlos, line 95: @martinwork could you confirm if these steps are the same in iOS, I've tested this in Android, but don't have easy access to an iOS device.

martinwork: Yes, it's the same.

docs/release-tests/ble-flashing.md
```diff
@@ -0,0 +1,97 @@
...
+  micro:bits after using the Python Editor.
+    - Go to https://python.microbit.org and download the hex file
+      from the default project.
+
+Two mobile/tablet devices needed:
+- Android device with the micro:bit app installed.
+    - To load a different MakeCode URL long touch the micro:bit logo in the
+      app main screen and a modal will appear with the option change
+- iOS device with the micro:bit app installed.
+    - TODO: Add instructions here to load a different MakeCode URL in iOS.
```
martinwork, line 28: Long press the MakeCode back chevron to choose beta.

martinwork:
The Bluetooth flashing tests should include pairing with and flashing over a hex built in the MakeCode under test, which I have just done for iOS/Android + V1/V2 with MakeCode 8.0.8 partial flashes.

docs/release-tests/ble-flashing.md
microbit-carlos: Thanks!

microbit-carlos:
> Could both WebUSB and Bluetooth tests 1 and 2 be combined into one test, with test 2 as a last step, as below?

Yes, my original intention was to separate the tests to be able to also test them independently, but it's better to combine them and have fewer steps overall.

> Should the webUSB tests specify starting hexes, like the Bluetooth tests, or at least starting with a live hex?

Ah, yes, the full-flash step might not be tested correctly if the micro:bit had a MakeCode-under-test programme already. I'll add a step to flash the "meet the microbit" programme or a different hex.  


> [for Bluetooth] ... Do we need to test so many starting hexes?

Unfortunately I think we want to be safe, as there are classrooms with tablets that need flash the micro:bit with the factory image always with bluetooth.
From all the test files Python is probably the lower priority one, but I think it's worth checking as well, or do you think the "meet the micro:bit" test would definitely Python too?

> After USB flashing each starting hex, previous pairings should be deleted in Bluetooth settings, and in the iOS app Manage Connections page, before trying to Bluetooth flash, rather than deal with the errors that occur otherwise.

Is this something that we expect users to do? Would the app do this automatically or provide instructions to the user to do this?

> I think we can test with a USB powered micro:bit, to rule out hitting problems caused by low batteries.

Sounds good, I'll add it as well.


> Tap Download to flash the project (choose "My pattern is different")

Is this step always needed? Even if the pattern is the same?


> [Bluetooth MY_DATA] .. Add a pause in the datalogging project, so there isn't so much data (I guess we're only testing the utility service is present, not how well it performs)

Yeah, I was thinking logging as quickly as possible would always guarantee have quite a bit of data on the log, but it's not really necessary, so I'll add a small delay which ensure it takes a bit longr to fill the storage and still provide plenty of entries by the time it is read via BLE.

> [Bluetooth MY_DATA] ... Add a "During logging" fetch to ensure the service is present in both modes.

Sorry, I'm not sure I understood this one.

> The Bluetooth flashing tests should include pairing with and flashing over a hex built in the MakeCode under test, which I have just done for iOS/Android + V1/V2 with MakeCode 8.0.8 partial flashes. And with a Bluetooth program to test full flash.

Ah, that's a good catch! I'll add something as well.

martinwork:
> Do we need to test so many starting hexes?
> Unfortunately I think we want to be safe, as there are classrooms with tablets that need flash the micro:bit with the factory image always with bluetooth.

My thought was mainly: how are these hexes different from one another from the USB flashing point of view?

> After USB flashing each starting hex, previous pairings should be deleted in Bluetooth settings, and in the iOS app Manage Connections page, before trying to Bluetooth flash, rather than deal with the errors that occur otherwise.
> Is this something that we expect users to do? Would the app do this automatically or provide instructions to the user to do this?
> 

No, but the main flow of the apps is not aimed at users who keep USB flashing hexes., losing the pairing information.

> Tap Download to flash the project (choose "My pattern is different")
> Is this step always needed? Even if the pattern is the same?
> 
Same reason as above. It makes the app redo pairing rather than hitting an error to then come back and redo pairing 
 

> [Bluetooth MY_DATA] .. Add a pause in the datalogging project, so there isn't so much data (I guess we're only testing the utility service is present, not how well it performs)
> 
> Yeah, I was thinking logging as quickly as possible would always guarantee have quite a bit of data on the log, but it's not really necessary, so I'll add a small delay which ensure it takes a bit longr to fill the storage and still provide plenty of entries by the time it is read via BLE.
> 
I'm not sure why we need lots of data.

> [Bluetooth MY_DATA] ... Add a "During logging" fetch to ensure the service is present in both modes.
> Sorry, I'm not sure I understood this one.
> 

This tests that the config flags are set so the utility service can be available in application mode when required, as well as in Bluetooth mode., so that when already paired with a Bluetooth enabled data logging hex, you can fetch data while the data logging continues, without resetting to Bluetooth mode.

microbit-carlos:
> > Do we need to test so many starting hexes?
> > Unfortunately I think we want to be safe, as there are classrooms with tablets that need flash the micro:bit with the factory image always with bluetooth.
> 
> My thought was mainly: how are these hexes different from one another from the USB flashing point of view?

My main concern is small changes/tweaks between CODAL releases years apart (for instance the OOB hex file has a CODAL version over 3 years old, without the 3-reset-tap-to-pair).

I think we need to keep the MakeCode Live and MakeCode under test hex files. The Python editor hex file we should probably keep as well as it could have changes in configuration (accidental or on purpose) in future releases. To be fair, that should be caught in the Python Editor release tests, but it's possible the change was introduced in an in-between CODAL release (compared with the CODAL versions in the different MakeCodes). So, this is a good way to ensure it's still working combined with a new MakeCode release as well.

Between the "meet the micro:bit" and "out of box", would you be confident that if it works on one it would work with both?
If that's the case we could drop one these two.


> > After USB flashing each starting hex, previous pairings should be deleted in Bluetooth settings, and in the iOS app Manage Connections page, before trying to Bluetooth flash, rather than deal with the errors that occur otherwise.
> > Is this something that we expect users to do? Would the app do this automatically or provide instructions to the user to do this?
> 
> No, but the main flow of the apps is not aimed at users who keep USB flashing hexes., losing the pairing information.
> > Tap Download to flash the project (choose "My pattern is different")
> > Is this step always needed? Even if the pattern is the same?
> 
> Same reason as above. It makes the app redo pairing rather than hitting an error to then come back and redo pairing


Right, I see what you mean now. Could you provide some very simple steps to add them to the instructions? Or is `Tap Download to flash the project (choose "My pattern is different")` enough for both Android and iOS?



> > > [Bluetooth MY_DATA] .. Add a pause in the datalogging project, so there isn't so much data (I guess we're only testing the utility service is present, not how well it performs)
> > Yeah, I was thinking logging as quickly as possible would always guarantee have quite a bit of data on the log, but it's not really necessary, so I'll add a small delay which ensure it takes a bit longr to fill the storage and still provide plenty of entries by the time it is read via BLE.
> 
> I'm not sure why we need lots of data.

No, we don't need loads of data, but I wanted to make sure there is enough, as I'd expect an experienced tester trying to run through these steps as fast as possible to flash via WebUSB, and immediately after trying to read the MY_DATA file.


> > [Bluetooth MY_DATA] ... Add a "During logging" fetch to ensure the service is present in both modes.
> > Sorry, I'm not sure I understood this one.
> 
> This tests that the config flags are set so the utility service can be available in application mode when required, as well as in Bluetooth mode., so that when already paired with a Bluetooth enabled data logging hex, you can fetch data while the data logging continues, without resetting to Bluetooth mode.

Ah, yes, I'll add that as well.

martinwork:
I could be remembering it wrong, so it's hard to be 100% sure, but I would expect to find that all the old hexes have the same memory map and bootloader, so are the same for full flashing. The differences between them might relate to pairing, which won't be any different because of the new MakeCode version. Incompatibilities in the new MakeCode might only become apparent after further flashes, or connection to BLE services.

docs/release-tests/ble-flashing.md
```diff
@@ -0,0 +1,115 @@
...
+A test run for each of these hex files is required:
+- "Out of Box" hex file
+- "Meet the micro:bit" hex file
+- MakeCode Live Editor hex file
+- Python Editor hex file (micro:bit V2 only)
+- MakeCode Under Test hex file
+
+Steps:
+1. Connect the micro:bit to the computer via USB.
+2. Copy the hex file under test to the `MICROBIT` USB drive.
```
martinwork, line 59: 2a. Use the Bluetooth section of the Settings app to remove previous pairings.

docs/release-tests/ble-flashing.md
```diff
@@ -0,0 +1,115 @@
...
+3. Open the micro:bit app and load the MakeCode editor under test.
+    - From the front page, select "Create Code"
+4. Create a new MakeCode project with a simple programme (e.g. display
+   something on the LED matrix).
+    - If a programme has already been created from a previous test run,
+      it can be reused.
+5. Tap the "Download" button and follow the instructions to connect the
```
martinwork, line 66: 5. Tap the "Download" button and use "My pattern is different" to force pairing, then flash the programme.

PR #6502, opened 2025-07-16
Updating to a better image 

abchatra:
And removing old experiment flags.


thsparks:
I don't feel strongly, but the reason I used the original image instead of one with more in it (like this new one) was to avoid having stuff that gets obscured. With this new image, the text in the bubble gets covered up by the "Enabled" / "Disabled" text in the corner, which I personally think ends up looking a little messier. But I leave it up to you.

New version, which looks a little cluttered to me:
image

Versus previous version, which looks tidier to me and highlights the entry point:
image

thsparks (approved):
Left a note about the image but up to you what you want to do

abchatra:
ChatGPT thinks new one is better: 
image

PR #6563, opened 2025-08-05
Releasing 8.0.15 to live

abchatra:


jwunderl (approved):

abchatra:
There is a regression in /beta and we need to wait till that is fixed to release this payload:  https://github.com/microsoft/pxt-microbit/issues/6564

PR #6584, opened 2025-08-27
Add new microbit.org rows to home page

microbit-robert:
As discussed, this PR is a subset of agreed homepage changes.

These are the shared project ids that will likely need reinstating before this can be merged (though these are currently still working):
- _edAayo1kC04Y (Dance steps)
- _D2fETcEuMCvX (Poetry generator)
- _DACivW4iMe0t (Distance calculator)
- _gh4CetMLC5i4 (Traffic survey data logger)
- _7L8hXcRUDCPF (Solar panel experiment)


jaustin:
Does https://makecode.microbit.org/#pub:_T6bXj01sFV9L also need to be restored?

Jaqster:
I had some comments about these projects in email.  I feel like these should be called "Examples" so people know what they are (and are not).  Also this adds 5 new rows to the Home Page - are we okay with making it even longer?  Any way we can condense the rows?

Jaqster:
Also I feel like Games and Radio rows should be higher on the page as those are pretty popular.

PR #6631, opened 2025-10-21
Update SHA for jacdac/pxt-jacdac dependency

thomasjball:


abchatra:
@microsoft-github-policy-service rerun

PR #6637, opened 2025-10-29
Update links in translate.md to use 'makecode'

abchatra:
Kindscript has moved to makecode

### DIFF
--- a/docs/translate.md
+++ b/docs/translate.md
@@ -77,8 +77,8 @@ Document pages contain the text for any markdown page available on the MakeCode 
 
 | File | Description |
 | - | - |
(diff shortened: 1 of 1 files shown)


abchatra:
@ganicke looks like lot of links are pointing to wrong things. Can you please fix this PR with the right ones?

ganicke:
I think these need to get unlinked. You now have to be logged in since there is no R/O view to the file.

abchatra:
@ganicke makes sense. Can you take over if you have bandwidth?

ganicke:
> @ganicke makes sense. Can you take over if you have bandwidth?

NP, I'll take care of it.

ganicke (approved):

PR #6643, opened 2025-11-05
Fix keyboard navigation for builtin melody

microbit-robert:
Removes an extra unwanted (and often invisible) navigation step when moving left and right.


microbit-robert:
Behaviour before fix:
https://github.com/user-attachments/assets/4df35795-7fcf-4cd6-a77e-113e9b5c4a63

Behaviour after fix:
https://github.com/user-attachments/assets/75016747-07b7-47d1-bc57-ad4b39180b60

PR #6661, opened 2025-11-20
add joystickbit-plus

nvmgo:
Fixed the button jitter issue in joystickbit.

https://github.com/nvmgo/pxt-joystickbit-plus

### DIFF
--- a/targetconfig.json
+++ b/targetconfig.json
@@ -460,7 +460,8 @@
                     "sha": "8bab6a87265dc3e1e008a716a478306964b6d889",
                     "devUrl": "https://jacdac.github.io/pxt-jacdac-test/index.html"
                 }
-            }
+            },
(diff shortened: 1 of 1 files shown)


martinwork:
Hi @nvmgo I would suggest... create an issue in https://github.com/tinkertanker/pxt-joystickbit to let them know what problem you have spotted, and submit a pull request there with your suggested changes.

Looking at the changes... https://github.com/tinkertanker/pxt-joystickbit/compare/master...nvmgo:pxt-joystickbit-plus:master
To make it easier to see what has changed, update the existing ts file rather than creating a replacement.

There is a process for extension approval:
https://makecode.microbit.org/extensions/approval
https://support.microbit.org/en/support/solutions/articles/19000054952-makecode-extension-and-tutorial-approval

PR #6665, opened 2025-12-04
fix hovering for sim buttons in high contrast

srietkerk:
Change explained here: https://github.com/microsoft/pxt-arcade/pull/7212


aznhassan (approved):
:shipit:

PR #6688, opened 2026-01-26
Add maximum buffer size to serial reference

martinwork:
Arising from support ticket https://support.microbit.org/helpdesk/tickets/95799 (private)


ganicke:
Thanks for pointing those out. Let's go ahead and update those docs with that size value.

Issue #6664, opened 2025-12-02
stringPlayable's specifications are easily missed/ignored, and maybe misleading


martinwork:
**Describe the bug**
Arising from support ticket https://support.microbit.org/a/tickets/94149 (private)

Related: https://github.com/microsoft/pxt-microbit/issues/6603

The support ticket is about coding in JavaScript and points out that note D5 sometimes sounds an octave low.
```
music.play(music.stringPlayable("G4 A4 C5 A4 C5 D5 - -", 120), music.PlaybackMode.UntilDone)
music.play(music.stringPlayable("D5 - - - - - - -", 120), music.PlaybackMode.UntilDone)
```
It turns out that stringPlayable changes the next note after C5 to be explicitly octave 4.
 
stringPlayable's tooltip says **"up to eight notes [C D E F G A B C5] or rests [-] separated by spaces"**. 

It's easy to miss that this really means only eight notes, and only C5 or 4th octave notes, with no sharps flats, octave numbers or durations.

It's easy to ignore because stringPlayable works more generally, so long as it doesn't include the exact note "C5". Workaround: change C5 to C55 or C5:4, for example.

**Is the tooltip and the help wrong?**
https://makecode.microbit.org/reference/music/string-playable

It's not obvious (**is it supported?**) that it's possible to replace the melody editor with a text block, although a round trip to JavaScript mode, tends to revert to the melody editor showing "invalid".
https://makecode.microbit.org/_aCxgiR6pAHDx
<img width="255" height="62" alt="Image" src="https://github.com/user-attachments/assets/24f96b34-f775-4c22-84e5-53f1d8167191" />

Would a separate feature be useful, like https://makecode.microbit.org/_FfRTkg2wq5AT?
<img width="509" height="71" alt="Image" src="https://github.com/user-attachments/assets/81f143f6-1860-4de4-9ed6-ebb20677df10" />

In stringPlayable, C5 simply adds "4" to the next note, so can generate surprises. Example  https://makecode.microbit.org/_7HD0dUibbJxR shows that "D5 C5 D C5 D5 C5 C:2" is converted to melody string array [ "D5", "C5", "D4", "C5", **"D54"**,  "C5", **"C:24"**]

Using string[] directly is very clunky.
`music.play(new music.StringArrayPlayable(["G4:4", "A", "C5", "A4", "C5", "D", "R", "R", "G4"], undefined), music.PlaybackMode.UntilDone)
`
[...]

PR #6762, opened 2026-04-06
Improve/fix melody string parsing

riknoll:
this is part of the fix for https://github.com/microsoft/pxt-microbit/issues/6664

as noted in the issue, our melody parsing was a little broken. specifically, it had some weird behavior around switching octaves.

first, an explanation of our melody syntax, which looks like this:
```
[NOTE][MODIFIER]?[OCTAVE]?(:[DURATION])?

NOTE = ABCDEFG
OCTAVE = number (defaults to 4)
DURATION = number (defaults to 4) in sixteenth notes
MODIFIER = # or b
```
we have two functions that let you play melodies:  `music.beginMelody()` and `music.stringPlayable()`

these both use the same syntax with one important distinction. in  `music.beginMelody()`, changes in octave persist for subsequent notes while in `music.stringPlayable()` they do not. for example:

```
music.beginMelody(["A", "C5", "D", "G1"])       // will play A4 C5 D5 G1
music.play(music.stringPlayable("A C5 D G1"))   // will play A4 C5 D4 G1
```

(note the change in the third note, in the first call the octave carries over and in the second it does not and returns to the default octave of 4)

at least, that's how it is supposed to work. a bug in our string playable parsing was causing the octave only to revert to the default when the previous note was C5. this worked fine for the melody editor which can't do any notes outside of octave 4 except C5 but was buggy when used directly in TypeScript since you can use whatever octave you want.

this PR corrects that bug, and also makes some improvements to melody parsing in general:
* you can now have any number of flats and sharps (e.g. "C##" or "D#b#b")
* durations and octaves can now be more than a single digit
* better tolerance around bad whitespace, for example "ABCD" will now play correctly even without spaces in between the notes
* increased note range; `getFrequencyForNote` will now let you go outside the bounds of the lookup table
* `music.stringPlayable()` no longer parses the notes into a string array, now it just reads directly from the string in flash

additionally, `StringArrayPlayable`'s constructor can now take in a string, string array, or buffer instead of just a string array. as a result, you now have two options to specify longer songs in flash: you can use a string or the hex buffer format that we use for builtin melodies

for the record, the buffer format is super simple: two bytes per note. first byte is the midi note number or 0 for rest, second number is the duration of the note in sixteenth notes


libs/core/music.ts
```diff
@@ -550,103 +518,282 @@ namespace music {
...
+                const c = this.melody.charCodeAt(this.melodyStringIndex);
+                if (c < 48 || c > 57) break;
+                result = result * 10 + (c - 48);
+                this.melodyStringIndex++;
+            }
+            return result;
+        }
+
+        eatWhitespace() {
+            while (this.melodyStringIndex < this.melody.length && this.melody.charAt(this.melodyStringIndex) == " ") {
```
srietkerk: Should this be `===`?

srietkerk (approved):
One small question, otherwise looks good!

PR #6768, opened 2026-04-08
add unit conversion block

### LINKED ISSUE #4510: Add blocks to convert from degrees to radians and vice versa
Broken out from https://github.com/microsoft/pxt-microbit/issues/4312

> Also as most people will be familiar with degrees rather than radians, providing converter functions as blocks would be helpful.

These would be very helpful blocks in both micro:bit and arcade. @jwunderl already has some in his arcade-sprite-utils extension.

### LINKED ISSUE #6689: Get temperature in Farenheit block
During work in the US pilot, we've had the following feedback:

> A Fahrenheit temperature block would help to simplify science projects for the US.

I think it would be great to provide this, ideally without a new block, but by evolving the 'temperature' block to have a units dropdown/selector?

### LINKED ISSUE #6751: Conversion block
Create a new Math block to do conversions.
Temperature: https://github.com/microsoft/pxt-microbit/issues/6689
Degree/Radian: https://github.com/microsoft/pxt-microbit/issues/4510

riknoll:
fixes https://github.com/microsoft/pxt-microbit/issues/4510
fixes https://github.com/microsoft/pxt-microbit/issues/6689
fixes https://github.com/microsoft/pxt-microbit/issues/6751

adds a conversion block with three type of conversions:
* degrees <-> radians
* fahrenheit <-> celsisus
* seconds <-> milliseconds

these were the only useful ones i could think of in micro:bit, but perhaps there are more.
@jaustin @Jaqster @abchatra let me know if you have any opinions here

image
<br/>

image
<br/>
image

### DIFF
--- a/libs/core/helpers.ts
+++ b/libs/core/helpers.ts
@@ -6,11 +6,7 @@ enum UnitConversion {
     //% block="celsius to fahrenheit"
     CelsiusToFahrenheit,
     //% block="fahrenheit to celsius"
-    FahrenheitToCelsius,
-    //% block="seconds to milliseconds"
-    SecondsToMilliseconds,
-    //% block="milliseconds to seconds"
-    MillisecondsToSeconds
+    FahrenheitToCelsius
 }
 
 namespace Math {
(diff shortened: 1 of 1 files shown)


srietkerk (approved):

Jaqster:
Color value conversions?  
image

riknoll:
@Jaqster doesn't really make sense here because the HSV block returns an RGB value

jaustin:
We think the seconds-->milliseconds block here is not like the others, it's almost making the conversion seem more magical than it is, and *1000 is fine! (good!)

The other two are great. 

Seconds --> minutes, and minutes-->hours are perhaps more useful though.

That said, I think keeping fewer options more readable is good.

riknoll:
removed milliseconds conversion. @jwunderl @srietkerk can i get another review?

PR #6794, opened 2026-04-22
Implement additional keyboard controls for sim sliders

microbit-robert:
This PR prevents slider values from looping and implements controls for the Page Up and Page Down keys (increment value in steps of range / 10) and the Home and End keys (set slider to min and max values).

Duplicated code has been pulled out into a common key handler function.

The pins will addressed in a follow up PR due to additional complexity in the pin values (digital vs analog) and their dual roles as both sliders and buttons.


riknoll (approved):
nice cleanup!

PR #6795, opened 2026-04-22
Improve keyboard controls and screen reader for sim pins

microbit-robert:
When the pins are acting as buttons (touch mode only), this PR:
- adds the button role and removes slider roles and attributes
- prevents the mouse from changing the pin value

When the pins are acting as sliders, this PR:
- prevents slider values from looping and implements controls for the Page Up and Page Down keys (increment value in steps of range / 10) and the Home and End keys (set slider to min and max values)
- the pinKeyHandler ensures that the current value is clamped appropriately before passing to the commonKeyHandler
- uses the live region to announce that the pins are in read-only mode and adds the readonly attribute to the pin element
- uses a separate variable to track the `aria-valuenow` value and only uses `aria-valuetext` when `aria-valuenow` is not a number (e.g., ~500, 100%)
- adjusts the mouse start / move position depending on whether the pin is in analog or digital mode

When the pin has a pressed listener **and** there is code to read the pin analog or digital value, the pin takes on the slider role and attributes. As this is more of an advanced feature, this is likely the value the user will want to interact with. However, both keyboard and mouse users can still trigger pin pressed events using Enter/Space and click events.

### DIFF
--- a/sim/visuals/microbit.ts
+++ b/sim/visuals/microbit.ts
@@ -634,7 +634,8 @@ path.sim-board {
                     this.pins[index].removeAttribute("aria-valuetext");
                     this.pins[index].removeAttribute("aria-readonly");
                 } else  {
-                    accessibility.setAria(this.pins[index], "slider", this.pins[index].firstChild.textContent);
+                    this.pins[index].setAttribute("role", "slider");
+                    this.pins[index].ariaLabel = this.pins[index].firstChild.textContent;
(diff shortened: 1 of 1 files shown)


riknoll:
@microbit-robert this is ready to be rebased

microbit-robert:
@riknoll Thanks. This has now been rebased with one additional tweak made. setAria doesn't replace an existing role, so it's not used for the pins where the role can switch.

riknoll (approved):

PR #6796, opened 2026-04-23
Give sim application role to fix screen reader user interaction

microbit-robert:
Screen readers don't always respect the role attribute on SVG elements. While running NVDA on Windows and using the current live version of MakeCode, keyboard interaction with the sliders work, but you are unable to trigger a button press on Button A or Button B for example using Space or Enter, even though these elements have role="button" and tabindex="0" on them. There may be a more nuanced way to address this (such as adding non-SVG buttons inside foreignObjects which is more disruptive), but role="application" is one fix. This only impacts the micro:bit simulator and I'm not sure there is much in the way of screen reader semantics that will be lost here. Thoughts welcome.

### DIFF
--- a/sim/visuals/microbit.ts
+++ b/sim/visuals/microbit.ts
@@ -1029,7 +1029,9 @@ path.sim-board {
                 "y": "0px",
                 "width": MB_WIDTH + "px",
                 "height": MB_HEIGHT + "px",
-                "fill": "rgba(0,0,0,0)"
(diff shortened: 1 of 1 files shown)


riknoll (approved):
seems fine to me! i guess the "proper" alternative would be to switch to using html button/input elements placed over the SVG for most of our interactions, but that's a much bigger lift. i wonder if screen readers are fine with HTML elements inside of foreignObject SVG elements

PR #6919, opened 2026-06-10
fix blockly overrides and bump pxt-core

riknoll:
fixing the overrides and bumping with the microbit prs i merged this morning


abchatra (approved):

abchatra:
FYI @microbit-robert @microbit-matt-hillsdon good to do this change with conjunction with the blockly PR in pxt to ensure both are in sync.

Issue #6467, opened 2025-07-10
pext/yotta:latest Docker image fails to build the with the latest microbit-v2-samples


microbit-carlos:
In PR https://github.com/microsoft/pxt-microbit/pull/6452 I originally updated the microbit-v2-samples repository to the latest tag. This is the error the CI showed:
https://github.com/microsoft/pxt-microbit/actions/runs/16171740734/job/45646848016
```
Traceback (most recent call last):
  File "build.py", line 144, in <module>
    build(options.clean, verbose=options.verbose, parallelism=options.parallelism)
  File "/home/build/prjclone/utils/python/codal_utils.py", line 16, in build
    use_ninja = shutil.which("ninja") is not None or platform.system() == "Windows"
AttributeError: 'module' object has no attribute 'which'
```

This is expected if running the `python build.py` command with Python 2 instead of Python 3, as `shutil.which()` was [added in Python 3.3](https://docs.python.org/3/library/shutil.html#shutil.which) released in 2012.

Looking at the invocations in the pxt repo it looks like it's using `python` instead of `python3` and that very likely uses Python 2:
https://github.com/search?q=repo%3Amicrosoft%2Fpxt+build.py&type=code

Some of the options could be to change the pxt invocations to `python3` or to install `python-is-python3` in the docker images so that `python` points to the Python 3 installation.

Amerlander:
I use that Dockerfile when I have to build daldts:

```dockerfile
FROM node:18-bullseye

# Install system dependencies needed for PXT builds
RUN apt-get update && apt-get install -y \
    git \
    python3 \
    python3-pip \
    build-essential \
    cmake \
    mercurial \
    ninja-build \
    gcc-arm-none-eabi \
    gdb-arm-none-eabi \
    openocd \
    wget \
    curl \
    vim \
    less \
    tree \
    srecord \
    && rm -rf /var/lib/apt/lists/*


# Ensure 'python' points to 'python3' for legacy scripts
RUN ln -s /usr/bin/python3 /usr/bin/python

RUN pip3 install "markupsafe<2.1" "jinja2<3.1"
RUN pip3 install --prefer-binary --no-cache-dir yotta

# Set working directory
WORKDIR /workspace

# Install PXT CLI and common packages globally
RUN npm install -g pxt


# Option A) Mount directories and build in Startup script
# This allows the user to mount their local directories into the container
# RUN mkdir -p /workspace/pxt-calliope \
#              /workspace/pxt \
[...]

Issue #5721, opened 2024-07-10
`pext/yotta:latest` image type has been deprecated and stopped working with latest docker


microbit-carlos:
Trying to build Makecode with Docker v26+ fails, as trying to pull the `pext/yotta:latest` image (for the V2 CODAL build) throws the following error:

```
$ docker pull pext/yotta:latest
latest: Pulling from pext/yotta
[DEPRECATION NOTICE] Docker Image Format v1 and Docker Image manifest version 2, schema 1 support is disabled by default and will be removed in an upcoming release. Suggest the author of docker.io/pext/yotta:latest to upgrade the image to the OCI Format or Docker Image manifest v2, schema 2. More information at https://docs.docker.com/go/deprecated-image-specs/
```
https://github.com/microsoft/pxt-microbit/blob/bd4bda19f33ba30ea61e3620c8fa069b8306d07e/pxtarget.json#L204-L207

This docker image type was deprecated in Docker v19, it has been disabled already in v26, and v27 (which was release a couple of weeks ago) has completely and permanently removed the functionality:
https://docs.docker.com/engine/deprecated/#pushing-and-pulling-with-image-manifest-v2-schema-1

The GH Actions runner images (which we use in the CODAL CI to test changes with a MakeCode build) currently use Docker v26, which needs a workaround to enable this image type, but it won't be long before they upgrade to v27 and things will stop working if there isn't a way to downgrade the docker version from the images.

riknoll:
we're working on publishing these images to Microsoft's public container registry. unfortunately, due to licensing issues, i can't update the existing images in the pext dockerhub repo

microbit-carlos:
Ah, good to know, thanks for the update @riknoll.
Will pushing the `pext/yotta:latest` image to the new registry using a modern docker version update the image manifest type? Or will it need to be rebuilt?

If it needs to be rebuilt it might be a good opportunity to upgrade the image, as `pext/yotta:latest` is using GCC v6, and something a bit more modern like v10.3 might improve build sizes (just an assumption based on experience, but of course we won't know until we try it and compare).

riknoll:
@microbit-carlos I'll be rebuilding the images before I push them, so no worries there! I've actually already been updating all of the dockerfiles used by our compile service backend, just need somewhere to put them for public consumption.

totally down for experimenting with different gcc versions!

abchatra:
If this is still an issue reopen.

microbit-carlos:
[...]

PR #6925, opened 2026-06-11
pxtarget.json: githubCorePackage bumped from v0.2.13 to v0.3.0

KierPalin:
Hi @abchatra @thomasjball,

This is the aforementioned compilation issue:
[CMake 4.0.0+ removed compatibility with versions older than 3.5](https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features). This is a problem because the current githubCorePackage is on v0.2.13 which means that the [microbit-v2-samples/utils/cmake/JSONParser.cmake](https://github.com/lancaster-university/microbit-v2-samples/blob/490a890626674aa859a2f9f19d96dc8e2367f137/utils/cmake/JSONParser.cmake#L23) has a cmake_minimum_required of 3.1, thus pxt-microbit cannot be compiled when using the latest versions of CMake.

Fortunately, [the latest microbit-v2-samples ](https://github.com/lancaster-university/microbit-v2-samples/blob/04b7089d82af24534f3dcd460a9c343850b60b5d/utils/cmake/JSONParser.cmake#L23) bumps the cmake_minimum_required to 3.6, so we need only update the githubCorePackage to it.

I tested the change by building [microcode-v2](https://github.com/microbit-apps/Microcode-v2/) and its dependencies with this change and the latest CMake version 4.3.0.

There are a few changes between v0.3.0 and v0.2.13, as visible in this [comparison for v0.3.0 and v0.2.13](https://github.com/lancaster-university/microbit-v2-samples/compare/v0.2.13...master).

Thanks,
Kier

### DIFF
--- a/pxtarget.json
+++ b/pxtarget.json
@@ -205,7 +205,7 @@
                 },
                 "codalBinary": "MICROBIT",
                 "githubCorePackage": "lancaster-university/microbit-v2-samples",
-                "gittag": "v0.2.13",
+                "gittag": "v0.3.0",
                 "serviceId": "mbcodal2",
                 "dockerImage": "pext/yotta:latest",
                 "yottaConfigCompatibility": true


abchatra:
@carlosperate

microbit-carlos:
I think using the lastest microbit-v2-samples tag had issues because the docker image is very old and was still using Python 2 by default, which was officially sunset by the PSF 6 years ago and deprecated from CODAL more than 3 years ago.
- https://github.com/microsoft/pxt-microbit/issues/6467

@KierPalin if I remember correctly we increased the minimum CMake from 3.1 to 3.6 because we found some files were using APIs only available in 3.6+, but I don't think old features were refactored in preparation for their deprecation, or at least we haven't checked if we are affected. So if CMake v4 drops compatibility with versions lower than 3.5 it's possible CODAL is using some of those deprecated features, or relaying on old behaviour and side effects of those APIs.

Have you tried building with CMake v4? Does everything work correctly?

KierPalin:
@microbit-carlos I compiled [MicroData](https://github.com/microbit-apps/MicroData) and [MicroCode v2](https://github.com/microbit-apps/microcode-v2) with this proposed change using cmake version 4.3.0 again just now. These are both relatively demanding applications which enabled me to test the micro:bits sensors, datalogging and display. 

Both of these programs work correctly on hardware but I don't know how to test for the older features or the micro:bit v1; do you want me to test this change with something else? I can easily compile any MakeCode programs with this change to test this PR further if you have any that would be particularly useful?

Thanks,
Kier

KierPalin:
As an addendum I should mention that I compiled everything locally and that I don't use Docker, so I cannot comment on that issue.

Compiled on Linux with:
cmake version 4.3.0
node version v18.20.8
ninja version 1.13.2
PXT_FORCE_LOCAL=1 PXT_RUNTIME_DEV=1 PXT_ASMDEBUG=1 PXT_NODOCKER=1 PXT_COMPILE_SWITCHES=rawELF

microbit-carlos:
Thanks @KierPalin knowing that CMake v4 builds and the compiled output works correctly for MicroData and MicroCode is a good test, so that great 👍
I just had a look at the repo and forgot we are running the microbit-v2-samples CI with both CMake 3.6 and "lastest", so we've been testing the builds there with CMake v4 for a while, but that only generates CI builds, we generally don't tests those in hardware, so confirming that's all good was definitely a needed test.

No need to worry about micro:bit V1 builds, it uses a different build system (yotta) and a different docker image, so that one should stay untouched.

@abchatra fixing https://github.com/microsoft/pxt-microbit/issues/6467 would be great to unblock the latest codal-microbit-v2-samples, MakeCode is currently stuck in a version released 3 years ago.

Also, because of the othrer docker issue https://github.com/microsoft/pxt-microbit/issues/5721 , users building locally cannot use docker for V2 builds, so they are forced to use their locally installed tooling, and anybody installing any CMake release from the last year would encounter this issue.

PR #6953, opened 2026-06-23
Update pxt-core dependency to version 12.3.21 to pick up offline app related fix

jwunderl:
Primarily bumping to pick up https://github.com/microsoft/pxt/pull/11342 for test offline app build, since the bugs listed there are most noticeable / annoying in offline app scenario and ideally addressed now


aznhassan (approved):
:shipit:

PR #6983, opened 2026-06-30
Remove 'showProjectDescription' from pxtarget.json

jwunderl:
Removed the 'showProjectDescription' property from the configuration as not enabling on share page


aznhassan (approved):
:shipit:

